PR #60, opened 2015-01-16
python binding module access improvement

maxvyaznikov:
Small improvement for better usage.

I saw next message into python binding:
> May need to set the PYTHONPATH to get this to work:
> PYTHONPATH=$PYTHONPATH:/usr/local/lib/python2.7/dist-packages/link-grammar
> or something similar ...

Problem is in the package name: It shouldn't be "link-grammar", but "linkgrammar" or "link_grammar".
Can you help me to fix it? - I found, that this name had been taken from PACKAGE (file "configure.ac") and uses globally. So, I can offer to convert a minus-char into underscore or drop it especially for python package name (See http://stackoverflow.com/a/1706459/2261861).

### DIFF
--- a/bindings/python/__init__.py.in
+++ b/bindings/python/__init__.py.in
@@ -2,3 +2,5 @@
 # __init__.py  file for Link Grammar Python bindings
 #
 __version__ = "@VERSION@"
+
+from .linkgrammar import *
--- a/bindings/python/linkgrammar.py
+++ b/bindings/python/linkgrammar.py
@@ -6,6 +6,10 @@ import locale
 
 import _clinkgrammar as clg
 
+
+__all__ = ['Parser', 'ParseOptions', 'Dictionary', 'Link', 'Linkage']
+
+
 class Parser(object):
     # A cost of 2.7 allows the usual cost-2 connectors, plus the
     # assorted fractional costs, without going to cost 3.0, which


linas:
I don't understand the comment about the dash. Other python packages have dashes in the names, e.g. right now I am looking at  nose-1.3.4.egg-info  and cocos2d-0.6.0-py2.7.egg each of which has not just one dash, but two in the package name ...

linas:
Anyway, thanks for the other patch.  I am not sure the python bindings are very well written .. they seem to perform several un-needed cpu-wasting steps.  And the longer-than-80-char lines are supremely annoying ...

maxvyaznikov:
Sorry for my poor English. I'm talking about a module name, not a egg-files. See https://www.python.org/dev/peps/pep-0008/#package-and-module-names

linas:
Unless there is some actual bug due there being a dash in the name, I would prefer to not change the name.

maxvyaznikov:
A see 2 ways:
* change install path only for python binding (/usr/local/lib/python2.7/dist-packages/linkgrammar)
* change install path after installation (e.g. move package from /usr/local/lib/python2.7/dist-packages/link-grammar to /usr/local/lib/python2.7/dist-packages/linkgrammar)

linas:
But why does it need to be changed at all? Is there  a bug?

maxvyaznikov:
Unfortunately, yes. Package is unavailable after installation (/usr/local/lib/python2.7/dist-packages/link-grammar):
```
>>> import linkgrammar
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named linkgrammar
>>> import link_grammar
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named link_grammar
>>> import link-grammar
  File "<stdin>", line 1
    import link-grammar
               ^
SyntaxError: invalid syntax
```

But if I will move it into /usr/local/lib/python2.7/dist-packages/link_grammar, then ```import link_grammar``` will be success. Also for /usr/local/lib/python2.7/dist-packages/linkgrammar and ```import link_grammar```.
It's all because of package and modules name rules into the python.

linas:
The first import should work, the second two are just wrong.   Take a look at  the example test case, in the folder bindings/python/tests.py  When everything is properly installed, then it runs just fine.

maxvyaznikov:
Yes, tests.py working. But only because of it's location - into the same folder. If you'll move it to another directory (/tmp, for example), then import will be broken. 
tests.py works only because of python will found linkgrammar.py file near the tests.py. But globally, this bondings installation is wrong.
Just try to create a new file with import linkgrammar in some folder, not the package location.

linas:
tests.py works from /tmp, for me ... did you say 'make install', to install the libraries, and then set the PYTHONPATH, as the README file recommends?

maxvyaznikov:
Ofcourse, yes. Except PYTHONPATH, because thats incorrect. User mustn't change PYTHONPATH for each new package. Problem here is in the python package name, then changing a PYTHONPATH is a... (How it will be in English on programmers slang?) a crutch.

maxvyaznikov:
Recently I wrote, that saw 2 ways to solve a problem without modification of PYTHONPATH:
* change install path only for python binding (/usr/local/lib/python2.7/dist-packages/linkgrammar)
* change install path after installation (e.g. move package from /usr/local/lib/python2.7/dist-packages/link-grammar to /usr/local/lib/python2.7/dist-packages/linkgrammar)

But now I see an another one:
* create a symbolic link /usr/local/lib/python2.7/dist-packages/linkgrammar -> /usr/local/lib/python2.7/dist-packages/link-grammar (just tested - working)

linas:
Hmm.  Yes, changing the install path to  /usr/local/lib/python2.7/dist-packages/linkgrammar is reasonable ... or creating a symbolic link, I don't care much.  However, I don't want to rename the entire package, just to make this one change.

Note, thought, that setting the "crutch" of PYTHONPATH is pretty normal, a lot of python packages require this, in order to work right.  So this is not unusual.  For whatever historical reason, the python folks have not yet solved this problem in any elegant way, they keep using the crutch...

linas:
I mean .. disallowing dashes in names is just fucked-up stupid.  They should fix python itself, to do the right thing, instead of forcing link-grammar and hundreds of other packages to conform to pythons bugs .. but alas, ...

maxvyaznikov:
Or, thanks. I'll make a patch in some time.

maxvyaznikov:
Nope. I tried to find, where is I can insert a command to create link, but... Have no enough expirience with make to do it =(

linas:
OK, So I think I misunderstood what you were saying before.   But I still don't understand.  If I do these things:
1) create a link from /usr/local/lib/python2.7/dist-packages/linkgrammar to link-grammar
2) don't set PYTHONPATH
3) copy tests.py to /tmp
4) cd /tmp ; ./tests.py

I get an error
```
$ ./tests.py
Traceback (most recent call last):
  File "./tests.py", line 9, in <module>
    from linkgrammar import Parser, Linkage, ParseOptions, Link
ImportError: cannot import name Parser
```
So, for me. modifying the install path is not enough to resolve the issue. I'm sorry I don't know python internals any better than this ...

linas:
There will be a version 5.2.5 in a few days or a week.    After that, the next release won't be for maybe 2-4 months.

linas:
Re: my previous note: I had failed to install yout new __init___.py  I did that.  I now get a new error:

```
$ ./tests.py
Traceback (most recent call last):
  File "./tests.py", line 10, in <module>
    import _clinkgrammar as clg
ImportError: No module named _clinkgrammar
```

I'll try hacking around that...

linas:
Never mind ... _clinkgrammar is supposed to be private anyway.   I'll twiddle the Makefile to automatically create the link, and I think this will close all the remaining issues, I think ...

linas:
I beleive its now all fixed in commit 6cb2ee5

maxvyaznikov:
>from linkgrammar import Parser, Linkage, ParseOptions, Link
>ImportError: cannot import name Parser
Yes, *without* commit 991c684c65bc301a4d6cbce7e8cd9b7aaa98e151, import can be done only like this:
from linkgrammar.linkgrammar import Parser, Linkage, ParseOptions, Link
But 991c684c65bc301a4d6cbce7e8cd9b7aaa98e151 "moved" all important classes into the top level.

>I now get a new error:
>import _clinkgrammar as clg
>ImportError: No module named _clinkgrammar
Now it can be imported as:
import linkgrammar._clinkgrammar as clg

linas:
Well, if I modify tests.py to say "import linkgrammar._clinkgrammar as clg", then the error move around: I get:
```
$ ./tests.py
Traceback (most recent call last):
  File "./tests.py", line 9, in <module>
    from linkgrammar import Parser, Linkage, ParseOptions, Link
  File "/home/linas/src/novamente/src/link-grammar-git/bindings/python/linkgrammar.py", line 7, in <module>
    import _clinkgrammar as clg
ImportError: No module named _clinkgrammar
```

and then, if I modify linkgrammar.py, I get:
```
$ ./tests.py
Traceback (most recent call last):
  File "./tests.py", line 9, in <module>
    from linkgrammar import Parser, Linkage, ParseOptions, Link
  File "/home/linas/src/novamente/src/link-grammar-git/bindings/python/linkgrammar.py", line 7, in <module>
    import linkgrammar._clinkgrammar as clg
ImportError: No module named _clinkgrammar
```

so yet another step seems to still be needed ..

maxvyaznikov:
```
  File "/home/linas/src/link-grammar-git/bindings/python/linkgrammar.py", line 7, in <module>
    import linkgrammar._clinkgrammar as clg
ImportError: No module named _clinkgrammar
```
You ran test.py from the /home/linas/src/link-grammar-git/bindings/python/ directory, there is we have linkgrammar.py file. So, python tries to import from file near, instead of use the global package.
So, only for test.py, than should be run from directory of package, import should be:
import _clinkgrammar as clg

But! We can make a more flexible solution like this:
```
try:
    import _clinkgrammar as clg
except ImportError:
    import linkgrammar._clinkgrammar as clg
```
P.S.: I'm sorry, I didn't tested yet

linas:
I'm saying that neither version works.  To make the discussion simple. consider instead example.py -- copy it to /tmp and run it. There's always an error, no mater what form.

linas:
Oh never mind. I wrote too soon. I just found aonther bug.

linas:
I think it all works, now, finally.

PR #65, opened 2015-01-22
Python binding improvement

maxvyaznikov:
Completely reworked python binding. There were many unneeded steps into API.

### DIFF
--- a/bindings/python/linkgrammar.py
+++ b/bindings/python/linkgrammar.py
@@ -69,7 +69,7 @@ class ParseOptions(object):
         """
         return clg.parse_options_get_linkage_limit(self._obj)
 
-    @verbosity.setter
+    @linkage_limit.setter
     def linkage_limit(self, value):
         if not isinstance(value, int):
             raise TypeError("Linkage Limit must be set to an integer")
@@ -86,7 +86,7 @@ class ParseOptions(object):
         """
         return clg.parse_options_get_disjunct_cost(self._obj)
 
-    @verbosity.setter
+    @disjunct_cost.setter
     def disjunct_cost(self, value):
         if not isinstance(value, float):
             raise TypeError("Distjunct cost must be set to a float")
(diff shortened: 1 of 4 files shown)


linas:
Wow.

bindings/python/linkgrammar.py
```diff
@@ -81,210 +38,306 @@ def __init__(self, verbosity = 0,
...
+    def __del__(self):
+        if hasattr(self, '_obj'):
+            clg.sentence_delete(self._obj)
+            del self._obj
+
+    def split(self):
+        return clg.sentence_split(self._obj)
+
+    def __len__(self):
+        return clg.sentence_length(self._obj)
```
linas: This is technically incorrect, and will need to change. The actual length of the sentence may vary, depending on the linkage.   So, for example: "We saw the seasand beach"  which is clearly mis-spelled .. but is it "we saw the seas and beach" (6 words) or is it "we saw the seasoned beach" (5 words)?   These are relatively rare, but the new tokenizer design allows these.

Similarly, the sentence no longer has a unique splitting into words, so the split() above cannot be used.

bindings/python/linkgrammar.py
```diff
@@ -81,210 +38,306 @@ def __init__(self, verbosity = 0,
...
+
+    def parse(self):
+        n = clg.sentence_parse(self._obj, self.parse_options._obj)
+        for i in xrange(n):
+            yield Linkage(i, self, self.parse_options._obj)
```
linas: Hmm, also, should probably not create linkages until they are actually wanted. That will save some cpu time (not much, but some)

linas:
Whoops .. I just merged this, and I realized I shouldn't have.   I was planning on having a version 5.2.5 first, which contains only bug fixes, and then have your code, and the new tokenizer, in version 5.3.0, sometime later.   Although I suspect that no one is actually using the python bindings, it still seems unfair to just change them without a somewhat bigger version bump.

maxvyaznikov:
Yeah, I thought you wait unitil next release =)
And, yes, python binding was very dirty to serious usage.
I'll fix bugs above and make more tests.

ampli:
Hello Max,
Please see my problem at #67.
I would appreciate your opinion on that.
Is there a way to run the tests without installing?

PR #68, opened 2015-01-24
Fix linking step for x64.

ampli:
Add missing linkgrammar-x64.lib definition in the <link> section.
Minor change: Add EOL at the last line.


linas:
cherry picked and merged in #82

Issue #56, opened 2015-01-12
path_found is not getting freed, and also can be incorrectly resued


ampli:
- It is not getting freed when the dictionary is closed .
- Reusing this variable may cause a mess if a chdir() or dictionary_set_data_dir() is invoked between dictionary opens.

So I propose to free and nullify it in dictionary_delete() and dictionary_six().
I thought of 2 ways to do that:

1. Make it a non-static global, and use extern to access it in the needed 2 other places.
2. Make it a private static in a new function cache_data_path(), that will be used to get/free/set this variable.

Linas,  I can send a fix if you tell me which way to choose.

```
/* XXX static global variable used during dictionary open */
static char *path_found = NULL;

static void * dict_file_open(const char * fullname, void * user_data)
{
	const char * how = (const char *) user_data;
	FILE * fh =  fopen(fullname, how);
	if (fh && NULL == path_found)
	{
		path_found = strdup (fullname);
		prt_error("Info: Dictionary found at %s", fullname);
	}
	return (void *) fh;
}
```

linas:
The more technically-correct fix is the second one ...

ampli:
I just sent pull request #59.
While fixing it I noted an old problem: the dictpath is still searched after path_found is prepended to the file name.  It could cause inconsistencies of finding the rest of the files in a different data directory than the first one, and also if a particular file is missing in a data directory it could be found in another one.

Another slight problem is the addition of what seems to me a useless path component on "first_time_ever".
For example, on dictionary_create_lang("en") it adds an "en:" component at the start of the dictpath, so the rest of searches start with "en/en/..." when using this component.

I looked at the old revisions and it seems this two problems exist from the start of this code. On dictionary_create_lang("ru") it causes 512 fopen() failuers (and zero failures after fixing it). Surprisingly, saving these 500+ fopen()'s result only in a few percent saving of CPU time on dictionary open...

So I would like to send a fix that removes two code sections, as indicated below by #if 0:
1. Remove adding the apparently useless (unless I missed something)  prevpath component.
2. Invoke opencb() unconditionally. On a full path it will do the same. On a path prepended by path_found it will prevent the dictpath search. However, if a file is not found in a place it should be, but exists in another dictpath component directory, a mess will still happen. This cannot not be fixed without restructuring the code.

In case of a full path that is not found, the code currently tries to append it to the dictpath components,
a thing that is useless, and may theoretically cause looking in undesired locations. But I don't know if it worth fixing.

```
[...]

PR #70, opened 2015-01-25
Rewrite the dictionary file opening stuff (issue #56)

ampli:
- Invalidate the dictionary path cache on each dictionary create.
   This has the benefit of notifying the path of each dictionary open,
   not only the first one.
 - The algorithm now really ensures that all the files will be opened
   from within the same dictionary directory.
 - Free the dictionary path cache on dictionary delete.
 - Don't try to append full-path dictionary arguments to the dictpath.
 - Prefer dictionaries at relative paths (usually user or source
   directory dictionaries) over a system-installed dictionary.
 - The path_found dictionary path cache now uses the previous prevpath
   mechanism (that didn't work as intended).
 - Fix opencb() last argument declaration to avoid the need of const
   conversion.
 - Change dictpath from a dynamically generated string path to an array
   of paths.
 - on binreloc, search essentially the same path (with the addition of
   the binreloc path after the relative paths, but before the shared
   system path).
 - dictionary_get_data_dir() is not invoked any more for each file
   open, but only until the dictionary path cache is created.
 - Remove the leading slashes in #include paths, so they are not
   tried any more as absolute paths. This allows not retrying opening
   failed real absolute paths as relative ones (using the dictpath or
   the dictionary path cache).
 - Simplified dict_file_open() and dictopen() to the minimum possible.
 - Minor cleanup: white space.

### DIFF
--- a/link-grammar/dict-common.c
+++ b/link-grammar/dict-common.c
@@ -113,6 +113,8 @@ Dictionary dictionary_create_lang(const char * lang)
 {
 	Dictionary dictionary = NULL;
 
+	object_open(NULL, NULL, NULL); /* Invalidate the directory path cache */
+
 	/* If an sql database exists, try to read that. */
 	if (check_db(lang))
 	{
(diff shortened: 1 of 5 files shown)


linas:
this got merged as a cherry pck into the 5.2 branch

PR #71, opened 2015-01-26
Fix compilation errors when defining _DEBUG.

ampli:
- Word' has no member named 'string' any more.
 - Debug function print_expression() doesn't exist.


linas:
cherry picked into the 5.2.5 main branch. Will merge into master shortly.

PR #72, opened 2015-01-27
Fix setting dict-lang to freed memory.

ampli:
If lang contains no slashes (the normal situation), dict->lang could get set to malloc'ed
memory that is being freed soon (in dictionary_create_from_utf8()
or dictionary_create_default_lang()). Fixed to always use string set.

Also add a debug message for printing the found language, using
_DEBUG (that already enables printing dictionary paths).
It can be included using:
./configure CPPFLAGS=-D_DEBUG=1


linas:
Cherry picked and pushed into 5.2.x-maint  I have not been able to fiogure oout how to pull your branches into the main branch without also pulling in all the python code

PR #75, opened 2015-01-28
utf8_next fix

dtim:
This is a pull request for the already discussed small fix, which affects unicode handling in the link-parser executable.

This fix works in Linux (and possibly other unices too). Visual C++ (MS Visual Studio 2013) compiles it, so I think there is no need to make a different implementation for Visual C++.

At the same time, it is hard for me to say if the patch works well in Windows, because it seems that Windows build of link-parser executable, LinkGrammarExe.exe, tries to convert between console code page and UTF8 on input, but not on output.

linkgrammarexe

I have no experience in using Cygwin/msys, so I could not build link-grammar in these environments yet.

Sorry for the delay.


linas:
I just now cherry-picked the changes that you made, and added them to the 5.2.x-maint branch, which was hen merged into master.  Due to some unrelated mistakes, this was the easiest way to do this; so I won't merge this pull request; your changes, however, are now in master (and will be in 5.2.5)

Issue #78, opened 2015-01-29
Tests.py fail (3 errors): TypeError in linkage_print_diagram


ampli:
I will send a pull request that fixes it.

linas:
i think this was fixed in pull req #80

Issue #67, opened 2015-01-24
Problem running tests.py


ampli:
I am trying to run tests.py without installing.

I have set  PYTHONPATH to the location of linkgrmmar.py and to the location of _clinkgrammar.so, as follows:
```
export PYTHONPATH=/usr/local/src/link-grammar-devel/master/bindings/python:/tmp/link-grammar/master/bindings/python/.libs
ls /usr/local/src/link-grammar-devel/master/bindings/python /tmp/link-grammar/master/bindings/python/.libs

/tmp/link-grammar/master/bindings/python/.libs:
_clinkgrammar.a   _clinkgrammar.lai                  _clinkgrammar.so    _clinkgrammar.so.5.2.5
_clinkgrammar.la  _clinkgrammar_la-lg_python_wrap.o  _clinkgrammar.so.5

/usr/local/src/link-grammar-devel/master/bindings/python:
AUTHORS  __init__.py.in  LICENSE  linkgrammar.py  linkgrammar.pyc  Makefile  Makefile.am  Makefile.in  README.md
```

However, I get an ImportError:

```
python2.7 /usr/local/src/link-grammar-devel/master/bindings/python-examples/tests.py
Traceback (most recent call last):
  File "/usr/local/src/link-grammar-devel/master/bindings/python-examples/tests.py", line 12, in <module>
    import linkgrammar._clinkgrammar as clg
ImportError: No moduIs there a way tole named _clinkgrammar
Exit 1
```
I can fix this import error by a change in tests.py:
```
-import linkgrammar._clinkgrammar as clg
+import _clinkgrammar as clg
```

Now it runs fine:
```

python2.7 /usr/local/src/link-grammar-devel/master/bindings/python-examples/tests.py
............................link-grammar: Info: Dictionary found at /tmp/link-grammar/master/data/en/4.0.dict
.....................
----------------------------------------------------------------------
Ran 49 tests in 8.183s

OK
```

So how can I run the current tests.py without installing?
(I would like to fix "make check" to do that.)

linas:
If you install, then the PYTHONPATH  is no longer necessary.  If you don't install ... I don't know. I am not a python expert. However, Max, who contributed the fix, might know .. you can try asking him ...

maxvyaznikov:
For that case we can make a tests.py to trying to import from multiple possible places.
[...]

PR #79, opened 2015-01-29
Fix TypeError (#78)

ampli:
Fix the following error, that causes 3 tests to fail:
      File ".../bindings/python/linkgrammar.py", line 313, in diagram
        return clg.linkage_print_diagram(self._obj, display_walls, screen_width)
    TypeError: in method 'linkage_print_diagram', argument 2 of type 'bool'

Plus two additional commits, see my next comment.

### DIFF
--- a/bindings/python-examples/Makefile.am
+++ b/bindings/python-examples/Makefile.am
@@ -6,7 +6,7 @@
 
 # Run tests.py if the user says 'make check'
 TESTS = tests.py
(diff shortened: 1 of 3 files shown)


ampli:
I have just added to this branch  (**tests**) 2 commits for two additional fixes, and I see that github concatenates such additions to the already-open pull request...

So I will leave it at that even though they are actually fixing what I mentioned in #67.

ampli:
I now realize that if **srcdir** is not defined, it is better that parses-xx.txt will be taken from the current directory and not from "/". Since this pull request has not been merged yet, I will just **overwrite** it with a new one.

PR #95, opened 2015-02-12
Continue the Hebrew development

ampli:
I converted some more things in the demo Hebrew dict to match Hebrew.
I also added several example sentences, and for the first time I added wrong sentences also.

Two of the added wrong sentences get parsed. I still need to find the reason for that.

While adding these stuff, I encountered and fixed a bug in mprefix_split().


data/he/4.0.dict
```diff
@@ -23,52 +23,46 @@
 % A few connector and other changes have been done for Hebrew.
 % [ap]
 
-% Dictionary version number is 5.0.0 (formatted as V5v0v0+)
+% D version number is 5.0.0 (formatted as V5v0v0+)
 <dictionary-version-number>: V5v0v0+;
```
linas: you probably want to change this to 5.3.0  And what's up with the letter "D"?

ampli:
It was due to my replacement of the D connector...
I will send you an additional pull request for both of these things when I
return home.

2015-02-12 22:54 GMT+02:00 Linas Vepštas <notifications@github.com>:

data/he/4.0.dict
```diff
@@ -3,7 +3,7 @@
  % Experimental prototype Hebrew dictionary                                  %
  %                                                                           %
  % Based on en/tiny.dict,                                                    %
- %       Copyright (C) 1991-1998  Daniel Sleator and Davy Temperley          %
+ %       Copyright (C) 1991-1998  D Sleator and D Temperley          %
```
linas: oh, well, in that case: some more D's here.

PR #98, opened 2015-02-13
Include @ in disjuct_list_str

williampma:
Not sure why the multi-connector property is not included in the string.  This change is not really necessary, but it does make sure the disjunct of a WordInstanceNode outputted by RelEx really is the original disjunct.

Apparently the corpus code uses the disjunct_str also.  I am not sure how to test whether this breaks that or not.  Where is the unit test for LG?

### DIFF
--- a/link-grammar/disjuncts.c
+++ b/link-grammar/disjuncts.c
@@ -31,13 +31,17 @@
 static char * reversed_conlist_str(Connector* c, char dir, char* buf, size_t sz)
 {
 	char* p;
-	size_t len;
+	size_t len = 0;
 
 	if (NULL == c) return buf;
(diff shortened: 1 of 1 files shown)


linas:
This code is also used in language learning.

It has another bug that, I swear, it did not use to have: I could have sworn that it used to print repeated connectors, so that "he ate a big round plump olive" would print out:
```
olive.s    0.000  A- A- A- Ds**x- Os-
```
I think it might be better for language learning, not sure, if the above was printed, instead of @A-  Weird, not sure how this happened.

linas:
The unit tests are thin. There are two: some native C unit tests in the tests directory, and some more extensive ones in the bindings/pythhon/tests.py file

PR #100, opened 2015-02-18
Hebrew demo dictionary: Introducing subject/verb/object person/gender/number agreement

ampli:
Instead of developing the noun definitions from scratch (comma and conjunction handling,
wall attachment, etc.) which I cannot see how I can do in a timely
manner, I chose to incorporate the noun definitions from the full English
dictionary and **gradually** change them for subject/verb/object/number
agreement.

Adjective agreement is not done yet.

For verbs, I took very little, by now, from the full English dictionary, partially because I didn't understand (yet) many aspects of the implementation, and partly because the Hebrew verb system is significantly different.

I'm not sure if it is better to restart all the work, this time taking the English dictionary as starting point.
But the needed  work looks very complex either way.

Morpheme handling of anything but prefixes has not been done yet.

The example sentences are rather "not interesting", and their aim is only to directly test the grammatical features that are supported by now, and to show the things that are needed to be fixed next.


linas:
FYI:  now that you are doing this, you should be aware of some additional markup that is unused or rarely used in the english edict.

A$ means   A+ or A-

(A+ ^ B+) means  (A+ & B+) or (B+ & A+)

these are useful if you ever have free word-order: where you don't much care what order things happen in.

When you do have free word order, you will want to mark a head and dependent.  Do this by putting small lower-case h in front of the head  and small lower-case d in front of the dependent.  h can only attach to d (or nothing) (but no constraint on word-order)  likewise d can attach to h (or nothing))

for an example in the english dict, see hWV+  and dWV-  I'm vaguely planning on adding h, d in many places in the english dict, but the urgency is very low.

PR #105, opened 2015-03-03
Fix off-by-one bug in the previous label fitting fix.

ampli:
It missed to consider the space between the first and the second words.


linas:
Hi Amir!

PR #106, opened 2015-03-05
History handling improvments

ampli:
3 patches, to fix these problems:

- If the user has a .editrc file with an history size of more than 20, e.g:
history size 100
it is not observed on start-up history reading - only the last 20  entries (the hard-coded default history size)  get read.

- Reading batch files overwrites the user input history.  This is a misfeature.

- After this problem was fixed, still several input history lines got overwritten.

### DIFF
--- a/data/ar/4.0.batch
+++ b/data/ar/4.0.batch
@@ -1,8 +1,8 @@
 
+!batch
 !verbosity=1
 !echo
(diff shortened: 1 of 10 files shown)


linas:
Thank you; the clobbered history file was an annoyance.

PR #107, opened 2015-03-12
Update parses-en.txt - labels between adjacent words are not truncated now

ampli:
Fix a tests.py  failure.

### DIFF
--- a/bindings/python-examples/parses-en.txt
+++ b/bindings/python-examples/parses-en.txt
@@ -4,9 +4,8 @@
 
 Ithis is a test
 O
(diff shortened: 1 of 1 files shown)


linas:
thank you!

linas:
p.s. have you seen the emails from rohit, on language learning?  Its on the other mailing list. Exciting! 
In a few weeks, he may start hacking the sqlite dictionary code...

ampli:
Yes, I have read that thread and also most of the others on opencog related to NLP.

PR #115, opened 2015-04-10
Fix two crashes

ampli:
Found by using a C file as input...


linas:
Hey, could you add the code snippets that cause the crash to either 4.0.fixes.batch or to 4.0.fixes-long.batch (if they take a long time to parse before crashing)?  There already is some C code in 4.0.fixes-long.batch

PR #116, opened 2015-04-10
Fix problems in handling "!" commands

ampli:


ampli:
If it is better to just exit in case of invalid unicode, I can send a new pull request.

linas:
this fix seems reasonable.

PR #119, opened 2015-04-18
Enhance the unit test

ampli:


linas:
thanks, sorry for late response, I'm still partly crippled but almost back to normal.

PR #120, opened 2015-04-18
Fix constituent printing problems

ampli:


ampli:
Hi Linas,
I can make a further improvement regarding this stuff per the added FIXME.
This needs a pointer from the Linkage struct back to shared data in the Sentence struct,
unless you see a better way to do that.

linas:
Go ahead, make further improvements ... 

When I do design changes, I mostly try to achieve these goals:

* have fewer pointers from one struct to another
* have smaller structs
* have fewer structs
* have fewer dependencies between different or "distant" parts of code.

Of course, its not always possible: so: if adding a pointer to Sentence in Linkage allows some other major simplification to happen, then its a good trade-off.  But if it doesn't actually result in simplification (of the code, or at least "conceptual simplification) then its a bad tradeoff.

ampli:
Hi Linas,
The problem I would like to fix is that linkage_post_process() for the constituents code, is currently performed per linkage in linkage_create(), and this is redundant if the user doesn't need the constituents.

To that end I moved the this call of linkage_post_process(), back into do_print_flat_constituents().  However, linkage_post_process() now has an additional argument **Parse_Options** in order to use the **twopass_length** option, and the constituents API functions don't have it (so I cannot pass it). Instead of changing the API further, I added a Boolean **is_sentence_long** to the Linkage struct.

Now everything works fine, but the said (recently added) **Parse_Options** augment to `linkage_post_process()` becomes redundant.

**Possible solutions:**
1. We can remove **Parse_Options** from `linkage_post_process()` and return to its documented API that doesn't include it (I didn't encounter a documentation to this parameter addition).
2. A **Parse_Options** struct member can be added back to the Linkage_struct, instead of **is_sentence_long** that I added in this modification.

BTW, I thought of adding **is_sentence_long** (or maybe **twopass_length**) directly to the Sentence struct, but couldn't find a nice way to do that (because sentence_create() doesn't have a **parse_Options** parameter).

Do you have a preferred solution?

linas:
Hmm. I would prefer not to have Parse_Options appear in any struct; it should stand on its own.

If  Parse_Options can be removed from linkage_post_process(), then yes, this seems like the right thing to do.

For constituents, it would be better to create a new API for do_print_flat_constituents() that takes the two-pass length.  For backwards-compatibility, the old API should just use the default two-pass length.  Yes, this means that the old API now effectively ignores parseOptions, but I think that's OK.

ampli:
The problem is that Parse_Options can be removed from linkage_post_process() **only** if it can get **twopass_length** in another way, e.g. from an API parameter or from a new Linkage strut member as in my proposed change.

Regarding constituents, the current user API consists of two functions - linkage_constituent_tree(Linkage) and linkage_print_constituent_tree(Linkage, mode). It doesn't seem to me a good idea to add **twopass_length** as an additional parameter to them, as it (twopass_length) looks to me as an implementation detail (for optimization) that has nothing to do with user API functionality. However, if **twopass_length** is added in the Linkage struct as I proposed, there is no need to add it as a user API parameter, because the Linkage struct is already a parameter in both of these functions.

If we will agree that exposing a **twopass_length** parameter in these 3 user API calls is not a good idea, then I propose to implement both of the following things:
-  remove **Parse_Options** from linkage_post_process(), returning it to its old (and documented) API form.
- put one of these in the the Linkage struct for the sole use of by linkage_post_process():  **twopass_length** OR a Boolean value **is_sent_long**.

linas:
OK Yes, put that way, it seems like the right thing to do.

PR #130, opened 2015-05-23
Man page: Generate the system dictionary path.

ampli:
- Additional man page modifications.
 - Add the directory search order of the dictionary.
 - Automatically update it in the man page according to the configuration
   options.


linas:
wow, thanks!

PR #132, opened 2015-05-24
Manpage

ampli:
- Fix autogen breakage caused by the previous man page fix.
- Adapt the man page for man2html and grohtml.


ampli:
I just pushed yet another small commit to this branch.
It seems the changes in it got joined to the last commit here, without my added commit text.
I guess that's OK.

PR #137, opened 2015-07-08
Fix memory reallocation in setup_domain_array

ampli:


linas:
Thanks Amir.

Is it time to publish version 5.3.0?  Is eerythign stable enough to publish a 5.3.0?

ampli:
It seems to produce essentially the same results as the previous version on detailed run of the batches.
The tokenizing needs a radical design revision, and I'm working on an analytic design (as opposed to the current ad-hoc code). I'm also working on extending the Hebrew demo.
It also contains several bug fixes and extensions.

However, the API has not been validated thoroughly. We know that the API part that is used internally produces the same results (on the batches), and that the tests suite passed when I recently checked it (but see below). But other aspects of the API are not checked (maybe users can contribute some code that does something real, that can be used for tests). One difference that I know of is that it is not possible any more to find the list of sentence tokens by looking at the unsplit_word elements, and there is no new API for that yet. It is also not well-checked that the assigned unsplit_word elements are always appropriate (I refer to the unsplit_word in the old Word struct, used by the parser).

In addition, the SAT parser is **totally broken** now, and usually segfaults. I propose to temporarily **disable** it! I looked at the problem and it is not trivial but seems fixable. I would like to do that just after I finish the new tokenizing code.

Regarding the tests, I recently upgraded to Fedora 22. Just now I found I have a problem to run the tests due to a yet unidentified strange problems (both "make check" and running tests.py directly fail,
each on a totally different problem).  I will try to fix that and return to you soon.

Beside of that, I think it 

Beside of that, I think it can be published.

(It may also be a good idea to publish a new maintenance release, with this last fix.)

linas:
Can you add the two sentences to the 4.0.fixes.batch file, at the end?  And also add a line to ChangeLog stating that this bug was fixed.

ampli:
Note that these two sentences are problematic.
The first one causes "The diagram is too high.".

The second one panics. So maybe it is better to put it in 

So maybe it is better to put both in 4.0.fix-long.batch?

linas:
OK, yes.

ampli:
Sent pull request #144.

PR #139, opened 2015-07-08
Avoid sorting the linkages if no linkages found

ampli:
1. Avoid sorting the linkages if no linkages found.
Else, qsort() is invoked with NULL as its first argument,
but the function template states the first argument should not be null.
2. Fix possible missing linkage->pp_info.

### DIFF
--- a/link-grammar/linkage.c
+++ b/link-grammar/linkage.c
@@ -857,6 +857,9 @@ void linkage_free_pp_info(Linkage lkg)
 
 /**
  * Part of the API. Although maybe it shouldn't be.
+ * It has been updated to initialize sent->constituent_pp, instead of
+ * doing it in linkage_create(), so it will be done only on demand.
+ *
  * This is just a wrapper around do_post_process, and all that it does
  * is to set up some domain data in the linkage->pp_info pointer.
  * There do not seem to be any users for this data !?  Should this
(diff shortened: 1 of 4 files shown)


link-grammar/linkage.c
```diff
@@ -876,7 +879,14 @@ void linkage_post_process(Linkage linkage, Postprocessor * postprocessor)
...
 
 	if (NULL == linkage) return;
-	if (NULL == postprocessor) return;
+
+	if (NULL == postprocessor)
+	{
+		Sentence sent = linkage->sent;
+
+		sent->constituent_pp = post_process_new(sent->dict->hpsg_knowledge);
+		postprocessor = sent->constituent_pp;
```
linas: I'm confused by this and the next fix.  If the user did not ask for constituent printing, then we should not be doing any constituent processing. 

Keep in mind tht the constituent post-processing is different than the "regular" post-processing.

ampli:
The problem is this:

The user invoked the API call:
`str = linkage_print_links_and_domains(linkage);`

This function, in turn, invokes the following functions:
`linkage_get_link_num_domains()`
`linkage_get_link_domain_names()`

Both of these functions need `linkage->pp_info`.

This struct member is assigned in `linkage_post_process()`.

Previously (before my recent postprocessing on demand change) this function has been invoked unconditionally from `linkage_create().`

So according to the above we see we need to add it to `linkage_post_process()`.
So I added at its start:
`linkage_post_process(linkage, linkage->sent->constituent_pp);`

But there is a problem here: its second argument (`Postprocessor *`) is NULL at this time!
And `linkage_post_process()` just returns when it is NULL,, because it needs it non-NULL.
So hence I initialized it there...

Can it be solved in another way?

It must be *solved*, since a NULL `inkage->pp_info` causes a crash. Such a crash doesn't happen in the **link-parser** program (and hence I missed this problem), but it does happen when running the multi-thread test! because it invokes `linkage_print_links_and_domains()` when `nkage->pp_info` is NULL.

linas:
Yeah, the problem here is that there are two pre-processors, and the constituent_pp is the wrong one to use.  It should have been the main PP instead.   I am currently doing a major refactoring of that code.

linas:
Fixed this in pull req #152.  Apparently, the domain printing was broken for a while, now.

PR #143, opened 2015-07-09
Fix tests/Makefile.am

ampli:
Add -lpthread for the multi-thread program. This is needed at list of
Fedora 22, and apparently on some other OSs.
Add the existing mem-leak program.
Add LDFLAGS.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -12,19 +12,23 @@ AM_CPPFLAGS = -I$(top_srcdir) -I$(top_builddir) $(CXX_FLAGS) \
 # -----------------------------------------------------------
 # TESTS declares the tests to actually run;
 # check_PROGRAMS are the binaries to build.
-TESTS = dict-reopen multi-thread
-check_PROGRAMS = dict-reopen multi-thread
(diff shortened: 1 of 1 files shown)


tests/Makefile.am
```diff
@@ -12,19 +12,23 @@ AM_CPPFLAGS = -I$(top_srcdir) -I$(top_builddir) $(CXX_FLAGS) \
...
 dict_reopen_SOURCES = dict-reopen.cc
 multi_thread_SOURCES = multi-thread.cc
+mem_leak_SOURCES = mem-leak.cc
 
 LDADD = $(top_builddir)/link-grammar/liblink-grammar.la
 if HAVE_SQLITE
 LDADD += $(SQLITE3_LIBS)
 endif
 
+multi_thread_LDADD = -lpthread $(LDADD)
```
linas: Older versions of configure.in had a test for HAVE_PTHREAD   -- I may have removed it, as an attempt to simlify things, as it was mostly unused.  I'm not sure, but perhaps MacOS or WIndows don't have pthreads.

PR #152, opened 2015-07-19
Fix broken domain printing

linas:
This fixes the previously-broken domain printing (see pull req #139) and re-organizes the code just a bit.


ampli:
The multi-thread test still fails on this:
```
str = linkage_print_links_and_domains(linkage);
```

PR #154, opened 2015-07-20
Regex file: Implement negative regex'es.

ampli:
Regex'es that are preceded by !, if they match a token, stop further match tries of the same regex name. Thus, they can serve as a kind of a negative look-ahead.


linas:
thanks

PR #160, opened 2015-08-02
Fix memory leak in freeing linkages->pp_info.

ampli:
It happens when num_links shrinks due to discarded stem links,
because the pp_info array is allocated before this potential shrinkage.

The shrinking of num_links due to discarded ZZZ links in remove_empty_words() is not
problematic because it happens before the pp_info array is allocated. (The shrinking of  num_words there is not problematic for the size of chosen_disjuncts because it is tracked by cdsz.)

Here is how this memory leak is observed:
```
$ link-parser ru
link-grammar: Info: Dictionary found at ../ru/4.0.dict
link-grammar: Info: Using locale en_US.utf8.
link-grammar: Info: Dictionary version 5.1.0.
link-grammar: Info: Library version link-grammar-5.3.0. Enter "!help" for help.
linkparser> вверху плыли редкие облачка.
Found 2 linkages (2 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS= 3.00 LEN=14)

    +-------------------------Xp-------------------------+
    +---------Wd---------+----------SIp----------+       |
    |         +----EI----+           +----Api----+       |
    |         |          |           |           |       |
LEFT-WALL вверху.e плыли.vnndpp редкие.api облачка.ndnpi . 

linkparser> ^Dlink-grammar: Info: Freeing dictionary ru/4.0.dict
link-grammar: Info: Freeing dictionary ru/4.0.affix
Bye.

=================================================================
==28181==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 16 byte(s) in 2 object(s) allocated from:
    #0 0x7f4e0e0e9a0a in malloc (/lib64/libasan.so.2+0x98a0a)
    #1 0x7f4e0dcc8292 in exalloc /usr/local/src/link-grammar-devel/master/link-grammar/utilities.c:504
    #2 0x7f4e0dc7a58a in linkage_set_domain_names /usr/local/src/link-grammar-devel/master/link-grammar/post-process.c:398
    #3 0x7f4e0dc2b676 in post_process_linkages /usr/local/src/link-grammar-devel/master/link-grammar/api.c:724
    #4 0x7f4e0dc310fc in chart_parse /usr/local/src/link-grammar-devel/master/link-grammar/api.c:1431
    #5 0x7f4e0dc3193c in sentence_parse /usr/local/src/link-grammar-devel/master/link-grammar/api.c:1506
    #6 0x406cda in main /usr/local/src/link-grammar-devel/master/link-parser/link-parser.c:835
    #7 0x3649c206ff in __libc_start_main (/lib64/libc.so.6+0x3649c206ff)

SUMMARY: AddressSanitizer: 16 byte(s) leaked in 2 allocation(s)
```

BTW, I cannot understand why it says
`Direct leak of 16 byte(s) in 2 object(s)`
since 3 links are discarded in this sentence, + a shorter pp_info array gets freed, so this is apparently a leak in 4 objects. Valgrind also reports on 2 blocks:
`16 bytes in 2 blocks are definitely lost`

I'm also not sure why we have not observed this leak earlier.
(Maybe it started due to the recent post processing rearrangements?)

### DIFF
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -369,6 +369,7 @@ struct Linkage_s
 
 	Linkage_info    lifo;         /* Parse_set index and cost information */
 	PP_info *       pp_info;      /* PP domain info, one for each link */
+	size_t          ppisz;        /* Alloc'ed length of pp_info */
 	PP_data         hpsg_pp_data; /* Used in constituent code */
 
 	Sentence        sent;         /* Used for common linkage data */
--- a/link-grammar/post-process.c
+++ b/link-grammar/post-process.c
@@ -349,9 +349,9 @@ void linkage_free_pp_info(Linkage lkg)
 	size_t j;
 	if (!lkg || !lkg->pp_info) return;
 
-	for (j = 0; j < lkg->num_links; ++j)
+	for (j = 0; j < lkg->ppisz; ++j)
 		exfree_domain_names(&lkg->pp_info[j]);
-	exfree(lkg->pp_info, sizeof(PP_info) * lkg->num_links);
+	exfree(lkg->pp_info, sizeof(PP_info) * lkg->ppisz);
 	lkg->pp_info = NULL;
 }
 
@@ -374,6 +374,10 @@ void linkage_set_domain_names(Postprocessor * postprocessor, Linkage linkage)
 
 	/* The only reason to build the type array is for this function. */
 	build_type_array(postprocessor);
+
+	/* Keep track of pp_info's alloc'ed length, in case linkage->num_links
+	 * shrinks due to discarded ZZZ links (and maybe others). */
+	linkage->ppisz = linkage->num_links;
 
 	linkage->pp_info = (PP_info *) exalloc(sizeof(PP_info) * linkage->num_links);
 


ampli:
Another side affect caused by discarding stem links is on the content of linkages->pp_info.
It no longer corresponds to the links, due to discarded links.
Is this a problem?
Should we also discard the corresponding elements of linkages->pp_info?

ampli:
I found just now that this fix is not complete.
There is still a related memory leak in build_graph().
I will try to fix it too.

linas:
Well, when the ZZZ links are discarded, does that change the numbering of the links?

grep for LinkIdx and you will see it used both to access pp_info and to access linkage_get_link_label() linkage_get_link_llabel(), linkage_get_link_lword() etc.  If these are returning incorrect values, because ZZZ was discarded, then its a fairly serious bug, since these interfaces are heavily used.

The same LinkIdx is used to access pp_info .. except .. (a) Russian does not use post-processing, and (b) I don't use the "domain info" for anything; it does., however, have to be correct to get the HPSG-style, tree-bank-style printing to work right. (It probably doesn't work for Russian, anyway...).

ampli:
Empty words are discarded in remove_empty_words().
It is careful to rearrange link_array and to update num_links.

In compute_chosen_words(), which may potentially remove stem links and may also remove other words with ZZZ links, this exact link array update is also done.

So it looks to me there is no problem in that regard.

So it is still needed to similarly rearrange pp_info.
Where is the best location to do that?
Possibilities:
1. At the end of compute_chosen_wors().
2. After compute_chosen_words() in linkage_create().

BTW, the discarded ZZZ and LL_* links still remain on the chosen disjuncts, and this is observable by using the API. Is there any problem in that? Should we discard them from there too?

linas:
Option 2 seems to be more "modular".. pp_info is not used by russian, and english doesn't use ZZ or LL, so I don't want to hurt English performance.

The ZZZ links need to be invisible to the API; they're just a weird LG artifact.  The LL links need to stay, they provide actual, needed information.

ampli:
After trying to implement `pp_info` trimming according to the the updated `link_array`, I found it is too cumbersome to do it not at the same time (I need to propagate info from nested functions regarding which links got discarded, since the links are not numbered by an ID field). However, it seems that doing it at the same time of shrinking `link_array` is easy and lightweight.

In any case I have to touch this code due to a severe bug I found in it (I will send a patch soon) with the `HIDE_MORPHO` code when the stem links are not `LL*`. This bug causes  wrong results for Turkish,
and it is clear the tr dict author encountered these problems, which prevented effective dict development.

Another option for correcting the `pp_info` is more code rearrangement. But for that I need to understand your comment regarding calling `linkage_set_domain_names()` in `post_process_linkages()`:
`/* ... the needed info is discarded much too soon. ... */`.

**Two questions:**
1. Where is this needed info discarded?
2. How can I validate that my changes to pp_info are correct, beside printing English constituents before and after, for sentences with discarded empty words? (Such sentences exist in English too). 

For 2, I guess I will need to a add python bindings for `get_link_domain_names()` and `get_link_num_domains`. (Are there more relevant calls?)

linas:
OK, so, perhaps we should discard PP_info entirely.

In the note to the function `linkage_set_domain_names()`, it suggests that the only place where domain names seem to be used is when they are printed.  (line 366)

Thus, perhaps the loop at line 405 can be moved to the function `linkage_get_link_domain_names()` at line 328, so that the names are only ever computed when they are asked for. If this is done, then we pretty much don't need PP_info any more. (We probably need to change the API for `linkage_get_link_domain_names` so that user specifies the arrary ...)

However, reading the code, I see that the pp_info domain names are computed from `p->d_type_array`  which means that if the length/alignment of PP_info is wrong, then so is the length/alignment for d_type_array.  So that would need to be fixed, too.

Now, `d_type_array` seems to be built here:  `build_type_array` at line 242, and the only place where it is called is ... you guessed it.   

And now, I finally find the spot you are talking about: line 724 of api.c ...  So `build_type_array` needs access to pp_data. But pp_data is deleted at line 726 of api.c   So, here is an easy fix:  replace the call to `linkage_set_domain_names` by a call to `build_type_array`.  Remove `build_type_array` from `linkage_set_domain_names`  and finally rework the code as noted up top; the domain names are not even computed, at all, unless they are asked for,   Seems like that should solve all the problems, I think ...

ampli:
Does it mean that `linkage_get_link_domain_names()` should then be be like:
`linkage_get_link_domain_names(Postprocessor *, Linkage, LinkIdx)`

linas:
You don't need the Postprocessor * argument, because you can ook that up from the sentence, right? linkage->sentace->postprocessor.   However, to deal with memory management and freenig of arrays, you need an interface like this:
```
void linkage_get_link_domain_names(const Linkage linkage, LinkIdx index, char **, int len)
```
i.e. the user passes in the array to be filled in, and its length so that we don't write past the end of it.

ampli:
Why not to allocate the memory in the library, and add an API function `free_link_domain_names()?

ampli:
I could not understand how all of that is going to solve the problem caused by removing  links after the postprocessing step.

The `d_type_array`, which is indexed by the link number, will still be build in `post_process_linkages()`.
Its index is the link number.

Later when the user calls `linkage_get_link_domain_names()`, at this point some links may have got dropped. How can we now extract info from this d_type_array, when our link numbering is potentially  different?

linas:
1) yes a free_link_domain_names would be better/more consistent with the rest of the api.

2) Not sure. My point was that if  pp_info that has a problem with indexing; then so does d_type_array.

linas:
Anyway, I am now cnfused as to whether there even is a problem, or not.  So, for example, `remove_empty_words()` does the removal; it is called by `compute_chosen_disjuncts()`, which is called before any post-processing is done.  So it seems like post-processing only happens after link-removal;  thus, how is it even possible that  pp_info and _d_type_array get the wrong size/indexing?

ampli:
>1) yes a free_link_domain_names would be better/more consistent with the rest of the api.

So I will implement the change this way (adding `free_link_domain_names()`).

>So it seems like post-processing only happens after link-removal; thus, how is it even possible that pp_info and _d_type_array get the wrong size/indexing?

As you observed, there is no problem in the link removal done before post-processing.
The problem arises from link removal done **after** post-processing, in `compute_chosen_words()`,
which is invoked from `linkage_create()`.
I think the changed link indexing can be solved by adding an index-number filed in `Link_s`:
```
struct Link_s
{
	LinkIdx link_num;       /* Ordinal number of the link */    <-------------
	size_t lw;              /* Offset into Linkage->word NOT Sentence->word */
	size_t rw;              /* Offset into Linkage->word NOT Sentence->word */
	Connector * lc;
	Connector * rc;
	const char * link_name; /* Spelling of full link name */
};
```
Then we can use it to index  `d_type_array`, not caring if links has been dropped meanwhile.

The remained question is how to test the correctness of the change, as the Russian dict don't use domain names.
Maybe the following can be used: Parsing an English sentence with quotation marks, and then creating indexes with and without `!test=removeZZZ` (which will drop the quotation marks, along with their ZZZ links). The result of `linkage_get_link_domain_names()` in this two cases can be then checked for correctness. The problem is that it may not be a good idea to add such a test to tests.py.

--

If this solution (adding `LinkIdx` in `Link_s`) seems to you fine, I will send you a patch that implements your proposed change.

ampli:
In continuation to my above post, and patch #174, here is the problematic section:
```
static void linkage_set_domain_names(Linkage linkage)
{
	...
	for (j = 0; j < linkage->num_links; ++j) <------------- j (1)
	{
		k = 0;
		for (d = pp->d_type_array[j]; d != NULL; d = d->next) k++; <------------- j (2)
		linkage->pp_info[j].num_domains = k;
		if (k > 0)
		{
			linkage->pp_info[j].domain_name = (const char **) exalloc(sizeof(const char *)*k);
		}
		k = 0;
		for (d = pp->d_type_array[j]; d != NULL; d = d->next) <------------- j (2)
		{
			char buff[5];
			snprintf(buff, 5, "%c", d->type);
			linkage->pp_info[j].domain_name[k] =
			      string_set_add (buff, postprocessor->string_set);

			k++;
		}
	}
```
But if  links are discarded in `compute_chosen_word()`, the numbering of (1) is not the same as in (2),
and `linkage->pp_info[j]` may not correspond to `linkage->link_array[j]`
What did I miss?

linas:
Sorry for all of the confusion.  The pull request #177 should fix the numbering problem, it just remaps the pp_info as well.

I got side-tracked with other code, as I was hoping to re-organize how domain names are handled.  I got part-way into that, before I realized it was fairly pointless.

PR #162, opened 2015-08-02
Travis testing support

AmeBel:
Some issues that need to be resolved.
 - The tests don't give non-zero exit code as such the badge isn't useful for evaluating pull-requests
 - The bindings are not tested.

Not sure how to proceed. Any suggestions?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,11 @@ sudo: false
 
 cache: apt
 
+# For installing packages
+addons:
+    apt:
+        packages:
+            - autoconf-archive
 before_script:
     - ./autogen.sh --no-configure
     - mkdir build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+os:
+  - linux
+
+# Using the container infrastructure.
+sudo: false
+
+cache: apt
+
+before_script:
+    - ./autogen.sh --no-configure
+    - mkdir build
+    - cd build
+    - ../configure
+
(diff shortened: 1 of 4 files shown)


ampli:
Currently the Travis tests always fail with this:
```
WARNING: We were unable to find a .travis.yml file. This may not be what you
  want. Build will be run with default settings.
```
Until recently this failure didn't happen.
What is the reason for that?
Can I do something in this regard?

--
Regarding the proposed changes:
The 4.0.*batch files are not really good for this automatic test.
Instead, the following should be used:
```
../configure --with-python-bindings
...
make check
```

AmeBel:
I am getting the following error
```
../configure: line 16677: syntax error near unexpected token `LIBGC,'
../configure: line 16677: `	PKG_CHECK_MODULES(LIBGC, bdw-gc, gc=yes, gc=no)'
```

AmeBel:
Here is my dev-space https://github.com/AmeBel/link-grammar/blob/travis/docker/development/Dockerfile

ampli:
You need to install also **pkg-config**.

AmeBel:
@ampli updated as per your suggestion.

linas:
Not gonna happen.

ampli:
Can the travis invocation be removed?
It is just annoying to see the "all tests failed" messages and all of these red x's...

AmeBel:
@ampli it is removed.

PR #163, opened 2015-08-03
Spell-checker related fixes

ampli:
Changes:
1. Change the default configured speller to hunspell, because aspell is currently not thread safe.
(I actually observed failures due to that several times.)
2. Fix accumulated **memory leak** in spellcheck-hun.c.
3. Fix a compilation warning when no spell checker is configured.
4. Fix a **bug** in spell results which can stem-suffix split.
5. Add more languages to the aspell/hunspell hard coded configuration.
6. If the spell checker feature is configured in, but the spell checker
initialization fails from some reason (e.g. system misconfiguration or
missing spelling support for the language), currently no warning
is printed. Add: `Info: Spell checker disabled.`
Add TODOs regarding the above:
  - Set the spell option to 0, to signify no spell checking is done.
  - On verbosity >= 1, add a detailed message on the reason.

-7. Remove a TODO comment that has already implemented:
`/* TODO Change !spell to be an integer which will be this limit. */`

---
Should it be done (it's easy to implement):
Instead of hard coding  the list of languages and their suffix file names, add a definition in the **affix** file, e.g:
`en_US:  SPELLCHECK+;`

This has the following benefits:
- Easy to add/remove the spelling feature for existing/new languages.
- The user can select the variant of the spelling, e.g. en_GB instead of en_US.

Actually, I would implement a list, when the first one that is found is used (7 here is the number of guesses used, so even this would not be hard coded):
`en_US en-US 7: SPELLCHECK+;`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -340,25 +340,8 @@ AC_ARG_WITH([hunspell-dictdir], [AS_HELP_STRING([--with-hunspell-dictdir=DIR],
 
 # ====================================================================
 
-ASpellFound=no
-if test "$do_aspell" = yes ; then
-	PKG_CHECK_MODULES([ASPELL], [aspell], [ASpellFound=yes], [ASpellFound=no])
-	save_cpp_flags=${CPPFLAGS}
-	CPPFLAGS="${CPPFLAGS} ${ASPELL_CFLAGS}"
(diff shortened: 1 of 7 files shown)


linas:
sorry to take so long, swamped in emiail

PR #166, opened 2015-08-12
Python binding fixes

ampli:
This patch includes a major implementation change (in a backward compatible way to the current implementation) of the linkage iterator. It enables obtaining the number of linkages without pre-computing all the linkages in advance.

It passes tests.py, and I also tested it with a sentence parse interleaving (using 2 sentences from different dictionaries).


linas:
Thanks!

PR #172, opened 2015-08-15
tr: Restore transitive verb, rearrange connector order

ampli:
This is a try to improve the **tr** demo by examples of solving some of the problems.
But since I don't know Turkish I leave the rest for the original author.
I can work with him on that if he has questions.


linas:
I'm merging, cause I don't see why not. You should write to Can Bruce, you should be able to find his email address from the mailing ist.

PR #182, opened 2015-08-20
Fix memory leak in !!word which is not in the dict

ampli:


ampli:
This fix is of course buggy...
I will replace it shortly.

ampli:
Replaced by the correct fix.

Issue #183, opened 2015-08-22
Double crash: Error: Illegal domain:


ampli:
The current version aborts on the start of running en/4.0.batch and also on tests.py.
It prints:
link-grammar: Error: Illegal domain:
(Nothing is printed after the colon. Most probably it is '\0'.)
It then crashes in error.c:83, because of invalid alternatives array (I haven't tried yet to find out why).

The last good version seems to be c1f504ca0f51416e1338f6f84db2629253272c16.
It is from 6 days ago, so it is strange that I didn't notice it before. So I cleared the compilation cache, recompiled from scratch, and even rebooted the computer, but the problem persists.

ampli:
It happened due to this commit: 94042b799d7cea755d88c61a39501fa095b7a9d9

linas:
Hah!? but that commit does not "do anything", it renames a few routines, changes how args get passed.

ampli:
Not exactly.

Here is the problem:
``` 
	linkage->hpsg_pp_data = sent->constituent_pp->pp_data;
-	post_process_new_domain_array(sent->constituent_pp);
+	pp_new_domain_array(&linkage->hpsg_pp_data);
```
Note that `post_process_new_domain_array()/pp_new_domain_array()` "clobbers" the  pp_data by `malloc()`:
```
void pp_new_domain_array(Postprocessor *pp)
{
	PP_data *pp_data = &pp->pp_data;
	pp_data->domlen = PP_INITLEN;
	pp_data->domain_array = (Domain*) malloc(pp_data->domlen * sizeof(Domain));
	memset(pp_data->domain_array, 0, pp_data->domlen * sizeof(Domain));
}
```
When you introduced the change to pass `&linkage->hpsg_pp_data`, it then clobbers the pp_data that has just got assigned to it by:
```
	linkage->hpsg_pp_data = sent->constituent_pp->pp_data;
```
The result is that when`pp_data->domain_array` is needed for constituents processing,  it is already zeroed out, which triggers the said error of illegal domain.
Previously `pp_new_domain_array()` zeroed it out in `sent->constituent_pp` but it remained valid in `linkage->hpsg_pp_data`.

I actually verified it by doing this:
**1.** Checked out at the said commit (c1f504ca0f51416e1338f6f84db2629253272c16).
**2.** Compiled from scratch and verified that the fault existed:
```
echo 'Last dog I saw a great movie' | link-parser -const=1
...
link-grammar: Error: Illegal domain:
```
[...]

PR #184, opened 2015-08-22
constituents.c: Initialize err_ctxt before error messages.

ampli:
This fixes the crash when notifying domain errors.
You may want to apply it before starting debugging #183.


linas:
thank you...

PR #185, opened 2015-08-27
Fix "Error: Illegal domain" (issue #183)

ampli:
Fix allocation/free of constituent's pp_data,  to prevent double allocation and double free.

I must admit I don't completely understand some fine details in the allocation/free of these data structures.
However:
1. This fixes the problem without adding any  more overhead.
2. I checked it by a detailed run of all the linkages, with `!constituents=1`, of `en/4.0.batch`,
comparing it to a and older version of version just before we started to do the `pp_data` rearrangements (de7458b3b6fee2947f914573c95d9273af37a725). Both runs were done using the dict directory of the older version. There were absolutely no changes.
3. I made the said batch run with memory checks, to validate no use-after-free and that everything gets freed.

### DIFF
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -444,7 +444,6 @@ static void free_linkages(Sentence sent)
 		/* Note: linkage->hpsg_pp_data.domain_array originally got
 		 * allocated in sent->constituent_pp->pp_data. */
 		pp_free_domain_array(&linkage->hpsg_pp_data);
-		free(linkage->hpsg_pp_data.domain_array);
 
 		linkage_free_pp_info(linkage);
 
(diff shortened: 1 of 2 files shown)


linas:
I'm closing this, I did a very similar-but-different fix in pull req #186 which did what this does, as well as removing the crazy hpsg_pp_data structure, which was at the root of the confusion: it recorded conflicting evidence for these arrays.

linas:
thanks for identifyingthe core problem, though; without this patch, I wouldn't have seen the other problem.

linas:
Yeah, the allocation was insane. Its slightly better now, ... but it is confusing. It seems to have been designed for perforamnce, since the domain array gets re-used for different linkages, in a serialized way.

I did not do a full suite of tests.

ampli:
It passes tests.py (compiled with ASan/UBsan).

PR #187, opened 2015-09-12
Fix the SAT solver

ampli:
Fixes:
1. Correctly (hopefully) reconstruct the disjunct connectors.
2. Put the appropriate wordgraph pointer in each of the disjuncts. This is a must for the wordgraph code.
3. Use the X_node strings (subscripted words) as the disjunct words.
4,. Discard empty words. This is also a must for the wordgraph code.
5. Improve the related Link-parser messages. Previously it was "too silent". Messages added:
- No linkages found.
- No more linkages.

Not as usual, I didn't clean the commits too much, so they contain intermediate versions of the changes (for example, reconstructing the disjuncts from the provided right/left expressions, a thing that proved to be a wrong idea).

Many problems still remained unresolved. For some I know the fix or a work round.
Some of the problems below are mentioned in the last commit, but I added more here.
- It crashes on !links, because the domain array doesn't exist. I need an advice on that.
- It doesn't work on Hebrew at all (but works on Russian).
- Batch runs are mostly meaningless, because the sat-solver code
  returns num-linkages=222 even when there are no linkages.
- The costs shown by !disjuncts are not always the same as those of the
  standard parser.
- In the cost vector, it always shows a disjunct costs of 0.
- It still checks alternative[0] in in guiding.cpp. It can be fixed, by much work, to check the wordgraph. However, if it is done only for solving speedup hints, it may not be urgent to improve it.
- It doesn't free all memory, and especially not the connectors.   This problem always existed.
It doesn't parse with null words. It seems possible and easy enough, so I don't know why it has not been done.
- The post-parsing code could be more similar to that of the standard parser. It can be rewritten so more code can be shared.

I guess there are even more problems. I will try to fix some of them.

EDIT:
I appended some more fixes. Please see #188 for a summary.

### DIFF
--- a/link-grammar/build-disjuncts.c
+++ b/link-grammar/build-disjuncts.c
@@ -497,7 +497,7 @@ void build_sentence_disjuncts(Sentence sent, double cost_cutoff)
 		for (x = sent->word[w].x; x != NULL; x = x->next)
 		{
 			Disjunct *dx = build_disjuncts_for_exp(x->exp, x->string, cost_cutoff);
-			word_record_in_disjunct(x, dx);
+			word_record_in_disjunct(x->word, dx);
 			d = catenate_disjuncts(dx, d);
 		}
 		sent->word[w].d = d;
--- a/link-grammar/build-disjuncts.c
+++ b/link-grammar/build-disjuncts.c
@@ -347,6 +347,41 @@ void prt_exp(Exp *e, int i)
 		printf("con=%s\n", e->u.string);
 	}
 }
+
+void prt_exp_mem(Exp *e, int i)
+{
+	const char *type;
+
+	if (e == NULL) return;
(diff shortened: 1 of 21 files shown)


link-grammar/build-disjuncts.c
```diff
@@ -347,6 +347,41 @@ void prt_exp(Exp *e, int i)
 		printf("con=%s\n", e->u.string);
 	}
 }
+
+void prt_exp_mem(Exp *e, int i)
```
linas: I believe that there already is a print function that is more-or-less the same as this, somewhere else in the code...

linas: never mind.  ...

linas:
sorry for the long wait.

PR #189, opened 2015-09-27
More SAT-solver commits

ampli:
Mosty comments.
Also a call to setActivity() has been commented out.

### DIFF
--- a/link-grammar/sat-solver/guiding.hpp
+++ b/link-grammar/sat-solver/guiding.hpp
@@ -109,7 +109,7 @@ protected:
     _parameters[var].priority = priority;
     _parameters[var].polarity = polarity;
 
-    //    printf("Set: %d %s .%g. .%g.\n", var, isDecision ? "true" : "false", priority, polarity);
+    //printf("Set: %zu %s .%g. .%g.\n", var, isDecision ? "true" : "false", priority, polarity);
   }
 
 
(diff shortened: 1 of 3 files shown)


linas:
Thanks!

So, is minisat 2.2 faster than the copy we've snapshoted?  Or about the same?

ampli:
On first tests of minisat 2.2 (actually 2.2+, I took the github version) it was significantly slower
(comparing to the current LG minisat version with setActivity() removed).. 
However, after a slight tweak its speed got to be exactly the same.

The tweak was in calling solver->setPolarity() (file guiding.hpp). I changed the `>0` to be `==0`.
This sped it up by 20%!

Now I'm the a process of trying it with Glucose 4.0.
I also would like to test it with CryptoMinisat, first without a modification.

CryptoMinisat also seems to have a feature that potentially could be helpful for LG -
it supports clauses with XOR instead of OR. But sat-solver needs some rewrite to use that.

From what I have read on SAT solvers, and from looking at the current LG sat-solver code, I think there is a potential of speeding LG SAT solving by several times, maybe even an order of magnitude or more.

BTW, the current LG SAT solver code has a major flaw regarding cost of null expression, that I would like to fix. There are other things that can be improved. I guess it is better to open an issue for that.

linas:
Huh. OK. Thanks!

PR #197, opened 2015-10-25
fast-matcher: Speed-up removing duplicates

ampli:
About 15% speed-up when processing en/4.0.fixes-batch.


linas:
Wow! Excellent!

linas:
I mean, I fiddled with this, and implemented a bad solution; yous is much better, so, really, I'm impressed!

PR #202, opened 2015-11-02
append_string(): Fix two bugs

ampli:
For a background, see:
[The trouble with snprintf](https://securityblog.redhat.com/2014/03/12/the-trouble-with-snprintf/)

1. A possibility to return a negative value (on error) is documented in
the manual. If from some reason an error occurs, a fatal malfunction will
result.  Check for a possible error.

2. If the vsnprintf() buffer is insufficient, the print result gets
silently truncated to the buffer size (with a proper NUL termination).
This may be considered a problem, as the program output is then
intermixed in the middle of a line with further output from other
sections of the program.
In addition (from the manual): "If  the  output was truncated  due to
this  limit, then the return value is the number of characters
(excluding the terminating null byte) which would have been written to
the final string if enough space had been available."
But the buffer is currently extended by this returned value. Hence
output from subsequent append_string() calls will be hidden due to the
resulted NUL in the middle of the buffer. This insufficient buffer
problem is not fatal for ISO C99 vsnprintf() (as currently in GNU C),
which doesn't return -1 on an insufficient buffer (else it is, see (1)
above).

In the same occasion, simplify the function somewhat.

### DIFF
--- a/link-grammar/print-util.c
+++ b/link-grammar/print-util.c
@@ -60,10 +60,9 @@ char * string_copy(String *s)
 void append_string(String * string, const char *fmt, ...)
 {
 #define TMPLEN 1024
-	char temp_string[TMPLEN];
+	char temp_buffer[TMPLEN];
+	char * temp_string = temp_buffer;
 	size_t templen;
-	char * p;
-	size_t new_size;
 	va_list args;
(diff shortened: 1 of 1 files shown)


linas:
Thanks!

Issue #201, opened 2015-10-29
Argument "dir" of match_in_connector_set()


ampli:
Here is the code of this funtion:
```
bool match_in_connector_set(Connector_set *conset, Connector * c, int dir)
{
	size_t h;
	Connector * c1;
	if (conset == NULL) return false;
	h = connector_set_hash(conset, c->string, dir);
	for (c1 = conset->hash_table[h]; c1 != NULL; c1 = c1->next)
	{
		if (easy_match(c1->string, c->string) && (dir == c1->word)) return true; <---
	}
	return false;
}
```
Note the line marked with **<---**.
Since this function appears as 1.4% in the profiling, I tried to speed it up by reversing the order of the **&&** check. However, it only increased its CPU usage. It turns out `(dir == c1->word)` is always true, because this function is now only invoked with **dir=='+'**.

The only place in which it got invoked with **dir=='-'** was in the file **and.c**, which has gone with the FAT-LINKS code.
So I guess it is OK to simplify it by removing the **dir** argument from it.

linas:
yes, seems right to me; convert this to a pull request...

PR #203, opened 2015-11-04
match_in_connector_set(): Remove the "dir" parameter

ampli:
(Issue #201.)

Notes:
connector_set_hash() still has a "dir" parameter.
However, it is useless, since match_in_connector_set() calls connector_set_hash() only with dir=='+'. Hence the code in build_connector_set_from_expression() which allows UNLIMITED-CONNECTORS with "-" is useless too.

So can "dir" be omitted also from connector_set_hash() and build_connector_set_from_expression()?

### DIFF
--- a/link-grammar/word-utils.c
+++ b/link-grammar/word-utils.c
@@ -294,15 +294,15 @@ void connector_set_delete(Connector_set * conset)
  * d='-' means this connector is on the left side of the disjunct.
  */
 
-bool match_in_connector_set(Connector_set *conset, Connector * c, int dir)
+bool match_in_connector_set(Connector_set *conset, Connector * c)
 {
 	unsigned int h;
 	Connector * c1;
 	if (conset == NULL) return false;
-	h = connector_set_hash(conset, c->string, dir);
+	h = connector_set_hash(conset, c->string, '+');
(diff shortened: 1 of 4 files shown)


link-grammar/word-utils.c
```diff
@@ -294,15 +294,15 @@ void connector_set_delete(Connector_set * conset)
...
  */
 
-bool match_in_connector_set(Connector_set *conset, Connector * c, int dir)
+bool match_in_connector_set(Connector_set *conset, Connector * c)
 {
 	unsigned int h;
 	Connector * c1;
 	if (conset == NULL) return false;
-	h = connector_set_hash(conset, c->string, dir);
+	h = connector_set_hash(conset, c->string, '+');
```
linas: is connector_set_hash ever called with dir='-' ?

ampli:
In build_connector_set_from_expression(), it has the potential to be called with dir='-', because it is called on UNLIMITED_CONNECTORS_WORD from the dict (by connector_set_create()). However, this "word" doesn't contain '-' connectors.

Is there still a need to allow '-' connectors in UNLIMITED_CONNECTORS_WORD?
But I guess this is not a too important issue since it doesn't incur significant overhead.

PR #204, opened 2015-11-04
init_connector() improvements

ampli:
- Make inline.
- Use in connector_new() in order to share code.

### DIFF
--- a/link-grammar/word-utils.c
+++ b/link-grammar/word-utils.c
@@ -207,19 +207,11 @@ void free_connectors(Connector *e)
 Connector * connector_new(void)
 {
 	Connector *c = (Connector *) xalloc(sizeof(Connector));
-	c->length_limit = UNLIMITED_LEN;
+	init_connector(c);
 	c->string = "";
-	c->hash = -1;
 	c->multi = false;
 	c->next = NULL;
 	c->tableNext = NULL;
-	return c;
-}
-
-Connector * init_connector(Connector *c)
-{
-	c->hash = -1;
-	c->length_limit = UNLIMITED_LEN;
 	return c;
 }
 
--- a/link-grammar/word-utils.h
+++ b/link-grammar/word-utils.h
@@ -34,8 +34,14 @@ void free_X_nodes(X_node *);
 
 /* Connector utilities ... */
 Connector * connector_new(void);
-Connector * init_connector(Connector *c);
 void free_connectors(Connector *);
+
+static inline Connector * init_connector(Connector *c)
+{
+	c->hash = -1;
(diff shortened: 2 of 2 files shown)


link-grammar/prune.c
```diff
@@ -573,14 +573,14 @@ static inline bool matches_S(connector_table *ct, Connector * c, char dir)
  */
 static int mark_dead_connectors(connector_table *ct, Exp * e, char dir)
 {
-	int count;
-	count = 0;
+	int count = 0;
+	Connector dummy;
+	init_connector(&dummy);
```
linas: why? you are changing the scope -- in general it is best to keep the scope closest to the location where it is actually used.   If dummy is not used anywhere else, then initialize it only in the block where it is used.

ampli: You are right. This was outlandish, including the comment (no loop there). It even increases the overhead, as it performes the initialization even when it has no use.
The only  way to eliminate this initialization is to restructure the code to use the connector directly.

I tried to touch this code because mark_dead_connectors() is in the top-10 hot functions.

So I can resend this pull request without this commit.

linas: You don't need a new pull request; just make changes and push to this same branch, and the new changes automatically become part of this pull request.

The compiler, in theory, is supposed to be able to make some of this optimization on its own, but its not clear just how smart the compiler actually is...

ampli: At least gcc5.1.1 (that I use) is very smart. However, init_connector() is a function in another file and it doesn't know, by default, to inline it, so this pull request makes it inline.

BTW, **gcc**, for some time now (gcc4.8) knows to do link-time optimizations (-flto) , including global inlining. I tested it and it produces a library which is much smaller (I don't know yet why, I would expect it to be bigger due to the inlining).
However, when I tested it I couldn't find evidences for an improved speed - the variation between runs was too high for a meaningful comparisons by a few runs. But at that time I didn't use yet my current method of averaging 200 runs, so I will eventually repeat this test.

I will resend to the same branch (but after you incorporate the length-limit pull request, which I have just also resent to the same branch - in order  to prevent a possible conflict again).

linas: I'm not surprised that LTO makes no difference -- everything that's been important has already been optimized by hand.   The smaller binary is surprising.   I'm guessing that maybe some of the debuggiung symbols are no longer being generated, or somehow got simplified.

LTO can maybe ever-so-slightly reduce binary sizes, by removing some of the subroutine glue linkage.  Maybe some symbol table sizes can decrease in size...

ampli:
Resent without the bogus change.

linas:
thanks

PR #205, opened 2015-11-06
Fix installation of fa words

hedayat:


linas:
thanks!  Its sort of disappointing how no one noticed this before. :-(

linas:
tell me how well it works, if you get the chance.   .. yes I know you have to romanize, and that's a lot of work; its not a standard romanization.

hedayat:
I've tried, but actually I don't know how well it works. If [?] is a bad sign, it probably needs more work!
But fortunately, there is no need to romanize, specially with the weird format it has. Using `morphology/fa/scripts/utf8_2_roman_1-7.pl` it is possible to convert UTF-8 Persian text to the special format needed for the rest of the code:

    % echo "من مشغول هستم." | scripts/utf8_2_roman_1-7.pl
    Use of the encoding pragma is deprecated at scripts/utf8_2_roman_1-7.pl line 10.
    mn mCGul hstm.

A sample output:
------------------------------------
Found 2 linkages (2 had no P.P. violations) at null count 1
	Linkage 1, cost vector = (UNUSED=1 DIS= 1.00 LEN=13)

    +-------------------Xp------------------+
    +-----------------Wi----------------+   |
    |       +------------Spn1-----------+   |
    |       |              +-----VMP----+   |
    |       |              +--VMT-+     |   |
    |       |              |      |     |   |
LEFT-WALL mn.pn [mCGul] hs[?].v t.vmt m.vmp . 

linkparser> 	Linkage 2, cost vector = (UNUSED=1 DIS= 2.00 LEN=9)

    +-------------------Xp------------------+
    |       +------------Spn1-----------+   |
    |       |              +-----VMP----+   |
    +---Wd--+              +--VMT-+     |   |
    |       |              |      |     |   |
LEFT-WALL mn.pn [mCGul] hs[?].v t.vmt m.vmp .

linas:
Huh, OK. [?] is good, it means the guessing algo is running. The  [mCGul] is bad, it means teh word was skipped. In principle, the dictionaries can now be converted to utf8, so that the perl script would not be needed. I don't know how much work this would be; I think the perl script also does morphology analysis.

hedayat:
AFAIK, there is another perl script, morphology/fa/stemmer.pl, which does the analisys. To get the final output, I gave the output of the utf8 converter to this script, and give its output to link-parser. 

Actually, there is a `persianparse.sh` script inside morphology/fa which runs both stemmer & link parser. I've created another script, `utf8_persianparse.sh`, which runs utf8 converter, stemmer & link parser so that it can parse a UTF8 sentence with one command.

If [?] is good, so the results are more promissing! :)

linas:
OK. The LG parser itself has changed a LOT since the persian dicts were
written. It can now handle utf8 and do simple stemming/morphology too.

--linas

PR #207, opened 2015-11-08
Remove "dir" from matches_S()

ampli:
... as it is not needed any more.
Also remove prune_match(), which remained from the FAT-LINKS code,
and directly use easy_match().


linas:
thanks!

PR #208, opened 2015-11-08
Improvements in the setting of length_limit

ampli:
- In the regular and SAT parsers:
  Avoid a very large number of calls (a call per connector) to the API
  call parse_options_get_all_short_connectors().  Instead, use
  opts->all_short.

- Simplify set_connector_list_length_limit() and in a similar way
  set_connector_length_limit().

- Use unsigned int instead of size_t for the temporary variable that is
  set to opts->short_length.

### DIFF
--- a/link-grammar/preparation.c
+++ b/link-grammar/preparation.c
@@ -29,10 +29,8 @@ set_connector_list_length_limit(Connector *c,
                                 Parse_Options opts)
 {
 	for (; c!=NULL; c=c->next) {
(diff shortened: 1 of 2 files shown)


linas:
there are conflicts in this branch, it can't be merged; can you resolve?

ampli:
I guess this is because my long queue of changes, that may touch the same areas.
I have so many active branches that I hardly remember what is in each...
I will resolve it and send again.

ampli:
I sent it again just now.

PR #211, opened 2015-11-25
String-set comparisons

ampli:
Per issue #306.
Add infrastructure and convert connector comparisons to use it.

### DIFF
--- a/link-grammar/string-set.h
+++ b/link-grammar/string-set.h
@@ -14,6 +14,7 @@
 
 #include <stddef.h>
 #include "api-types.h"
+#include "utilities.h"
 
 struct String_set_s
 {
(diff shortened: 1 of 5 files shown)


link-grammar/string-set.h
```diff
@@ -27,4 +28,28 @@ const char * string_set_add(const char * source_string, String_set * ss);
 const char * string_set_lookup(const char * source_string, String_set * ss);
 void         string_set_delete(String_set *ss);
 
+/**
+ * Compare 2 strings, assuming they are in the same string-set.
+ * Return true if they are equal, else false.
```
linas: Please add a note explaining the "obvious": two string-set strings are equal if and only if their pointers are equal.  Otherwise, one foggy-headed day in the future, the == compare will be very  ... confusing.

PR #212, opened 2015-11-25
string_set_cmp(): Improve the function description.

ampli:
Two string-set strings are equal if and only if their pointers are equal.


linas:
thanks!

PR #213, opened 2015-11-26
Connector length-limit setting improvments

ampli:
Note the comment about ZZZ in the first commit.
I will open an issue to discuss that.

(Opened #214.)

### DIFF
--- a/link-grammar/sat-solver/word-tag.hpp
+++ b/link-grammar/sat-solver/word-tag.hpp
@@ -139,10 +139,9 @@ public:
 
     if (len > UNLIMITED_LEN) len = UNLIMITED_LEN;
 
-    if (!_opts->all_short &&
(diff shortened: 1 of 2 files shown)


link-grammar/sat-solver/word-tag.hpp
```diff
@@ -139,10 +139,9 @@ class WordTag
 
     if (len > UNLIMITED_LEN) len = UNLIMITED_LEN;
 
-    if (!_opts->all_short &&
-        (conset == NULL || match_in_connector_set(conset, c))) {
-      c->length_limit = UNLIMITED_LEN;
-    } else {
+    if (_opts->all_short ||
+        !(conset == NULL || match_in_connector_set(conset, c)))
```
linas: can you change this ti `(conset != NULL && !match_in_connector_set(conset, c))`  that would make it slightly easier to read ...

PR #215, opened 2015-11-26
Make the condition to set length_limit clearer.

ampli:


ampli:
It turns out my last patch to set_connector_length_limit() of the SAT-solver broke it, because somehow length_limit is not UNLIMITED_LEN by default. I will check why (later). meanwhile I changed it back in this pull request.

PR #221, opened 2015-11-29
configure: Enable compiling using -std=c11

ampli:
On MinGW it remained gnu99, and may need (or not) to be changed too.


linas:
My impression is no one is using LG+minGW

Issue #227, opened 2015-12-05
Internal error !


linas:
There's an internal error:
```
cat ../data/en/corpus-fixes.batch | link-parser > x
```
prints
```
link-grammar: Error: Word 'Sui': Internal error: NULL X_node
link-grammar: Error: Word 'Zeus': Internal error: NULL X_node
link-grammar: Error: sentence_split(): Internal error detected
link-grammar: Error: Word 'vibes': Internal error: NULL X_node
link-grammar: Error: sentence_split(): Internal error detected
```

ampli:
These words are not found (even as substrings) in the released  en/corpus-fixes.batch, so I cannot identify the sentences which caused this problem.

Can you please run it with errors redirected to the same file, and then scan the file for "NULL X_node"?
Something like:
```
$ cat ../data/en/corpus-fixes.batch | link-parser >  x   2>&1
$ grep -B 1 'NULL X_node'  x
```

I already know about such kind of error, involving spelling and contractions:
```
linkparser> this test,'s,
link-grammar: Error: Word 'Estes': Internal error: NULL X_node
```
Meanwhile I will try to fix this one and send a pool request.

BTW, this "NULL X_node" internal error means that the tokenizer marked a word as "in dict", but when constructing the X-Nodes, the word is actually not found in the dictionary.

linas:
Ah, interesting, its the spell checker:
```
*Синтезируется из глюкозы в тканях и органах, например в сердце, печени, почках.
*You need it most.Either way you cant lose.
spell set to 4
*lgdebug(+D_IWA, "(%s) Gword %zu:%s split (split_counter=%zu) into", label,
link-grammar: Error: Word 'Sui': Internal error: NULL X_node

link-grammar: Error: Word 'Zeus': Internal error: NULL X_node

link-grammar: Error: sentence_split(): Internal error detected
*@s
vibeix here
link-grammar: Error: Word 'vibes': Internal error: NULL X_node

link-grammar: Error: sentence_split(): Internal error detected
spell set to 0
*"T" This is a test
*"To" It's to let; there's a notice on the door
```

linas:
Well, adding the words 'vibes' and 'Zeus' to the dict make the corresponding errors go away.  Adding 'sui_generis' does not.

I dunno, so this is a dictionary - spellcheck mismatch. I'm tempted to do nothing ...

ampli:
It looks as if these particular cases have been already fixed (and hence these sentences exist in the batch).

[...]

PR #232, opened 2015-12-05
Don't accept a contracted word which is not in the dict as spell corr…

ampli:


linas:
Thanks, I guess this is a fix for #227

PR #235, opened 2015-12-07
Change the "s" wordgraph display to make a more readable subgraph

ampli:
... but less compact.

Usage:
link-parser -test=wg:s

test wg s

The previous (more compact on the X axis) display is still possible by:
link-parser -test=wg:sc

test wg sc


linas:
yes thanks, the linear ordering is less confusing

PR #239, opened 2015-12-08
Fix assert  strip_right: unexpected empty word

ampli:
It was due to allowing a null-string "stem", which is not supported by the rest of the code.
So an empty string entered into the wordgraph (maybe we need an assert on that).
When separate_word() encountered this newly inserted empty string word, it aborted in strip_right() on this assert(), that was originally intended to guard against a miscalculation of the word end before calling the function.

In the same occasion I found and fixed another bug, which caused mis-displaying of linkage null-words which happen to be prefixes, when INFIX_MARK is not defined (their last character, in our case `/` or `'`, got stripped from the displayed word, as if it was an INFIX_MARK).

### DIFF
--- a/link-grammar/linkage.c
+++ b/link-grammar/linkage.c
@@ -43,24 +43,25 @@
  * Append an unmarked (i.e. without INFIXMARK) morpheme to join_buff.
  * join_buff is a zeroed-out buffer which has enough room for morpheme to be
  * added + terminating NUL.
- * In case INFIXMARK is not defined in the affix file, it is '\0'. However,
- * in that case there are no MT_PREFIX, MT_SUFFIX and MT_MIDDLE morpheme types.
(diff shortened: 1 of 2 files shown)


linas:
Thanks. re ZZZ I've seen a semi-rare bug, where I turn on disjunct printing !disj  and every now and then a ZZZ shows up in one of them, even though there's no ZZZ in the printed linkage. Never looked at it carefully.

ampli:
Indeed.
There is currently no code to remove ZZZ from the disjuncts. The current code only removes the ZZZ links. If desired, I can add code to remove the ZZZ links from the disjuncts too.

linas:
That would be nice, not urgent. One trick is to accumulkate statistics, see how often each disjunct gets used. This gives a statistical snapshot of what LG is actually doing ...  I've put that to some use, but not as much as planned (i.e. be much more statistically driven in the dict contents.)

Issue #244, opened 2015-12-11
Parser recursive calls and my new match fields in Disjunct_struct


ampli:
Hi Linas,
I implemented the idea to move connector matching to form_match_list() and use 2 new fields in the Disjunct_struct to indicate left/righ connector matching, as I mentioned in #219.
```
struct Disjunct_struct
{
    Disjunct *next;
    const char * string;       /* subscripted dictionary word */
    Connector *left, *right;
    double cost;
    bool marked;               /* unmarked disjuncts get deleted */
    bool match_left, match_right; /* used only during parsing */       <--- Added
    const Gword **word;        /* NULL terminated list of originating words */
};
```
However, I have a question regarding the safety of this idea.

My said implementation allows several shortcuts in connector matching, and speeds up en/corpus-fix-long.batch by about 40%.  To be sure it doesn't introduce any problem, I checked the detailed run of en/corpus-basic.batch, en/corpus-fixes.batch, ru/corpus-basic.batch and also he/corpus-basic.batch with limit=10000, and the results were totally identical to those of the current master branch.

However, due to the recursive working of the parser, new form_match_lists() calls are issued while disjunct list from previous ones have not been completely handled. So if the same disjuncts are **ever** happened to be touched, havoc may happen, as the said match fields would get overwrite, possibly with different results.

So I used a debug check to validate that the match list disjunct fields are not modified in the recursive calls. And indeed they are never modified by these calls. I don't yet understand the parsing algorithm, but from inspection of the code it seems this is a property of the algorithm that the same `w` is not used again in the recursive calls.

**Question**:  Can we depend on this to always be true?

linas:
Heh.  There are three ways to answer this question:
1)  for you to fully understand the parse algo.
2) to perform a statistically test/analysis
3) for me to fully understand the algo, and for you to ask me.

Seems like you opted for 2 & 3. :-)  For me to answer 3), I would have to look what you did, study and ponder for at least a few hours or a day.  Which is time I don't have right now.  My intuition is that your study (2) is sufficient, assuming you did not make any mistakes.

Perhaps it is time to take a shot at (1)  Its actually not that hard, *if* you read the paper(s) first.  They present pseudocode which is relatively easy to understand (easier than the code; they actually explain it), and you can then compare with the actual C code, and it matches up well.  

[...]

PR #247, opened 2015-12-13
Additional significant speedup of long sentences

ampli:
See the discussion at #244.

Speed up by these changes (comparing to the current master branch):

ru/corpus-fixes.batch: ~0%
en/corpus-basic.batch: ~0%
en/corpus-fixes.batch: ~10%
en/corpus-fix-long.batch [not including 3 last sentences, to shorten the benchmark]: 40+%

I tried it also on the following sentence which was not included above - it is currently the last one in en/corpus-fix-long.batch, and the result is consistent with the speed up as indicated above:

And yet he should be always ready to have a perfectly terrible scene, whenever we want one, and to become miserable, absolutely miserable, at a moment’s notice, and to overwhelm us with just reproaches in less than twenty minutes, and to be positively violent at the end of half an hour, and to leave us for ever at a quarter to eight, when we have to go and dress for dinner when, after that, one has seen him for really the last time, and he has refused to take back the little things he has given one, and promised never to communicate with one again, or to write one any foolish letters, he should be perfectly broken-hearted, and telegraph to one all day long, and send one little notes every half-hour by a private hansom, and dine quite alone at the club, so that every one should know how unhappy he was.

Results:
Current: 3966.762u 3.320s 1:06:16.47 99.8%  0+0k 192+128io 0pf+0w
After this modification: 2235.120u 2.471s 37:21.20 99.8%  0+0k 200+16io 8pf+0w
(Speed up of ~44%.)

A detailed run with **-limit=10000** of en/corpus-basic.batch, en/corpus-fixes.batch, ru/corpus-basic.batch and he/corpus-basic.batch, with the library before and after this modification gives **exactly the same results**, modulo order changes of linkages with the same cost - if any, which is not important and which my check is not sensitive to. At least with the sentences tested in the test suite, this order is preserved too.

### DIFF
--- a/link-grammar/count.c
+++ b/link-grammar/count.c
@@ -93,19 +93,6 @@ static void init_table(count_context_t *ctxt, size_t sent_len)
 	memset(ctxt->table, 0, ctxt->table_size*sizeof(Table_connector*));
 }
 
-/*
- * Returns TRUE if s and t match according to the connector matching
- * rules.
- */
-bool do_match(Connector *a, Connector *b, int aw, int bw)
-{
-	int dist = bw - aw;
-	if (NULL == a || NULL == b) return false;
-	assert(aw < bw, "do_match() did not receive params in the natural order.");
-	if (dist > a->length_limit || dist > b->length_limit) return false;
-	return easy_match(a->string, b->string);
-}
-
 /**
  * Stores the value in the table.  Assumes it's not already there.
  */
--- a/link-grammar/count.c
+++ b/link-grammar/count.c
@@ -286,10 +286,16 @@ static Count_bin do_count(fast_matcher_t *mchxt,
 	{
 		Match_node *m, *m1;
 		m1 = m = form_match_list(mchxt, w, le, lw, re, rw);
+#ifdef VERIFY_MATCH_LIST
+		int id = m ? m->d->match_id : 0;
+#endif
 		for (; m != NULL; m = m->next)
 		{
 			unsigned int lnull_cnt, rnull_cnt;
 			Disjunct * d = m->d;
+#ifdef VERIFY_MATCH_LIST
+			assert(id == d->match_id, "Modified id (%d!=%d)\n", id, d->match_id);
+#endif
 			bool Lmatch = d->match_left;
 			bool Rmatch = d->match_right;
 
--- a/link-grammar/count.c
+++ b/link-grammar/count.c
@@ -303,8 +303,8 @@ static Count_bin do_count(fast_matcher_t *mchxt,
 		{
 			unsigned int lnull_cnt, rnull_cnt;
 			Disjunct * d = m->d;
-			bool Lmatch = do_match(le, d->left, lw, w);
-			bool Rmatch = do_match(d->right, re, w, rw);
+			bool Lmatch = d->match_left;
+			bool Rmatch = d->match_right;
 
 			/* _p1 avoids a gcc warning about unsafe loop opt */
 			unsigned int null_count_p1 = null_count + 1;
--- a/link-grammar/count.c
+++ b/link-grammar/count.c
@@ -100,6 +100,7 @@ static void init_table(count_context_t *ctxt, size_t sent_len)
 bool do_match(Connector *a, Connector *b, int aw, int bw)
 {
 	int dist = bw - aw;
+	if (NULL == a || NULL == b) return false;
 	assert(aw < bw, "do_match() did not receive params in the natural order.");
 	if (dist > a->length_limit || dist > b->length_limit) return false;
 	return easy_match(a->string, b->string);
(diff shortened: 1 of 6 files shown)


link-grammar/count.c
```diff
@@ -318,10 +314,6 @@ static Count_bin do_count(fast_matcher_t *mchxt,
 
 				/* Now, we determine if (based on table only) we can see that
 				   the current range is not parsable. */
-				Lmatch = (le != NULL) && (d->left != NULL) &&
-				         do_match(le, d->left, lw, w);
-				Rmatch = (d->right != NULL) && (re != NULL) &&
-				         do_match(d->right, re, w, rw);
```
linas: interting. Just pulling this out of the `lnull_cnt` loop, that alone might account for a big part of the speedup.

link-grammar/fast-match.c
```diff
@@ -203,6 +204,59 @@ static Match_node * add_to_left_table_list(Match_node * m, Match_node * l)
...
+	if (string_set_cmp(c1->string, c2->string)) return true;
+	if (c1->hash != c2->hash) return false;
+	if (c1->uc_length != c1->uc_length) return false;
+
+	const char *uc1 = &c1->string[c1->uc_start];
+	const char *uc2 = &c2->string[c2->uc_start];
+	/* We arrive here for less than 50% of the cases for "en" and
+	 * less then 20% of the cases for "ru", and the following
+	 * condition is always true.
+	 * How come 2 different uc strings never have the same hash value? */
```
linas: You have a terrible hashing algo.

There are about 100 or so different UC connectors in English. With a good hash, that should fit just fine into a single byte (256 points).  Should fit trivially into 32 bits, as none of the connectors are longer than 3 UC letters.  Most are 1 or 2 letters, and only a couple are 3 UC letters long, so if your hash can't fit them into 65536 slots, you've got a terrible hash.  I mean, you could just hand-assemble them ..
```
int16 mask (char c) { return  (c-0x40)  & 0x20; }
int16 hash = mask(s[0]) | (mask(s[1])<<5) | mask(s[2]) << 10;
```
the above should give you a perfect hash for the English dicts, since all connectors are 3 UC letters long or less. thus the strcmp would never-ever happen.

linas: Oh, interesting.  I forgot about the ID and the LL which are 5-6 UC letters long. I've got an idea, let me try that.

link-grammar/count.c
ampli: Note that the reported speedup was measured on batch runs, and in batch mode `lnull_cnt` is supposed to always be **0**, so the loop is executed only once, and pulling it out of the loop cannot contribute to the speedup.
However, when I tested this pulling out on "detailed batch run" (producing all the linkages, as if it was interactive run) I then observed a speed up which I attributed to this statement move (as sentences which don't get parsed with null count 0 are tried again with lnull_cnt = sent_length).

link-grammar/fast-match.c
linas: Oh, I misunderstood your note.  If its a good hash algo, then two different UC connectors should never have the same hash value; the fact that they actually don't is a good thing!    So, can you remove that question?

ampli: Of course I can.
But I still just **don't understand** how come it is so perfect, as the number of slots in the hash table is much less than the number of different connector UC. It looks that something must be wrong, or otherwise something is very tricky.

In any case it is desired to know the exact answer...

ampli: I mentioned "slots" in my message above, but of course this doesn't have anything to do with the number of slots in the hash table, since I compare the raw hash values, not after masking them with the table size.
So different hash values of several K of strings over 64K values is actually making sense...
And it seems that in any case the check is needed since there is no guarantee that an added different UC will not get an already-used value.

I will wait to hear all your comment about this proposed speedup and then will add commits as needed,

link-grammar/fast-match.c
```diff
@@ -203,6 +204,59 @@ static Match_node * add_to_left_table_list(Match_node * m, Match_node * l)
...
+{
+		unsigned int h, s;
+
+		s = h = connector_hash(c) & (size-1);
+
+		if (dir == 1) {
+			while (NULL != t[h])
+			{
+				if (con_uc_eq(t[h]->d->right, c)) break;
+				h = (h + 1) & (size-1);
```
linas: this is  interesting and unusual.  Confused me at first; I see that it works, and that its needed.  Could you add a note that says "every hash bucket MUST have a unique upper-case connector, since easy_match_list only compares the lower-case parts".

linas:
thanks!

link-grammar/fast-match.c
linas: well, the other thing to note here is that although you are storing 16 bits of hash, there are NOT several K of UC strings -- there are only 5 or 10 of these strings -- maybe dozens at most -- since you are storing only those connectors that actually happen for some fixed word in the sentence, and typically, there simply won't be very many on any given word .. so you are hashing 5 or ten things into a space of 65536 so a hash collision would be very very unlikely.

link-grammar/count.c
linas: ah OK

link-grammar/fast-match.c
```diff
@@ -276,64 +334,262 @@ fast_matcher_t* alloc_fast_matcher(const Sentence sent)
...
+ * This function uses shortcuts to speed up the comparison.
+ * Especially, we know that the uc parts of the connectors are the same,
+ * because we fetch the matching lists according to the uc part or the
+ * connectors to be matched. So the uc parts are not checked here. The
+ * head/dependent indicators are in the caller function, and only when
+ * connectors match here, to save CPU when the connectors don't match
+ * otherwise. This is because h/d mismatch is rare.
+ * FIXME: Use connector enumeration.
+ */
+static bool easy_match_list(Connector *c1, Connector *c2)
```
linas: could you rename this to "match_lower_case", and change the first two setneces of description above to say "Compare only the lower-case parts of two conectors. When this function is called, we already know that the upper-case parts of the connectors match" ... err, never mind, I will make this change myself right now ...

link-grammar/fast-match.c
linas: never mind I'm doing this now.

PR #253, opened 2015-12-15
Prune length_limit shortcut + setting ZZZ length_limit=1

ampli:
These fixes were validated by a detailed batch runs of this version and the version before introducing the pruning length_limit shortcut (which is fixed here). 
There were no diff for the English basic and fixes batch.
I also compared a detailed parse of the sentence:
```
In vivo studies of the activity of four of the kinases, KinA, KinC, KinD (ykvD) and KinE (ykrQ), using abrB transcription as an indicator of Spo0A~P level, revealed that KinC and KinD were responsible for Spo0A~P production during the exponential phase of growth in the absence of KinA and KinB.
```
It was identical, and took about 50% less CPU in this fixed version.

In Russian there were diff in a few sentences, all of them with null count > 0.
My guess is that previously empty words could get linked with distance>1 in that case (I observed such a thing in a recent check), which causes printing of extra identical partial linkages.

This length_limit shortcut doesn't solve the whole pruning slowness that got introduced with the empty word. I continue to work on that problem (apparently in expression pruning).

BTW, it is possible to validate that the ZZZ length_limit=1 "does something" by observing the **power prune cost** message with -verbosity=3. Setting it for short sentences (< short_lenght), for which there was no need to set the length_limit otherwise, actually increase the processing time, but I think  it is more correct to do that. (I thought of a way to do this setup in a more efficient way but this is left for a future fix.)

### DIFF
--- a/link-grammar/preparation.c
+++ b/link-grammar/preparation.c
@@ -50,7 +50,15 @@ set_connector_length_limits(Sentence sent, Parse_Options opts)
 	bool all_short = opts->all_short;
 	Connector_set * ucs = sent->dict->unlimited_connector_set;
 
-	if (len >= sent->length) return; /* No point to enforce short_length. */
+	if (0)
+	{
+		/* Not setting the length_limit saves observable time. However, if we
+		 * would like to set the ZZZ connector length_limit to 1 for all
+		 * sentences, we cannot do the following.
+		 * FIXME(?): Use a flag that the sentence contains an empty word. */
+		if (len >= sent->length) return; /* No point to enforce short_length. */
+	}
+
 	if (len > UNLIMITED_LEN) len = UNLIMITED_LEN;
 
 	for (i=0; i<sent->length; i++)
--- a/link-grammar/preparation.c
+++ b/link-grammar/preparation.c
@@ -30,8 +30,12 @@ set_connector_list_length_limit(Connector *c,
 {
 	for (; c!=NULL; c=c->next)
 	{
-		if (all_short ||
-		    (conset != NULL && !match_in_connector_set(conset, c)))
+		if (0 == strncmp(c->string, "ZZZ", 3))
+		{
+			c->length_limit = 1;
+		}
+		else if (all_short ||
+		         (conset != NULL && !match_in_connector_set(conset, c)))
 		{
 			c->length_limit = short_len;
 		}
--- a/link-grammar/prune.c
+++ b/link-grammar/prune.c
@@ -1141,7 +1141,7 @@ right_connector_list_update(prune_context *pc, Sentence sent, Connector *c,
 
 	/* n is now the leftmost word we need to check */
 	foundmatch = false;
-	for (; n < ub ; n++)
+	for (; n <= ub ; n++)
 	{
 		pc->power_cost++;
 		if (left_table_search(pc, n, c, shallow, word_c))


link-grammar/prune.c
```diff
@@ -1141,7 +1141,7 @@ right_connector_list_update(prune_context *pc, Sentence sent, Connector *c,
 
 	/* n is now the leftmost word we need to check */
 	foundmatch = false;
-	for (; n < ub ; n++)
+	for (; n <= ub ; n++)
```
linas: thanks good catch

link-grammar/preparation.c
```diff
@@ -30,8 +30,12 @@ set_connector_list_length_limit(Connector *c,
 {
 	for (; c!=NULL; c=c->next)
 	{
-		if (all_short ||
-		    (conset != NULL && !match_in_connector_set(conset, c)))
+		if (0 == strncmp(c->string, "ZZZ", 3))
+		{
+			c->length_limit = 1;
+		}
```
linas: according to my measurements; doing this actually gives a run that is a few percent slower than not doing this.  I just pushed a change which replaces it by string_set_compare, and only then is it a few percent faster... (for the In vivo sentence)

linas:
I don't understand " took about 50% less CPU in this fixed version." -- for me there is no change at all after pulling this request -- in fact, it was a few percent slower. replacing by the string-set compare made it a few percent faster.

ampli:
The comparison I did is to the version before implementing the **length_limit** check  shortcut in prune.c.
Please check that.

linas:
You mean this one:  6c052904e08ffcf08be226f88a49a9c6189256d6  -- yes, for me it is more than twice as fast; that single commit makes it 2.5x faster.

linas:
russian corpus is unaffected by any of these changes, and maybe even a few percent slower. Also, this change makes no difference:
```
--- a/link-grammar/preparation.c
+++ b/link-grammar/preparation.c
@@ -32,6 +32,10 @@ set_connector_list_length_limit(Connector *c,
 {
        for (; c!=NULL; c=c->next)
        {
+               if (0 == strncmp ("LL", c->string, 2))
+               {
+                       c->length_limit = 1;
+               }
                if (string_set_cmp (ZZZ, c->string))
                {
                        c->length_limit = 1;
```

which makes me think that we could/should eliminate the ZZZ check on the grounds that it makes almost no perceptible differrence; the few percent is hard to measure; my timing data is noisy.

ampli:
It is slightly slower due to the removal of the following optimization in order to allow the ZZZ length_limit=1 setup:
`if (len >= sent->length) return; /* No point to enforce short_length. */`

It has the benefit of reducing duplicate parses on Russian sentences with null-count>1, due to bogus ZZZ distant linkages that bypass the null-linked words.

One of my checks regarding the pruning algo is whether length_limit==1 can serve as a hint to improve the efficiency of the pruning.

ampli:
> my timing data is noisy.

I mainly use for timing Python's **timeit** (via a script).
With 7 sets of best-of 7 (total 49 runs) the benchmarks are much less noisy.
I also use things like 50 sets of best-of 3 (or even more total runs) in over-night benchmarks.
The "**best-of**" runs reduce by much the caching variation artifacts.

To farther reduce the noise, I run the benchmarks of old/new in parallel (I have 6 CPU machine, each heaving 2*thread), because in serial runs I found variations that depend on the CPU temperature.

Recently I added a flag to the benchmark script (an additional script - I have scripts for everything I do more than several times...) to lock the CPU  and run in realtime, to further reduce variations (one of them could be enough). I still don't have much experience with it. (Such a thing is a rather bad idea on a 2-CPU machine, but on 6-CPU it is not noticed.)

linas:
OK, I think I now understand: `length_limit==1` for ZZZ can never improve pruning. Here's why. the pruning checks *always* start at length=1, and look in the has tables for a matching connector. This check *always* succeeds, the connector is always found, and so length=2 is never checked; doesn't have to be, so changing the length limit does nothing.

Almost the same thing happens almost all the time for the russiaan LL links; although if a stem and a suffix can never link, then a matching LL wil not be found at any length, so limiting the search to length=1 should make a difference, when there are bad stem-suffix combinations possible.

linas:
In fact, there's a speedup: if the connector is ZZZ, then don't bother looking in has table; we know it will succeed always.

ampli:
For some reason there is this speedup: Less invocations of right_table_search()/left_table_search(), as indicated by the reduced **power_cost**.

linas:
Huh, that's strange.  If my theory was correct, about always finding ZZZ in the hash table at distance 1, then power_cost should be the same. But its not.  So I don't understand that.

linas:
strange. For the second K in D, first ZZZ, it searches way to the left to fnid a connector. Is there an off by one error??? (so that it doesn't see the very first one?)

ampli:
This is indeed strange. Maybe a few printf's there can clarify it.

linas:
word_c and w are always identical.

ampli:
Indeed when c->length==1 then always word_c==w.
I don't know if this is fine or not...

linas:
you can read the source, the routines are always called with exactly the same value.  Its probably a left-over from fat links, back when it was possible for the two values to be different.

linas:
never mind. It seems simply that word 43 just has 18K disjuncts on it !  I still don't understand why my theory isn't correct.

PR #254, opened 2015-12-15
Fast matcher: Fix length-limit check of the match-list connectors

ampli:
A match-list contains only connectors with the same UC prefix.  However,
some connectors may have the same UC prefix but a different
length_limit, e.g. X and Xc which have unlimited length, and Xd which
have short_length.

If lc or rc would be an unlimited-length connector (e.g. X)
and the match-list would contain matching connectors with different
length-limit, e.g. Xc and Xd, the whole match-list could be missed
if the first connector is a short_length one. (If lc or rc would be
a short_length connector then we don't care.)

Such a thing has not been observed by a debug print. In addition,
absolutely no linkage in all the languages got changed due to this fix
(in English the fixes.batch has been verified too).

However, the code seems wrong in principle and in order to be on the safe
side it is fixed here. (This is a very subtle issue.)

EDIT: If connectors with the same UC prefix could have more than 2 different values of length_limit, then the remaining global length_limit check would need to move to the mach-lists traversing code.

### DIFF
--- a/link-grammar/fast-match.c
+++ b/link-grammar/fast-match.c
@@ -534,20 +534,12 @@ form_match_list(fast_matcher_t *ctxt, int w,
 	if ((lc != NULL) && ((w - lw) <= lc->length_limit))
 	{
 		mxp = get_match_table_entry(ctxt->l_table_size[w], ctxt->l_table[w], lc, -1);
-		if ((NULL != mxp) && (NULL != *mxp) &&
-		    ((w - lw) <= (*mxp)->d->left->length_limit))
-		{
-			ml = *mxp;
-		}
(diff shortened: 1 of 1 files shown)


linas:
4.0.dict:
```
UNLIMITED-CONNECTORS:
      S+ & O+ & CO+ & C+ & Xc+ & MV+ & CC+ & TH+ & W+
      & RW+ & Xp+ & Xx+ & CP+ & SFsx+ & WV+ & CV+;
```

So SFsx is unlimited, but SFi is limited.   A more common case is Xd an Xc what are limited, but Xx which is not.  Xx example:  "this is a very very very very very very long sentence; this is shorter, with a comma."   has Xd to comma  maybe you can use this to test...

Issue #255, opened 2015-12-18
compile error on macosx el_capitan


nijikon:
```
Making install in sat-solver
  CXX      sat-encoder.lo
  CXX      util.lo
  CXX      variables.lo
  CXX      fast-sprintf.lo
  CXX      word-tag.lo
sat-encoder.cpp:1320:38: error: comparison between pointer and integer ('int' and 'const char *')
      display_linkage_disconnected = test_enabled("linkage-disconnected");
                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../link-grammar/error.h:57:9: note: expanded from macro 'test_enabled'
        (('\0' != test) ? feature_enabled(test, feature) : NULL)
          ~~~~ ^  ~~~~
1 error generated.
make[2]: *** [sat-encoder.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [install-recursive] Error 1
make: *** [install-recursive] Error 1
```

```
==> Configuration
HOMEBREW_VERSION: 0.9.5
ORIGIN: git@github.com:polishgeeks/homebrew.git
HEAD: 828c23965741a943cfee0f9759f6fe797766072b
Last commit: 3 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: quad-core 64-bit haswell
OS X: 10.11.2-x86_64
Xcode: 7.0
CLT: 7.2.0.0.1.1447826929
Clang: 7.0 build 700
X11: 2.7.7 => /opt/X11
System Ruby: 2.0.0-p645
Perl: /usr/bin/perl
Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /Users/n/.rvm/rubies/ruby-2.2.3/bin/ruby
Java: 1.8.0_45
==> ENV
HOMEBREW_CC: clang
HOMEBREW_CXX: clang++
MAKEFLAGS: -j4
CMAKE_PREFIX_PATH: /usr/local
CMAKE_INCLUDE_PATH: /usr/include/libxml2:/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers
[...]

PR #257, opened 2015-12-18
Disable Perl bindings by default

ampli:
It includes string literal append without a space, which causes and
error like that when compiling as c11:

/System/Library/Perl/5.18/darwin-thread-multi-2level/CORE/pad.h:323:17:
error: invalid suffix on literal; C++11 requires a space between literal and identifier [-Wreserved-user-defined-literal]
              "Pad 0x%"UVxf"[0x%"UVxf"] set_cur    depth=%d\n", \

The solution is to compile Perl SWIG stuff with c03.

(Fixes issue #255, message started by "Making install in perl")

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -219,6 +219,13 @@ AC_ARG_ENABLE( python-bindings,
   [enable_python_bindings=no]
 )
 
+# Perl now disabled by default
+AC_ARG_ENABLE( perl-bindings,
+  [  --enable-perl-bindings  create perl bindings to the link-grammar library],
+  [],
+  [enable_perl_bindings=no]
+)
+
 # ====================================================================
 # SAT solver now enabled by default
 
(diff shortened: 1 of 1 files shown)


ampli:
For some reason this file has not been updated on this system to C11.

ampli:
This fix allows the library to compile, but is not good.
I will overwrite it by another one.

linas:
I'll wait. If you want me to merge this anyway, let me know.

ampli:
I was not successful  in changing the Perl SWIG compilation to use -std=x++03, as CXXFLAGS got appended too, with its -std=x++11, and I didn't find how to overwrite that (changing it temporary had no effect from for some reason).

So I have just implemented the disabling by default as discussed in issue #255.

ampli:
This pull request is ready for merging.

linas:
thnaks

PR #261, opened 2015-12-19
Allow to increase the maximum number of displayed linkages - for development

ampli:
This is something I need...
It is an extension of the current `-test=auto-next-linkage` that I use in order to compare detailed batch runs of development library versions.

### DIFF
--- a/link-parser/link-parser.c
+++ b/link-parser/link-parser.c
@@ -339,6 +339,26 @@ static void print_parse_statistics(Sentence sent, Parse_Options opts)
 	}
 }
 
+/**
+ * Check for the auto-next-linkage test request (for LG code development).
+ * It is given using the special command: test=auto-next-linkage[:display_max]
+ * when :display_max is an optional indication of the maximum number of
+ * linkages to auto-display (the default is DISPLAY_MAX).
+ * For example, to issue up to 20000 linkages for each batch sentence,
+ * the following can be used:
+ * link-parser -limit=30000 -test=auto-next-linkage:20000 < file.batch
+ */
+static int auto_next_linkage_test(const char *test)
+{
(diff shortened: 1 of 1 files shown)


link-parser/link-parser.c
```diff
@@ -339,6 +339,26 @@ static void print_parse_statistics(Sentence sent, Parse_Options opts)
...
+ * It is given using the special command: test=auto-next-linkage[:display_max]
+ * when :display_max is an optional indication of the maximum number of
+ * linkages to auto-display (the default is DISPLAY_MAX).
+ * For example, to issue up to 20000 linkages for each batch sentence,
+ * the following can be used:
+ * link-parser -limit=30000 -test=auto-next-linkage:20000 < file.batch
+ */
+static int auto_next_linkage_test(const char *test)
+{
+	char auto_next_linkage_str[] = ",auto-next-linkage";
```
linas: should be `const char`

ampli: Why is that better?
Doing it as an array saves a pointer indirection (which is of course not important at all in this case).

linas: I just meant `const char auto_next_linkage_str[] =` makes the string into a compile-time const, and the compiler can do more optimization. e.g. skip creating the pointer and just pull straight out of a `.ro.data` secion.

ampli: I thought that on the highest optimization level, good compilers are doing it anyway, disregarding any const declaration that one can put in any possible place (i.e., they know from the code that it is const, and adding it doesn't add info for them). I didn't actually check this assumption.

linas: maybe you are right, and here it doesn't make a much of a difference.  I just have a habit: use const as often as you can.  So that just jumped out at me.

Issue #263, opened 2015-12-21
Order of linkages with the same cost + an array instead of list in form_match_list()


ampli:
I made the following change in form_match_list():
Instead of using a linked list to return the matching disjuncts, I used an array.
This has less overhead due to:
1. It is faster just to fill-in array elements.
2. Memory allocation:
- Can just malloc() all the array elements at once, instead of an malloc() per list element.
- No need to free() until finishing handling the sentence.

However, in case of returning an empty list it has (very) slight more overhead because it then returns one element with a NULL disjunct instead of just NULL (can be changed with a few more code lines, but the possible gain is very small).

Speed-up (**200** runs as 40 sets of best-of-5):
ru/corpus-fixes.batch: -0.71% __[*]__
en/corpus-basic.batch: -0.43%  __[*]__
en/corpus-fixes.batch: 5.2%
__[**]__ In vivo studies ...: 2.8%

__[*]__ May be consistent with ~0% due to the benchmark resolution.
__[**]__ The en/corpus-corpus-fix-long.batch with only this sentence.

However, because currently (i.e. before this change) the list is returned in reverse order, and in my change the array is in the match-list table order (because it is more efficient and elegant code this way), linkages of the same cost are returned in a different order.

For example, currently the first linkage of `This is a test" is:
```
	Linkage 1, cost vector = (UNUSED=0 DIS= 0.00 LEN=6)

    +----->WV----->+---Osm--+
    +---Wd---+-Ss*b+  +Ds**c+
    |        |     |  |     |
LEFT-WALL this.p is.v a  test.n 
```
However, after the said change, the 1st and 2nd linkages order is changed:

```
	Linkage 1, cost vector = (UNUSED=0 DIS= 0.00 LEN=6)

    +----->WV----->+---Ost--+
    +---Wd---+-Ss*b+  +Ds**c+
    |        |     |  |     |
LEFT-WALL this.p is.v a  test.n 
```

Supposing this change (of list to array) is fine by you (we can discuss the apparent speed reduction for short sentences - a benchmark of 1000 runs may be needed for a finer-grain resolution and to find out if to add returning NULL when the array is empty, which happens in several tens percents of the cases), it means there is a need to change the test suite in 2 places (a hard coded result and parses-en.txt).

Such a problem can happen due to other reasons:
1. Making pruning more aggressive.
2. Further efficiency changes in the match-lists, that may change their order.
[...]

PR #264, opened 2015-12-22
Fast matcher: Return the match-list as an array instead of a linked list

ampli:
This is the change I referred to in #263.
Two things to add:
**1.** I repeated the benchmarks several times, this time with RT priority (running current and new branches at once, as usual) in order to reduce fluctuations:
en/corpus-fixes.batch: ~>5%
en/corpus-basic.batch: ~<1%
Rest: ~0%
**2.** For now I took the short-range path of fixing tests.py to reflect the changed order of equal-cost linkages.

### DIFF
--- a/bindings/python-examples/parses-en.txt
+++ b/bindings/python-examples/parses-en.txt
@@ -4,7 +4,7 @@
 
 Ithis is a test
 O
-O    +----->WV----->+---Osm--+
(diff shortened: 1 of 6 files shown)


linas:
I glanced at the code, but did not carefully review it. It looked OK.  Thanks!

PR #265, opened 2015-12-23
Fast matcher etc.: "Cosmetic" changes

ampli:
1. In the previous change I added match-list push/get functions, but forgot to change the a description and use the "get" function in two places in which the match-list array was directly accessed.
2. Also included here old-typo fixes which I found while working on these files.

### DIFF
--- a/link-grammar/count.c
+++ b/link-grammar/count.c
@@ -290,7 +290,7 @@ static Count_bin do_count(fast_matcher_t *mchxt,
 		int id = get_match_list_element(mchxt, mlb) ?
 		            get_match_list_element(mchxt, mlb)->match_id : 0;
 #endif
-		for (; mchxt->match_list[mle] != NULL; mle++)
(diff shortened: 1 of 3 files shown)


linas:
code did seem to work without these changes!?

linas:
p.s. plan to publish 5.3.3 today or tomorrow let me know if I should wait.

ampli:
> code did seem to work without these changes!?

They are all cosmetic...

> p.s. plan to publish 5.3.3 today or tomorrow let me know if I should wait.

No need to wait.

I'm working now on an initial and ad-hoc multi-suffix split function, but it will take several days more at least, and meanwhile I postpone all the other LG work until I finish it. (I say  "initial and ad-hoc" because I have a wider plan for general tokenization, which I started to implement but it will take some more (unknown) time.)

linas:
OK 5.3.3 is out. I'm not announcing on mailing list though.

PR #266, opened 2015-12-25
Tokenization efficiency fixes

ampli:
**1.** The tokenization code tries to further suffix-split stems and affixes. It always fails to do that, as
stems include a stem mark and affixes include an affix mark, and the dicts are not supporting splitting them further. Not only its slows down the program, but it also clutters the debug output.
This fix prevents further tokenization handling of such tokens - 4% saving on the ru batch.

**2.** In the same occasion I observed that there is a redundant buffer copy when the morphemes are not marked (mostly happens on English). The saving is relatively small and has not been benchmarked.

**3.** strncpy() has been replaced by memcpy(). This can be done in other places in which strncpy() is currently used, and there is no NUL character in the copied string.

### DIFF
--- a/link-grammar/tokenize.c
+++ b/link-grammar/tokenize.c
@@ -465,18 +465,20 @@ Gword *issue_word_alternative(Sentence sent, Gword *unsplit_word,
 			token_ord++;
 			switch (at)
 			{
-				size_t sz;
-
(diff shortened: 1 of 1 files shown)


linas:
thanks!

PR #267, opened 2016-01-02
Add support for static linking on windows

JasonPaulos:
I recently tweaked some of the link-grammar headers in order to compile as static library for windows, and I thought others who want to do the same might benefit from the changes I made. I added the explanation below for how to build a static library to the README:

By default, the library is configured to create a DLL. If you want to instead build a static library, the macro LINK_GRAMMAR_STATIC must be defined before the inclusion of any header files for both the compiling of the link-grammar library and for the application that uses it. Other compiler settings will also have to be changed to create a static library of course.

### DIFF
--- a/README
+++ b/README
@@ -243,6 +243,13 @@ BUILDING on Windows
    See also:
    http://ftp.gnome.org/pub/gnome/binaries/win32/dependencies/regex.README
 
+   By default, the library is configured to create a DLL. If you want
+   to instead build a static library, the macro LINK_GRAMMAR_STATIC must
+   be defined before the inclusion of any header files for both the compiling
+   of the link-grammar library and for the application that uses it. Other
+   compiler settings will also have to be changed to create a static library
+   of course.
+
    The different build methods below are NOT regularly tested, and
    some link-grammar versions may have build issues.  If you are an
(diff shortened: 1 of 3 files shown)


linas:
Thanks Jason.   Why would you ever want to build it static, though? You won't get any performance boost from doing this... ?

JasonPaulos:
There are a few advantages of linking statically, such as not having to include the Visual C++ runtime DLL with your application. As far as I know, performance is about the same, so it usually comes down to preference whether one chooses to link statically or dynamically. Moreover, I'm working on some link-grammar bindings for node.js, and since node is built statically I had to also build link-grammar statically.

linas:
! There are node.js bindings for a very old/obsolete version of link-grammar here  https://github.com/dijs/link-grammar . I'm rather frustrated and irked that the fellow there hasn't bothered to update those bindings, esp since that seems to come up in the #1 spot in google search.  If there was some way you could "fix" that e.g with a pull request, that would be great.

Alternately, I can include node.js bindings in the default link-grammar dist. That is, in a subdirectory here:  https://github.com/opencog/link-grammar/tree/master/bindings Naively, I would think that the easiest way to get started would be to write a wrapper around the swig FFI.

PR #269, opened 2016-01-03
Fix contraction regognition

ampli:


link-grammar/utilities.h
```diff
@@ -201,6 +201,10 @@ int strncasecmp(const char *s1, const char *s2, size_t n);
 #define MAX(X,Y)  ( ((X) > (Y)) ? (X) : (Y))
 #endif
 
+#if !defined(ARRAY_SIZE)
+#define ARRAY_SIZE(a) (sizeof(a) / sizeof((a)[0]))
+#endif
+
 /* Optimizations etc. that only gcc understands */
```
linas: fyi

http://stackoverflow.com/questions/8018843/macro-definition-array-size
http://zubplot.blogspot.com/2015/01/gcc-is-wonderful-better-arraysize-macro.html

ampli: Opened issue #275.

PR #272, opened 2016-01-07
tokenization_done(): Only dict words are done

ampli:
Also remove breaking the loop on MT_INFRASTRUCTURE, as it seems it is
not needed any more (detailed results of all languages are the same).

--

Note: This is one in the series of pull requests I back-port from the multi-affix branch (which still needs more polish and tests, including completing a demo dictionary).

I need the rest of already-sent pull requests to be applied, because I fear that I may start to generate conflicting ones...

### DIFF
--- a/link-grammar/tokenize.c
+++ b/link-grammar/tokenize.c
@@ -817,6 +817,29 @@ static Gword *wordgraph_getqueue_word(Sentence sent)
 	return w;
 }
 
+/**
+ * Prevent a further tokenization of all the subwords in the given alternative.
+ * To be used if the alternative represents a final tokenization.
+ */
+static void tokenization_done(Dictionary dict, Gword *altp)
+{
+
+	Gword *alternative_id = altp->alternative_id;
+
+	for (; altp->alternative_id == alternative_id; altp = altp->next[0])
+	{
+		if (NULL == altp) break; /* just in case this is a dummy word */
+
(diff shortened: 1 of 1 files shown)


linas:
again, sorry for delay.   One way to avoid conflicting pulls is to merge all these into your own temp-master, and keep developing from the temp-master.  As I merge the old pulls, and as you submit new pulls, the diffs in the new pulls will "resolve themselves", as long as you keep merging from master. 

The graph ends up looking messy, and the intermediate changes can look confusing, but the end result is still a clean set of diffs.

ampli:
A problem I encountered with that technique is when I would like to continue sending pull requests while the previous ones has not been handled yet.
In that case a new pull requests includes all the previous commits (that are already included in the said previous pull requests that I have sent and have not been handled). This is maybe not harmful but it adds confusion. I don't know if there is a way to avoid that.

linas:
yes, it does do that (and does add to the confusion.) But after the older pulls get merged into master, if you then merge master into the newer pull, the older changes become 'invisible'. 

Your pick. Sometimes I can respond quickly, sometimes not. I suppose I should just give you permission to do tit yourself...

ampli:
It is better if you do it, supposing you review the changes.
I addition, sometimes I need your guiding regarding the direction of the changes, and
sometimes I also add notes that need your attention.

PR #274, opened 2016-01-07
Fix test suite according to latest dict changes

ampli:
As changed in commit 662da7d0df6b1220e8afea1c9df4c11f79fc13ae.


linas:
thanks!

Issue #276, opened 2016-01-08
add_empty_word() and multi-suffix


ampli:
add_empty_word() currently doesn't add EMPTY_CONNECTOR+ (ZZZ+) to suffixes.
This causes a problem when there are alternatives with different number of suffixes, as an EMPTY-WORD cannot then be attached to the last suffix. This effectively discards all the alternatives which need an EMPTY-WORD (because they are shorter than the longest alternative).

Depending on the split method of separate_word(), this may happen with multi-prefix too (and with multi-stems in multi-stem languages). My initial muiti-suffix code used incremental splits, and thus had this problem too (incremental splits is maybe more "elegant" but it has a problem with unknown words and I didn't find a non-cumbersome solution for that so I abandoned this path of code).

However, there is no problem with multi-prefix with my current (yet unpublished) code for multi-affixes, because it splits the whole word at once (and it doesn't handle multi-stems anyway). The problem with it is only, as said above, with multi-suffix.

Possible solutions:
**1.** Add EMPTY_CONNECTOR+ to suffixes too.
If may cause a slight performance degradation to Russian, but this is said to be cured in the proposed new expression-pruning code that considers length-limit. Alternatively, the EMPTY_CONNECTOR+ addition to suffixes can be implemented only if SUFMAX>1 (in my multi-affix split code, SUFMAX is the maximum number of suffixes to split off a word).
**2.** Maybe better: Add EMPTY_CONNECTOR+ only to words that absolutely need it (since their next word is EMPTY-WORD).
In [determine_word_expressions()](https://github.com/opencog/link-grammar/blob/0ff04c28049afd5c952ee8dfb389a4c16e889ae2/link-grammar/tokenize.c#L2634), call [build_word_expressions()](https://github.com/opencog/link-grammar/blob/0ff04c28049afd5c952ee8dfb389a4c16e889ae2/link-grammar/build-disjuncts.c#L446) with an additional argument `bool add_empty_connecrtor` that will tell whether to invoke add_empty_word() (which will then always add EMPTY_CONNECTOR+).

ampli:
It seems option **2** cannot be implemented this way, because when `build_word_expressions()` is invoked it is still not known whether the next word is an EMPTY-WORD.
[...]

PR #278, opened 2016-01-11
add_empty_word(): Invoke only when needed

ampli:
This is a step toward a multi-affix split.
See issue #276.
This change adds an empty-word connector on suffixes too, but only when needed.
It also significantly increases the speed of English processing, especially for long sentences.

The benchmark showed a very slight efficiency decrease for Russian, but a detailed inspection ensured that the new code never adds  the ZZZ+ connector to more words than before, and actually (on average) adds it to less than 50% of the words comparing to what the current code does. Also, the new `is_exp_like_empty_word()` doesn't seem to be a problem since replacing it by a simple variable comparison doesn't improve the speed my a noticeable amount.

In the same occasion, I cleaned up (from the files changed in these commits) unneeded include files that have been accumulated over time.

### DIFF
--- a/link-grammar/dict-file/read-dict.c
+++ b/link-grammar/dict-file/read-dict.c
@@ -1021,6 +1021,21 @@ void add_empty_word(Dictionary const dict, X_node *x)
 	/* Replace plain-word-exp by {ZZZ+} & (plain-word-exp) in each X_node.  */
 	for(; NULL != x; x = x->next)
 	{
+		/* Ignore stems for now, decreases a little the overhead for
+       * stem-suffix languages.
+       * This line should be removed if these 2 conditions happen together:
+       * 1. Multi-stem splits are to be supported.
+       * 2. Affix splits are done by wordgraph splits.
+       *
+		 * FIXME: A more general solution instead of this line is to add
+		 * empty-word connectors only to the x-nodes that need them, instead
+		 * of adding them, like now, to all the x-nodes of the word that come
+		 * before the empty-word. This will support wordgraph affix splits (if
+		 * will ever be done) and will slightly increase the efficiency of
+		 * handling sentences with multi-suffix splits (less empty-word
+		 * connectors in the sentence). */
+		if (is_stem(x->string)) continue; /* Avoid an unneeded overhead. */
+
 		/* zn points at {ZZZ+} */
 		zn = Exp_create(&eli);
 		zn->dir = '+';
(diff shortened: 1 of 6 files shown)


link-grammar/dict-file/read-dict.c
```diff
@@ -994,70 +989,75 @@ static Exp * make_connector(Dictionary dict)
...
+
+	/* Replace plain-word-exp by {ZZZ+} & (plain-word-exp) in each X_node.  */
+	for(; NULL != x; x = x->next)
+	{
+		/* Ignore stems for now, decreases a little the overhead for
+       * stem-suffix languages.
+       * This line should be removed if these 2 conditions happen together:
+       * 1. Multi-stem splits are to be supported.
+       * 2. Affix splits are done by wordgraph splits.
+       *
```
linas: indentation

linas:
What happens if a word can also be broken into 5 parts -- does that mean that we have to add the word, and four ZZZ's? Or its a multi-connector? Or maybe that is what "depth" is, in your new smart match?   I did not really take the time to understand the new smart match.

ampli:
Say we have a word:
`ABCDE`
that can be broken to:
`A B C D E`

Then we end with this in the final 2D array (ø represents the empty word): 
```
ABCDE ø ø ø ø 
A     B C D E
```

Using one empty word with a multi-connector is an interesting idea, but it will add code and I'm not sure it will increase the parse speed.

The added smart_match flag in exp_has_connector is not related.
I used this function to identify the empty-word and empty-word-like words (currently only doublequote and the experimental CAP tokens) by searching ZZZ- in words. To improve the efficiency of this search, I used smart_match=false so a simple connector comparison will be used.
"depth" was added in order to make this search more efficient, by making the search "shallow"
supposing ZZZ- is in the upper level of such word expressions.

BTW, the empty word device can be eliminated totally by marking optional words (B C D E in the above example are optional). Then the parse algorithm can be changed to allow such optional words to by "null words" without increasing the null count.

linas:
Oh, I like the "optional word" idea.  Will it work?  I'm hoping that it could be simpler than the null-word stuff...

ampli:
I forgot to add (regardin the example above):
The old code already adds the empty words as needed.
The new code adds ZZZ+ to ABCDE and to the first 3 empty words.
Currently, it also adds ZZZ+ to B C and D, a thin that is not needed __[*]__.
However, it doesn't add ZZZ+ to any other word in the sentence (supposing no more empty words in the sentence). As opposed to that, the old code added ZZZ+ to every word in the sentence.

__[*]__ This can be eliminated if desired.

ampli:
> Oh, I like the "optional word" idea. Will it work? I'm hoping that it could be simpler than the null-word stuff...

It seems to me it must work in principle, and also the SAT solver can be changed accoringly.
I think the needed changes are very small, but they may not be trivial to implement.

linas:
On Mon, Jan 11, 2016 at 2:15 PM, Amir Plivatsky <notifications@github.com>
wrote:
>
> I think the needed changes are very small, but they may not be trivial to
> implement.
>

:-)

PR #281, opened 2016-01-13
en/4.0.dict: Restore a missing apostrophe

ampli:
The problem was detected in this sentence from en/corpus-basic.batch:
The boys' bedrooms will be enlarged

This problem happened in commit 220b86f9510800d2d9cf2c6cf93692ec6ec6468c.

### DIFF
--- a/data/en/4.0.dict
+++ b/data/en/4.0.dict
@@ -11341,7 +11341,7 @@ $ USD.c US$.c C$.c AUD.c AUD$.c HK.c HK$.c
 
 "&": G- & {Xd- & G-} & G+;
 
-"’" "": YP- & (({AL-} & {@L+} & (D+ or DD+)) or [[<noun-main-x>]] or DP+);
+"’" "'": YP- & (({AL-} & {@L+} & (D+ or DD+)) or [[<noun-main-x>]] or DP+);
 
 % Possessives
 "'s.p" "’s.p":


linas:
Hmm that's bad, that file is generated from 4.0.dict.m4

ampli:
How did this suddenly happen?
4.0.dict.m4 still contains `"'"` as always.
Using m4 on it now strips it, unless I double it (but it was never doubled befoere).

ampli:
I now see that you also solved it by doubling.
But I still wonder how it worked before.

linas:
did it work before?  I did not look back to see last time it was correct.

linas:
It got broken in this commit:  
220b86f9510800d2d9cf2c6cf93692ec6ec6468c Dec 7, 2015

and was fine before that. This suggests that my m4 recently started mis-behaving.  That's bad. There were other commits just days/weeks earlier, and it was fine.  Does your m4 do this?  Just say `m4 4.0.dict.m4 > 4.0.dict`   Perhsaps some other change int the file is causing this weird behavior?

ampli:
> Perhsaps some other change int the file is causing this weird behavior?

On a the version before that commit, m4 works just fine, so it seems it is some change in the file. From a glance I could not identify the change, and there may be a need to selectively undo parts of it in order to find out the reason.

linas:
Removing the comment `% CP or [()]: "Let's go!" she said.` is what breaks it.

linas:
Removing the `We do need Wa for comma's` further on un-breaks it, but then breaks Possiesives:
```
 % Possessives
-"'s.p" "’s.p":
+"s.p" "’s.p":
```

linas:
Yuck. I pushed one more fix.  There is some fragility with tcs and backtics due to how m4 works.

ampli:
As you noted, the problem results due to the backticks in:
```
« 《 【 『 ` „:
  QUd-;
» 》 】 』 ` “:
  <post-quote>;
```
This can be fixed by using changequote (no real need here to redefine also ' , but two arguments are to be used):
```
changequote(\,/)dnl
% The line above redefines the m4 quotes to allow specifing the ` character.
% It is restored below by changequote without arguments.
« 《 【 『 ` „:
  QUd-;
» 》 】 』 ` “:
  <post-quote>;
changequote dnl
```
Then the apostrophe deletion commit can be reverted.
I can send a pull request if such a change is fine by you.

linas:
Yes, that seems like a better solution.  It might be needed in several spots, there seemed to be some oddball behavior I did not fully understand, as I played with the earlier fixes.

ampli:
Another solution uses a macro to generate a backtick:
``` 
define(`BACKTICK',`changequote(<,>)`dnl'
changequote`'')
```
And then:
```
« 《 【 『 BACKTICK() „:
  QUd-;
» 》 】 』 BACKTICK() “:
...
until 'til ’til ‘til BACKTICK()til til till.r:
```
Please tell me which solution you prefer and I will send a pull request.

linas:
Hmm. Not sure I care any more. What's there works now ... neither of the above changequote tricks will prevent any future mistakes, so it seems OK to leave as-is.

PR #284, opened 2016-02-27
KazakhDict

TatianaBatura:
Prototype Kazakh dictionary and some changes to Turkish dictionary.


linas:
Thank you Tatiana!

linas:
By the way, you do NOT have to remove examples and README's just to make it look like the other languages.  You can have as many examples, documents and README in here as you want -- any notes, anything to make things easier for yourself or for others.

linas:
Also, the examples, documents and README do NOT have to be in English -- it's OK to have them in Russian or any other language.

linas:
Oh, and finally -- if you have web pages that you would like to publish, that provide additional documentation and so on, I can provide those too -- in this general area:  http://www.abisource.com/projects/link-grammar/dict/index.html  -- see the bottom of the page.  If I ever get enough web pages on other languages, I will re-organize and split this all up.

Of course, you are welcome to publish web pages anywhere else, as well.  Its often easier to just have everything in one place.

data/kz/4.0.dict
```diff
@@ -0,0 +1,135 @@
...
+%
+% Dictionary version number is 5.3.0 (formatted as V5v3v0+)
+<dictionary-version-number>: V5v3v0+;
+% The empty word.
+EMPTY-WORD.zzz: ZZZ-;
+
+мен.pron: {S1s+};
+сен.pron: {S2s+};
+сіз.pron: {Sr2s+};
+ол.pron: {S3s+};
```
linas: I notice that you are using curly braces {} everywhere.  You realize that this means that the connector is optional?  That is,  the connector does not have to be used if its not there?    I believe that in most cases, you should not be using the curly braces, since you DO want to use these connectors.  

Thus, for the above, I believe that this would be more correct:
```
мен.pron: S1s+;
сен.pron: S2s+;
```
and so on ...

If a word cannot be connected, then the parser has a built-in mechanism for ignoring such words;  you do not need to make them optional to allow them to be ignored.

TatianaBatura: Yes, you're right. I'll correct that.

TatianaBatura:
Thanks, I'll take that into account.
BTW, we figured out why non-Latin output get truncated on Windows. I wrote my comment in #285

linas:
Its better if you add new comments, than editing the old ones .. new comments show up as email, so I see them. Edits are silent, and won't be noticed, unless you point them out.

PR #291, opened 2016-04-01
Further fix the long-irritating backtic bug

ampli:
Just now an apostrophe is not a known word in the English dictionary.
This pull-request continues the fix introduced in f79d2492b0ef5c08cb08d26956713fc176a4bf57.
The commit message explains the reason of the problem and how to avoid such problems in the future.
(Issue #281.)

### DIFF
--- a/data/en/4.0.dict
+++ b/data/en/4.0.dict
@@ -9329,7 +9329,7 @@ as_soon_as:
 % J+ & CO+: "Until yesterday, ..."
 % The double-back-tick becomes a single backtick after m4
 until 'til ’til ‘til `til til till.r:
-  ((Mgp+ or J+ or JT+ or UN+)
+   ((Mgp+ or J+ or JT+ or UN+)
     & (({Xc+ & {Xd-}} & CO+) or ({Xd- & Xc+} & MVp-) or [Mp-]))
   or (<subcl-verb> & (({Xc+ & {Xd-}} & CO*s+) or ({Xd- & Xc+} & MVs-)));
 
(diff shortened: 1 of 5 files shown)


data/en/4.0.dict
```diff
@@ -11375,7 +11375,7 @@ $ USD.c US$.c C$.c AUD.c AUD$.c HK.c HK$.c
 "&": G- & {Xd- & G-} & G+;
 
 % The double-single-quote is converted to a single single quote by m4.
-"’" "''": YP- & (({AL-} & {@L+} & (D+ or DD+)) or [[<noun-main-x>]] or DP+);
+"’" "'": YP- & (({AL-} & {@L+} & (D+ or DD+)) or [[<noun-main-x>]] or DP+);
```
linas: the change above is immediately contradicted by the text in the line above it: viz: "% The double-single-quote is converted to a single single quote by m4."

data/en/4.0.dict.m4
```diff
@@ -9334,7 +9336,7 @@ $ USD.c US$.c C$.c AUD.c AUD$.c HK.c HK$.c
 "&": G- & {Xd- & G-} & G+;
 
 % The double-single-quote is converted to a single single quote by m4.
-"’" "''": YP- & (({AL-} & {@L+} & (D+ or DD+)) or [[<noun-main-x>]] or DP+);
+"’" "'": YP- & (({AL-} & {@L+} & (D+ or DD+)) or [[<noun-main-x>]] or DP+);
```
linas: Same here: the lien above explains why there are two tick marks.

ampli:
I removed the comments in pull request #295.

linas:
The file `parses-quotes-en.txt` appears to be missing from this pull req ... the python unit test fails with a file-not-found.

PR #295, opened 2016-04-26
Fix comment rot

ampli:
Remove comments about doubling m4 quotes, per issue #291.

The opening ` quote is now ignored using changequote.
The closing ' quote doesn't need any special handling since it doesn't have a special meaning for m4 unless it is in quoting mode.


linas:
OK, thanks. I'd observed earlier that the double-quote stuff was just behaving weird.

PR #296, opened 2016-04-28
Add a missing test-suite file

ampli:
Needed for the test added at dc4ec39e00aeb60f549588257b3e5db0f75f4e6a.


linas:
thanks!

It just missed version 5.3.5 by just a few minutes!  (which I just pushed for the macos bug)

PR #301, opened 2016-05-12
Fix some MSVC warnings

ampli:
The first commit fixes a real problem.
The second one just prevents cluttering the compilation output with 32 warnings.

PR #306, opened 2016-06-15
Fix ifdef/endif indentation + a long comment

ampli:
- Fix ifdef/endif indentation: see #301.
BTW, I have left the #pragma lines aligned with the C code, in a hope this is more readable.
- Also fold a long comment line.


linas:
thanks!

Issue #299, opened 2016-05-12
Windows build break in version 5.3.7/5.3.8


ampli:
Due to the inclusion of utilities.h in b7fe27b4b435370ed88bc487608fe5871e218ba5, the linker cannot find printf_compat, fprintf_compat and sprintf_compat.

If we still would like to continue to include it, then I cannot find a good solution for this problem.
Possible solutions:
1. Incorporate these functions inline in utilities.h.
2. Put the new Mac lg_* functions in a separate file and include it instead.
3. Protect the macro definitions of __*printf_compat__ with #ifndef LG_APPLICATION, and #define LG_APPLICATION before including **utilities.h** in link-parser.c.
4. Use explicit #undef's for __*printf_compat__ after the inclusion of **utilities.h** in link-parser.c..

I'm for implementing option (3), unless/until a better solution is found. If this is fine by you I will send a pull request.

(I have several more fixes, including the off-by-one bug, that all can be included in a new library version.)

ampli:
Maybe a better solution:
Rename the __*printf_compat__ functions to __lg_*printf__, and protect the corresponding #ifdef's with #ifdef WINDOWS_PRINTF_COMPAT. Define this macro in the MSVC project file of the library.

I suppose this idea is fine, so I will implement it.
I got this idea because I wanted to add other macros to support smooth MSVC compilation and didn't want to clutter utilities.h with such (nonsense) macros.

BTW, the sprintf compatibility function should be changed to snprintf, since sprintf is not used any more anywhere in the code. I will do this too.

linas:
Those changes seem reasonable.

I also liked option 2 -- just create a new file, parser-utilities.c/h and use that for anything that is a part of the parser.  It could have a copy(!) of the `lg_fgetc` code -- somehow, having a copy seems to be better than exposing utilities.h to the "outside world" -- because, in the end, I really do not want "outsiders" to use utilities.h .

linas:
Done in pull #312

PR #308, opened 2016-06-17
Parser utilities

ampli:
Create parser-utilities.h, per issue #299.
For now, I don't find a need to create parser-utilities.c.
Maybe more compatibility things can be moved to parser-utilities.h (and/or a future parser-utilities.c).

Building on Windows was not checked after this change.
A latter pull-request will address Windows.

### DIFF
--- a/link-parser/parser-utilities.h
+++ b/link-parser/parser-utilities.h
@@ -12,6 +12,10 @@
 #ifndef _PARSER_UTILITIES_
 #define _PARSER_UTILITIES_
 
+#if !defined(MIN)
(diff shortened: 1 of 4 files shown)


link-parser/parser-utilities.h
```diff
@@ -0,0 +1,57 @@
+/***************************************************************************/
+/* Copyright (c) 2016 Linas Vepstas                                        */
```
linas: Uh, Could be your name here, too. I see that its mostly a cut-n-paste but really you're doing the work.

PR #316, opened 2016-06-28
Debug enhancement: Allow restricting messages by file name(s) too

ampli:
Example of usage:
link-parser -verbosity=7 -debug=tokenize.c

The value of the debug variable can be a comma delimited list of function and file names.

In order to get a full benefit of this enhancement, there is a need to convert all the debug messages to use lgdebug() / debug_level().


ampli:
Fixed in PR #333.
This was a bad error. But somehow it seemed to work (but maybe not really fine - I didn't recheck the erroneous code...).
I also didn't get the warning. I use GCC 5.3.1.

linas:
thanks. I didn't want to dig around to figure it out.

PR #322, opened 2016-06-30
Remove obsolete and unused BINRELOC support

jbicha:
Wikipedia says that [Autopackage](https://en.wikipedia.org/wiki/Autopackage) was replaced in 2010 by [Listaller](https://en.wikipedia.org/wiki/Listaller) which was replaced in 2014  by Limba.

I don't think Limba has BINRELOC. Now Limba's been mostly [superseded](http://blog.tenstral.net/2016/06/a-few-words-about-the-future-of-the-limba-project.html) by Flatpak and Snap for Linux systems which don't use BINRELOC either.


linas:
Thanks!  I always wondered what binreloc was for ...

PR #324, opened 2016-07-03
Bug fix: Initialize the special global options on parse-options creation

ampli:
You can easily see this inconsistency now that the `Info: Dictionary found at ...` message is set to be printed in verbosity>=1, but is not actually printed by `link-parser` unless any option is set in the command line.

### DIFF
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -102,10 +102,16 @@ Parse_Options parse_options_create(void)
 	po = (Parse_Options) xalloc(sizeof(struct Parse_Options_s));
 
 	/* Here's where the values are initialized */
-	po->verbosity = 1;
-	po->debug = (char *)"";
-	po->test = (char *)"";
+
+	/* The parse_options_set_(verbosity|debug|test) functions set also the
+	 * corresponding global variables. So these globals are initialized
+	 * here too. */
+	verbosity = po->verbosity = 1;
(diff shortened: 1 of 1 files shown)


link-grammar/api.c
```diff
@@ -1403,10 +1409,6 @@ int sentence_parse(Sentence sent, Parse_Options opts)
 {
 	int rc;
 
-	verbosity = opts->verbosity;
-	debug = opts->debug;
-	test = opts->test;
```
linas: and how are the globals changed, if not here?

ampli: This was an old workaround for this initialization bug (that this pull request fixes).

The globals are actually changed in the corresponding parse_options_set*() functions:
```
void parse_options_set_verbosity(Parse_Options opts, int dummy)
{
	opts->verbosity = dummy;
	verbosity = opts->verbosity;                <--------------------- here
	/* this is one of the only global variables. */
}
```

As you can see, these assignments would be redundant unless the said initialization bug.
They were added only to catch the setup done in parse_options_create(). Afterwards, they always assign to the globals the same value they already have.

Currently (without the pull request fix), these globals don't get their correct values in the dictionary reading code, and also not in a sentence_split() before calling sentence_parse().

So the fix is needed.

linas: OK. Thanks, just checking, it wasn't obvious.

PR #327, opened 2016-07-09
Python3 binding adaptation.

eugeneai:
Hi!

The following problems are fixed:
1. In my Arch Linux, compiled .so-module _clinkgrammar could not be loaded. Changed slightly the way of its loading.
2. Examples did not work in Pythin3 due to inconsistent syntax of examples and binding module itself. Fixed.
3. Test cases also fixed to run with Python3.

Best regards,
Evgeny Cherkashin

### DIFF
--- a/bindings/python-examples/example.py
+++ b/bindings/python-examples/example.py
@@ -4,6 +4,7 @@
 # Link Grammar example usage
 #
 import locale
+from __future__ import print_function, division  # We require Python 2.6 or later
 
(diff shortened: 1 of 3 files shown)


linas:
Thanks Cherkashin,

I made some additional changes to make things more consistent with python2.

How did you compile things for python3?  Normally, one would have to add the `-py3` flag to `swig` (i.e. change the Makefile.am to say `$(SWIG) -python3` etc at line 35)  Did you do that?  Or did you let it default?

Issue #325, opened 2016-07-03
User verbosity levels


ampli:
I tried an idea to make using link-parser (or the lib) more user friendly in case of problems, by enabling the user seeing more info regarding library "hidden" operations.
Will it be fine to **reserve verbosity values 1-3** for displaying user info?
(Programmer debug info will appear only on verbosity>=4, skipping the user info display.)

For example (please ignore the inconsistency of the missing program name from the dict messages - this can be fixed later):
```
$ link-parser -verbosity=2
verbosity set to 2
Info: Current directory: /usr/local/src/link-grammar-devel/windows/data/en
Info: Last-resort data directory: NULL
Info: Opening file ./en/dict.db (Not found)
Info: Opening file ./data/en/dict.db (Not found)
Info: Opening file ../en/dict.db (Not found)
Info: Opening file ../data/en/dict.db (Not found)
Info: Opening file /usr/local/share/link-grammar/en/dict.db (Not found)
Info: Opening file en/dict.db (Not found)
Info: Opening file ./en/4.0.dict (Not found)
Info: Opening file ./data/en/4.0.dict (Not found)
Info: Opening file ../en/4.0.dict
Info: Opening file ../en/4.0.regex
Info: Opening file ../en/4.0.knowledge
Info: Opening file ../en/4.0.constituent-knowledge
link-grammar: Info: Dictionary found at ../en/4.0.dict
link-grammar: Info: Using locale LC_CTYPE=en_US.utf8 ...
link-grammar: Info: Dictionary version 5.3.7.
link-grammar: Info: Library version link-grammar-5.3.7. Enter "!help" for help.
>
```

linas:
Sounds fine to me. But ...

The stark reality is that there are very few users of LG, and I find it very unlikely that any of them will ever play with the debug flags or attempt to do any significant debugging at all, with them.  You would be the only user of this feature; I might try it on rare occasions, but I am used to just adding printf to the code I am interested in.

If you are casting about for something to do, I would still encourage you to delve deeper into the theory.  The most important thing lacking from LG today is an ability to have it learn new grammars for new languages. Probably little else matters.

ampli:
[...]

PR #329, opened 2016-07-10
Add extended user information verbosity level

ampli:
Summary:

Per issue #325, with these differences:
- The reserved levels for user info are 2-4.
- Level 0 and 1 are not affected by this change.
- Using verbosity levels>4 (intended for developers) doesn't (mostly) produce the user info of levels 2-4.
- The level to show file search and open is 3.
- Level 2 produces the various steps CPU time. It is currently shown on levels>4 too. This can be changed.

Level 4 is not in use yet. Developer debug levels start from 5.

So for checking dictionary open, the following can be used, when (optionally) LL is the language or the language directory:
  link-parser LL -v=3

When debugging a particular "subsystem" (say post-processing), the easiest way to see only the relevant message is something like:
  link-parser -v=9 -debug=post-process.c

NOTE: The first commit was requested by GIT even though nothing at all (as far as I could see) got changed in these files. I guess this is due to the new .gitattributes. Initially avoiding to commit these files caused me much troubles (git misbehaved).  For example, in a master repository - as fetched from GitHUB - they are marked as modified. I hope this commit fixes this strange thing.


linas:
The msvc14 changes are making my git repo go completely insane!

linas:
I think it has something to do with the .gitattributes file ...

linas:
Git was going insane, the only fix was to create a branch, delete `msvc14/.gitattributes` then `git commit -a` then go back to master, and git pull.  Otherwise, what would happen was that all of the files in `msvc14` were marked as modified, no matter what, and trying to use `git checkout` failed to mark them as unmodified.

linas:
... the problem can be reproduced by saying
```
rm -r msvc14
git checkout msvc14
git status
```
which shows that all the files in the msvc14 directory are modified, and there is no wy to get an unmodified version, except to do the following:
```
git checkout -b foobar
rm msvc14/.gitattributes
git commit -a
git checkout master
git branch -D foobar
```

which will briefly restore things to sanity.  

I am going to permanently delete .gitattributes until some future time ...

PR #330, opened 2016-07-10
SWIG now generates wrapper for Python 3.

eugeneai:
Hi, Linas!

I've checked the ability switching to -python3 (for Python 3 only). It worked, I mean, all the tests are ok, examples too.

Addition: I've compared SWIG output for python2 and pyhon3, they are the same.

Best regards,
Evgeny

### DIFF
--- a/bindings/python/Makefile.am
+++ b/bindings/python/Makefile.am
@@ -32,7 +32,7 @@ DISTCLEANFILES =                                   \
 
 if HAVE_SWIG
 $(BUILT_SOURCES) $(pkgpython_PYTHON): $(SWIG_SOURCES)
-	$(SWIG) -python -module clinkgrammar -I$(top_srcdir)/link-grammar -o $@ $<
+	$(SWIG) -python3 -module clinkgrammar -I$(top_srcdir)/link-grammar -o $@ $<
 endif
 
 # The la MUST have the same name as the pm,
--- a/bindings/python/Makefile.am
+++ b/bindings/python/Makefile.am
@@ -32,7 +32,7 @@ DISTCLEANFILES =                                   \
 
 if HAVE_SWIG
 $(BUILT_SOURCES) $(pkgpython_PYTHON): $(SWIG_SOURCES)
-	$(SWIG) -python3 -module clinkgrammar -I$(top_srcdir)/link-grammar -o $@ $<
+	$(SWIG) -python -py3 -module clinkgrammar -I$(top_srcdir)/link-grammar -o $@ $<
 endif
 
 # The la MUST have the same name as the pm,
--- a/bindings/python/Makefile.am
+++ b/bindings/python/Makefile.am
@@ -31,7 +31,8 @@ DISTCLEANFILES =                                   \
    $(top_builddir)/bindings/python/__init__.py
 
 if HAVE_SWIG
-$(BUILT_SOURCES) $(pkgpython_PYTHON): $(SWIG_SOURCES)
+#$(BUILT_SOURCES) $(pkgpython_PYTHON): $(SWIG_SOURCES)
+$(BUILT_SOURCES) : $(SWIG_SOURCES)
 	$(SWIG) -python -py3 -module clinkgrammar -I$(top_srcdir)/link-grammar -o $@ $<
 endif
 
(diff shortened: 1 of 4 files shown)


ampli:
Hi Evgeny,
Have you changed configure.ac to search for `AM_PATH_PYTHON(3.0, ..." instead of `AM_PATH_PYTHON(2.6,...`? If not, and the bindings still works for you in python3, then my only guess is that you don't have python2 installed on that system so configure find python3.

For me, the tests succeed in python3 also without adding -py3, and it seems the Python bindings just do not happen to use anything which is handled by this flag. See: swig/swig#695.

----
----


In principle, configure.ac can be changed to generate binding for both python2 and python3.
If this is desired, I can look at it.

BTW, the tests run 15% slower under python3...

eugeneai:
The problem I faced and could not overcome for two days is that when I feed parser with sentences (most of them should not have valid linkage), at some time moment the library degrades and cannot do the linkage at all. But when I feed parser with repeated three clauses everything is ok.... Might be a library interference. ;-)

linas:
I don't want to commit this pull request, for the reasons I am about to remark on, line by line.

bindings/python/Makefile.am
```diff
@@ -31,8 +31,9 @@ DISTCLEANFILES =                                   \
    $(top_builddir)/bindings/python/__init__.py
 
 if HAVE_SWIG
-$(BUILT_SOURCES) $(pkgpython_PYTHON): $(SWIG_SOURCES)
-	$(SWIG) -python -module clinkgrammar -I$(top_srcdir)/link-grammar -o $@ $<
+#$(BUILT_SOURCES) $(pkgpython_PYTHON): $(SWIG_SOURCES)
+$(BUILT_SOURCES) : $(SWIG_SOURCES)
+	$(SWIG) -python -py3 -module clinkgrammar -I$(top_srcdir)/link-grammar -o $@ $<
```
linas: This permanently changes the bindings to python3, from python2, and I would rather not make this be permanent.  Ideally, the user would have a choice...  or maybe there is a way to build both!?

bindings/python/linkgrammar.py
```diff
@@ -29,7 +29,13 @@ def __init__(self, verbosity=0,
                  spell_guess=False,
                  use_sat=False,
                  max_parse_time=-1,
-                 disjunct_cost=2.7):
+                 disjunct_cost=2.7,
+            #     debug="",
+            #     test="",
+                 use_viterbi=False,
+                 use_cluster_disjuncts=False,
```
linas: The viterbi code does not work, and is unlikely to, in the next few years, if ever.  Thus, it is not suitable for the API.

The cluster-disjuncts code works, but the database for is is 7 years out-of-date, and it is unlikely that it will ever be rebuilt.  Thus, this is not suitable for the API.

bindings/python/linkgrammar.py
```diff
@@ -404,19 +492,60 @@ def __init__(self, sent):
...
-            if self.num == clg.sentence_num_valid_linkages(self.sent._obj):
+            if self.num == self.sent.num_valid_linkages():
                 raise StopIteration()
             linkage = Linkage(self.num, self.sent, self.sent.parse_options._obj)
             self.num += 1
             return linkage
 
         __next__=next      # Account python3
 
+    class sentence_linkages(object):
```
linas: I don't see the point of this object, at all.

eugeneai: I did not write a comment. Sorry. 
This class is for enumerating linkages found, which are no valid. It is possible in the command utility link-parser, when the following variable is turned on
bad          Display of bad linkages                               0 (Off)

It is useful for debugging input.

Evgeny

bindings/python/linkgrammar.py
eugeneai: May it be just commented out or for mo clearance of understanding be removed ad all.
Should I do it for interface file link_grammar.i or it could be as is?

bindings/python/Makefile.am
eugeneai: It would be better make a separate project that could be reached from Python Package Index, IMO.

linas: That's one way to solve it. The problem is that no one ever maintains the PPI packages, and then they get old, stale, and sooner or later they break, and stay broken. It was easier to just pull things into this project.

eugeneai: I could be dialectically mediated:
Pull things in the project as You said, but create a project with a scripts that constructs PPI packages by copying files to a directory of package.

linas: Sure!  I don't really know how to do that, and am out of free time to mess with that ... if you think you can provide a patch, that would be fine.

PR #342, opened 2016-08-03
Windows cleanups + more

ampli:
Small cleanups.

These and previous commits were extracted from my current work on Windows UTF-8 transparency and locale setup, in order to remove unrelated changes from it and reduce its volume.


linas:
thanks sorry for late response

Issue #337, opened 2016-07-19
Locale in the dictionary


ampli:
I rewrote the locale setup code for  Windows to solve most of the problems, and I finally realized that in addition to that it will be a good idea to implement what said in this comment from `link-parser.c`:
```
	/* Get the locale from the environment...
	 * Perhaps we should someday get it from the dictionary ??
	 */
```
For dictionaries which doesn't specify a locale, the mechanism will be like now (setting the locale from the environment), so the behavior will be backward compatible.

Regretfully we cannot currently use something like:
```
<dictionary-locale>: ru_RU.UTF-8;
````
Or better:
```
<dictionary-locale>: "u_RU.UTF-8";
```
(To use the last format, struct Dict_node_s needs a field addition, which seems an overkill
for the few variables that can be needed.)

So I propose to use:
```
<dictionary-locale>.ru_RU\.UTF-8: ();
```
I found that a very small code change is needed to support that:
1. When checking for idioms, ignore words starting with **<**.
2. When checking for SUBSCRIPT_DOT,  ignore a **\**-escaped dots.

Storing values in the subscript may seem strange, but since a lookup can be done by a word without a subscript, one maybe can see the subscript also as a part of the data about the word.

(Alternatively it can be defined in4.0.affix.) 

I first thought about defining a locale **list** like `ru_RU.UTF-8:Russian", when the first one whose set succeeds will be used, but for supporting Windows, at least, this is not needed, as it is easy to convert this locale format to values acceptable by Windows (using a system query to convert it - the system knows about format like ru-RU even though setting a locale directly to it doesn't work).

BTW, it seems the locale setup is usually not critical on Linux, since en_US.UTF-8 and almost any other UTF-8 locale (but tr_TR.UTF-8) seem to be appropriate for all languages but Turkish (which in principle must use tr_TR.UTF-8 due to the tolower of `"I"`).
However, on Windows the locale (and the code page which is associated with it by default) is critical for correct keyboard input and tolower.

If this is fine by you I will send pull requests that implement it.

**Comment by linas** [comment]
I'm thinking that maybe putting the locale in the dict is overkill -- how hard is it for the user to just specify this at the shell level, for windows?

But -- if we do put it in the dict, I would rather not modify the dict.  It would be better to encode:
```
<dictionary-locale>:  RUvruvUTFv8+;
```
[...]

PR #350, opened 2016-08-14
The dictionary-locale version

ampli:
(See issue #337.)

These changes intended to make it very trivial to work with link-parser on Windows.
No locale or code page setup is needed, and file redirection work fine(for pipe output a workaround is documented).

The Windows version also work mostly fine on Cygwin (besides non-functional ^D/^Z/^C). This version also compiles and runs fine under Cygwin (but running the resulted binary on the Windows console is not supported). The Python binding on Cygwin scceeds, but I yet need to find out the correct PYTHONPATH. No Python binding on MSVC yet.

The library also compiles fine using MinGW under Cygwin, but there is a UTF-8 output problem (truncated strings) that still need a debug.

On Window, link-parser now reads Unicode from the keyboard,  and sets transparent UTF-8 path to the output.

The Windows ifdef's got cleaned up, and now _WIN32 means a native Windows compiler like MSVC or MinGW. There are no Cygwin ifdef's since it is POSIX compliant. In two places there is a detection of Cygwin runtime environment in a WIN32 native compiled binary, in order to support running also under Cygwin.

Many cleanups of the old code are still needed. For example, to my surprise I found that is_utf8_space is not in use in version 5 at all. There are also several too similar utf8 handling functions, and their naming convention is confusing.

The dictionary locale stuff has some documentation in msvc14/README.
A test that check the dictionary-locale setup is added (exploiting the Turkish `I`  problem).

The locale setup is now handled in the library, and programs that use the library don't have to set the locale at all, so this code got removed from link-parser.

I tried to make comprehensive tests.
BTW, a test suite that tests `link-parser` is needed.

### DIFF
--- a/bindings/python-examples/tests.py
+++ b/bindings/python-examples/tests.py
@@ -15,7 +15,7 @@ try:
 except ImportError:
     import _clinkgrammar as clg
 
-locale.setlocale(locale.LC_ALL, "en_US.UTF-8")
+locale.setlocale(locale.LC_CTYPE, "en_US.UTF-8")
 
 def setUpModule():
     datadir = os.getenv("LINK_GRAMMAR_DATA", "")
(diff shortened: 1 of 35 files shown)


linas:
wow.

PR #351, opened 2016-08-14
Restore multi-thread.cc. Clarify comments/doc.

ampli:


linas:
why revert? tests passed, for me.

ampli:
It was just an intermediate "junk" and not even the final version I used for the test.
The final version takes care to make parsing interleaving of the various languages (by inserting usleep()), else each language is parsed as one continuous block. Another problem is that there is a need to inspect the output in order to see that the test is OK, and that it could be much better if it would include a "tr" test sentence that starts with an `I` (I don't have such).

linas:
Uh, careful with inserting usleep() -- this is a system call, it will descend to the kernel, and cause the scheduler to run. This will alter which threads are running, and prevent exactly the kinds of race conditions that the unit test is trying to find. ... or it might expose races that are otherwise hidden (!)

PR #353, opened 2016-08-17
Changes for Windows

ampli:
Main changes:

1. Finally fix the UTF-8 problem in MinGW.  It is fully working now. The Python bindings is not done yet, and the Java bindings has not been tested yet. A readme file of how to configure it under Cygwin is provided.

2. Fix Python bindings for Cygwin. There is a need to rename the DLL to `_clinkgrammar.dll` (still need to send PR for that).

3. Fix tests.py to skip the spell-related testst if spell-guess is not configured. For now it depends on a library change that makes the spell_guess option to be always 0 if it is not configured.

4. Update the main README.

--

Maybe an issue should be opened for the following:

### Problem
One test fails on the Cygwin version (test_quotes - a linkage test). It turned out the linkages are produced just fine - but in a different order (within equal linkage metrics). This bites us from time to time, but until now we just fix the test suite to use another linkage for the comparison.
This cannot be done any more if we would like the test suite to tun on different systems (not only Linux).

Possible solutions:

1.  Sort the linkages in a canonical form, in the library. The idea is that if the linkages metrics are equal, then compare the sentence words, and if they are equal, then compare the links (word by word). This adds overhead, but can be done only if a special parse-option is set.

2. Sort the linkages in a canonical form, in tests.py. The idea is to produce all the graphical linkages in advance, an sort them alphabetically.

3. A complex comparison algorithm in tests.py: produce all linkages and divide them to groups of equal metrics. If the first such group has N linkages, search the first N given linkages in it, and so on.

I think option 1 is the best, because it adds a feature to the library - consistent linkage order, and the test suite can validate the order.

Are there other ways to solve this problem?
What can be done until a solution is implemented?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -46,4 +46,6 @@ EXTRA_DIST =            \
 	msvc14/MSVC-common.props               \
 	msvc14/post-build.bat                  \
 	msvc14/README                          \
+	mingw/README.Cygwin
+	mingw/README.MSYS
 	TODO
--- a/README
+++ b/README
@@ -70,6 +70,7 @@ CONTENTS of this directory:
    configure                The GNU configuration script
    autogen.sh               Developer's configure maintenance tool
    msvc14                   Microsoft Visual-C project files
+   mingw                    Information on using MinGW under MSYS or Cygwin
 
 
 UNPACKING and signature verification:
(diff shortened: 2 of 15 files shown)


linas:
I would prefer solution 2 or 3 -- there is no natural canonical linkage order. One could invent something artificial; but artificial isn't good for natural language.

PR #356, opened 2016-08-24
SAT: Bug fix - preallocate linkages memory

ampli:
Currently the linkage array in the SAT parser is allocated on demand -
for each new linkage it is enlarged by one element, using realloc().

The problem is that realloc() may return a pointer to another block of
memory, and since the API uses linkage pointers, such pointers would
become invalid.

This fix allocates the linkage array in advance, according to the
linkage_limit option.

### DIFF
--- a/link-grammar/sat-solver/sat-encoder.cpp
+++ b/link-grammar/sat-solver/sat-encoder.cpp
@@ -1352,13 +1352,29 @@ Linkage SATEncoder::get_next_linkage()
 
   assert(linkage, "No linkage");
 
+  /*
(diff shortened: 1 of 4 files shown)


linas:
So was something crashing? bad pointers typically mean a crash, somewhere....

ampli:
Yes, I got the problem after I implemented the "canonical" linkage order in tests.py (by an algo which is a combination of 2 and 3 as described in #353). For SAT it means producing all the linkages in advance (not so bad because the test sentence has only a few linkages). I then got a bad memory reference in linkage_print_diagram() on referring linkage->num_links.

Issue #358, opened 2016-08-24
Python imports


ampli:
Currently the following is used:
```
from linkgrammar import Sentence, Linkage, ParseOptions, Link, Dictionary
try:
    import linkgrammar._clinkgrammar as clg
except ImportError:
    import _clinkgrammar as clg
```
The **try** block is used so tests.py can work even before installing LG in the system.

This has the following drawbacks:
1. It is cumbersome.
2. It mentions a module starting with an underscore for no real need.
3. An IDE cannot check the usage of the clg.* functions, as _clinkgrammar is not Python code (it just flags as error all of them, e.g. `Module '_clinkgrammar' has no 'dictionary_set_data_dir' member`).

Drawbacks 2 and 3 can be solved by using **clinkgrammar** instead of **_clinkgrammar**. This is because SWIG also generates **clinkgrammar.py**, which imports **_clinkgrammar**.
However, drawbacks 1 still remains.

In order to make the usage of the Python interface more streamlined, I propose to "hide" the import of **_clinkgrammar**, as it is an implementation detail. This can be done by adding this in **linkgrammar.py**:
```
Clinkgrammar = clg
__all__ = ['ParseOptions', 'Dictionary', 'Link', 'Linkage', 'Sentence', 
           'Clinkgrammar']
```
(The last statement already exists - I only added to it `'Clinkgrammar' `.)

Then, in Python programs **only** the following is needed:
```
from linkgrammar import Sentence, Linkage, ParseOptions, Link, Dictionary, \
    Clinkgrammar as clg
```
Or even just:
```
import linkgrammar
```

I will send a PR for that (unless I missed something).

linas:
Yes, hiding this stuff makes sense.

ampli:
Implemented in #359.

PR #359, opened 2016-08-26
Python test fixes

ampli:
These changes were extracted from my current work (yet unfinished - it needs polishing) on tests.py involving same-cost linkage order independence.

Main changes here:
- Remove dead code.
- Fixes for Windows.
- Hide _clingrammar importing. (Issue #358.)
- Better IDE interaction, including much less pylint warnings and importing of clinkgrammar instead of _clinkgrammar for better pylint checks. (Issue #358.)
- Code cleanup.
- Print program info to validate that the intended locations are tested.
- "make check" now uses python3 is configured.
- Fix minor memory leak.
- Use assertTrue() instead of assert.
- Add several binding calls.
- Convert some comments to docstrings.

### DIFF
--- a/bindings/python/linkgrammar.py
+++ b/bindings/python/linkgrammar.py
@@ -332,6 +332,9 @@ class Linkage(object):
     def link_cost(self):
         return clg.linkage_link_cost(self._obj)
 
+    def disjunct_cost(self):
+        return clg.linkage_disjunct_cost(self._obj)
+
     def link(self, i):
         return Link(self, i, self.word(clg.linkage_get_link_lword(self._obj, i)),
                     clg.linkage_get_link_llabel(self._obj, i),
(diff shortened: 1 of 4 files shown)


bindings/python/linkgrammar.py
```diff
@@ -51,10 +50,6 @@ def __del__(self):
             del self._obj
 
     @property
-    def version(self):
-        return clg.linkgrammar_get_version()
```
linas: well, thee is a dictionary version, and there is also a C library version, and the two can sometimes skew.  The goal of having a library version was to identify users with bad installs.

ampli: Of course.
The function linkgrammar_get_version() is fine and a good idea.
It is just not a Parse Option.
It is still accessible as clg.linkgrammar_get_version().
Did I miss something?

linas: Oh, no I guess I missed something. I didn't see the parse option thingy.

PR #361, opened 2016-08-28
Fix missing dictionary DEFAULTPATH

ampli:
I had a rather bad typo!
Regretfully I don't usually make installation tests.
But just now I made and found this bug.

I will define the installcheck target (or equivalent) to run tests.py with the system pathes, and make sure I run it before sending any PR.

It may be a good idea to issue 5.3.9, and recall 5.3.8 if still possible.
Maybe an issue with a workaround should also be opened.


linas:
Yow!  I will push a 5.3.9 in a minute

linas:
Done. 5.3.9 is out I wonder if this explains my opencog test failures!

Issue #303, opened 2016-05-25
API Parsing Inconsistency


Cutha:
I followed the official [API Guide](http://www.abisource.com/projects/link-grammar/api/index.html) and wrote a simple C++ test program.

For this sentence:
`I feel is the exciter than other things.`

The command-line parser can identify 10 linkages with 3 main variations in terms of the positions of null links:
1. `LEFT-WALL I.p feel.v [is] the [exciter] [than] other things.n . `
2. `LEFT-WALL I.p feel.v [is] the exciter[?].a [than] [other] things.n .`
3. `LEFT-WALL [I] [feel] is.v the exciter[?].n [than] other things.n .`

However, the API is only able to find 4 linkages and all of them are of the 2nd structure, i.e.
`LEFT-WALL [I] [feel] is.v the exciter[?].a [than] other things.n .`
with exciter either being guessed as adjective or noun and some minor differences in the linkage diagram (see output below).

Is the API being updated along with the parser? I looked at link-includes.h and it lacks documentation. I have played with certain parse options such as max null count / disjunct cost etc. but it still doesn't give me the same result as the command-line tool. Also, the [another example] (http://www.abisource.com/projects/link-grammar/api/index.html#preps_example) in the API guide doesn't run either. It includes "constituents.h" which doesn't seem to exist in the lib directory after installation. Below the example, it says "functions used are described in the section on extracting constituent structure" and that link isn't working either. The whole API guide seems to be a messy mix of outdated and new information. 

Below is my test code for reference:
```
Dictionary    dict;
Parse_Options opts;
Sentence      sent;
Linkage       linkage;
int           i, num_linkages;

setlocale(LC_ALL, "");
opts  = parse_options_create();
dict  = dictionary_create_default_lang();

string s = "I feel is the exciter than other things.";

sent = sentence_create(s.c_str(), dict);
num_linkages = sentence_parse(sent, opts);
cout << "Number of null links: " << sentence_null_count(sent) << endl;
cout << "Number of linkages: " << sentence_num_linkages_found(sent) << endl;
cout << "Number of valid linkages: " << sentence_num_valid_linkages(sent) << endl;
cout << "Number of linkages post-processed: " << sentence_num_linkages_post_processed(sent) << endl;

for (int i=0; i<num_linkages; ++i)
{
    linkage = linkage_create(i, sent, opts);
    cout << linkage_print_diagram(linkage, true, 800);
[...]

PR #371, opened 2016-09-03
Parse1step

ampli:
This change fixes a problem that may be an obstacle for API users. See issue #303.
In verbosity>0, it mimics the current link-parser printing of "No complete linkages found."
when no parse is found with null_count==0.

A test for this fix is added, and python-examples/example.py is updated to make a 1-step parse.

Main changed functions:
- chart-parse
- prepare_to_parse

Added functions:
- connectors_dup
- disjuncts_dup

The following functions moved up in the source files:
- free_sentence_disjuncts
- disjuct_gwordlist_append

### DIFF
--- a/bindings/python-examples/example.py
+++ b/bindings/python-examples/example.py
@@ -32,13 +32,21 @@ def linkage_stat(psent, lang):
                  psent.num_valid_linkages(), random))
 
 
+en_lines = [
+    'This is a test.',
+    'I feel is the exciter than other things', # from issue #303 (10 linkages)
+]
+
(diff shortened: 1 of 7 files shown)


linas:
This breaks the !bad flag -- the post-processing rules sometimes cause the number of valid linkages to be zero, and so then parsing continues onwards with a null-count greater than zero.  However, if the !bad flag is set, then parsing should NOT continue to larger null-counts.

This capability is critical for debugging.

linas:
Never mind, this was broken earlier; I just noticed it now.

PR #388, opened 2016-09-11
link-parser fixes: Fix  -max-cost  and allow  -panic-*

ampli:
Fix the following:
1. max-cost cannot not be set from the command line.
2. No way to set -panic-* from the command line. 

After this fix, link-parser uses the disjunct_cost as set by the library.

### DIFF
--- a/link-parser/link-parser.c
+++ b/link-parser/link-parser.c
@@ -87,28 +87,10 @@ typedef enum
 	NO_LABEL = ' '
 } Label;
 
-static char *
-fget_input_string(FILE *in, FILE *out, bool check_return)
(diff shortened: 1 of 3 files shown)


link-parser/link-parser.c
```diff
@@ -666,12 +669,6 @@ int main(int argc, char * argv[])
 		}
 	}
 
-	/* The English and Russian dicts use a cost of 2.7, which allows
-	 * regexes with a fractional cost of less than 1 to be used with
-	 * rules that have a cost of 2.0.
-	 */
-	parse_options_set_disjunct_cost(opts, 2.7);
```
linas: In the long run, the default disjunct cost should be encoded in the dictionary.

PR #389, opened 2016-09-11
Windows fixes

ampli:
The dictionary locale setting fix is general, but the garbage locale name printing was observed only on Windows.

### DIFF
--- a/link-grammar/dict-file/dictionary.c
+++ b/link-grammar/dict-file/dictionary.c
@@ -518,14 +518,14 @@ dictionary_six_str(const char * lang,
 	/* If we didn't succeed to set the dictionary locale, the program will
 	 * SEGFAULT when it tries to use it with the isw*() functions.
 	 * So set it to the current program's locale as a last resort. */
-	if (NULL == dict->locale)
+	if (NULL == dict->locale_t)
 	{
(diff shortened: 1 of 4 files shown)


link-grammar/dict-file/dictionary.c
```diff
@@ -518,14 +518,14 @@ dictionary_six_str(const char * lang,
 	/* If we didn't succeed to set the dictionary locale, the program will
 	 * SEGFAULT when it tries to use it with the isw*() functions.
 	 * So set it to the current program's locale as a last resort. */
-	if (NULL == dict->locale)
+	if (NULL == dict->locale_t)
 	{
 		dict->locale = setlocale(LC_CTYPE, NULL);
-		dict->locale_t = newlocale_LC_CTYPE(setlocale(LC_CTYPE, NULL));
+		dict->locale_t = newlocale_LC_CTYPE(dict->locale);
```
linas: Calling a member of a struct "locale_t" seems like a dangerous thing to do -- because locale_t is already a typedef.

I'm sure that the compiler can handle this correctly; its just for humans editing the code, it's a potential source of confusion and errors.

Can you change this to something like locale_ctype or something similar?

PR #391, opened 2016-09-11
Change Dictionary_s locale_t  to ctype

ampli:
In the same occasion, add descriptive comments and fix typos.

### DIFF
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -297,8 +297,8 @@ struct PP_data_s
 	size_t vlength;                 /* Length of visited array */
 };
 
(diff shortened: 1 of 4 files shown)


linas:
I had some merge conflicts with this, so I re-wrote the code a little bit:  it now says this:
```
#ifdef HAVE_LOCALE_T
   dict->lctype = newlocale_LC_CTYPE(dict->locale);

   /* If dict->locale is still not set, there is a bug. */
   assert((locale_t) 0 != dict->lctype, "Dictionary locale is not set.");
#else
   dict->lctype = 0;
#endif /* HAVE_LOCALE_T */
```

PR #392, opened 2016-09-13
Move up set_utf8_program_locale()

ampli:
This is needed in order to establish the program locale early,
so setlocale(LC_CTYPE, NULL) will return the correct locale.

This matters for dictionaries that don't define the locale, when
the user's locale doesn't have UTF-8 codeset.

Test by:
```
LANG=NoSuchLocale link-parser any
```
(Dictionary locale is set to C instead of to en_US.UTF-8.)

Because the locale setting stuff is subtle, add detailed comments.
Also rearrange and improve existing comments.

### DIFF
--- a/link-grammar/dict-file/dictionary.c
+++ b/link-grammar/dict-file/dictionary.c
@@ -505,16 +505,29 @@ dictionary_six_str(const char * lang,
 		return dict;
 	}
 
-	/* Get the locale from the dictionary. */
+	/* Get the locale for the dictionary. The first one of the
+	 * following which exists, is used:
+	 * 1. The locale which is defined in the dictionary.
+	 * 2. The locale from the environment.
+	 * 3. On Windows - the user's default locale.
+	 * NULL is returned if the locale is not valid.
+	 * Note:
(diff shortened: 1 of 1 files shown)


link-grammar/dict-file/dictionary.c
```diff
@@ -553,17 +571,17 @@ dictionary_six_str(const char * lang,
 	 */
 	if (read_regex_file(dict, regex_name)) goto failure;
 
-	const char *locale = setlocale(LC_CTYPE, NULL);
-	locale = strdupa(locale); /* setlocale() uses static memory. */
+	const char *locale = setlocale(LC_CTYPE, NULL); /* Save current locale. */
+	locale = strdupa(locale); /* setlocale() uses its own memory. */
```
linas: so strdupa is available on windows, mac?

linas:
so does this all work on mac os?

ampli:
>so strdupa is available on windows, mac?

It is defined in utilities.h (unless it is already defined, like on Linux.)

>so does this all work on mac os?

I tested this branch on Windows and Mac just before sending the PR, because I now try  to do extended tests after each non-trivial change.

PR #394, opened 2016-09-13
Solver.h: Fix format errors

ampli:
The offending code uses %zu to print "unsigned long long", which just
happens to be the same size as size_t on Linux.

There is a strange use of dots and a strange number alignment, but I basically left it intact (beside a small fix).

After this fix, the SAT solver works fine on Mac too.

(I still need to send a PR for replacing MiniSAT with a modern equivalent, a thing that I already implemented but it needs a finishing touch, and maybe replacing by yet newer code.)

### DIFF
--- a/link-grammar/minisat/Solver.h
+++ b/link-grammar/minisat/Solver.h
@@ -22,6 +22,7 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 
 #include <cstdio>
 #include <map>
+#include <inttypes.h>
 
 #include "Vec.h"
 #include "Heap.h"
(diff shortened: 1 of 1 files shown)


linas:
gahh. How ugly.

linas:
So, anyway, I think its OK to publish version 5.3.10 now, right? I think this was the last outstanding OSX bug...

ampli:
> gahh. How ugly.

Any better idea?

> So, anyway, I think its OK to publish version 5.3.10 now, right? I think this was the last outstanding OSX bug...

Just now `make check` before installation doesn't work because the `_clingrammar.so` module and the `tests/*` programs cannot find `liblink-grammar.dylib` in the build tree (after the installation it works fine).
In previous versions of OS X these could be fixed by a proper `DYLD_LIBRARY_PATH`.
However, in El-Capitan this method doesn't work any more.
This can be fixed by adding an rpath. It could be a good idea to add it only in debug mode.

But this can be for the next version. So I think it's OK to publish 5.3.10 now.

linas:
> gahh. How ugly.
>
> Any better idea?
>
Convince Apple to do it the Gnu way?

> But this can be for the next version. So I think it's OK to publish 5.3.10
> now.
>
Here it comes!

PR #400, opened 2016-09-16
Python bindings + more

ampli:
Main fixes:
1. Add LG_TimerExpired exception.
2. Add tests for it and for the previously added LG_DictionaryError exception.
3. Avoid assertion failure in debug mode if a timeout happens during the PP.

Note that now tests.py prints
```
link-grammar: Error: Could not open dictionary No such language/4.0.dict
```
This is currently unavoidable because this error is printed unconditionally when a dict is not found.

Related comment:
In order that the library will be able to get embedded seamlessly in applications (like grammar checkers or GUIs) it should not print such error messages, but instead only provide an error indication, and put the error in a variable (a per-thread variable if multi-threaded), to be retrieved by a new function link-grammar_error(). Such an arrangement is found in other libraries,.

Linas, if implementing such an error mechanism seems to you fine, I will send a PR for that.
I can provide a compatibility feature to prevent the need to rewrite existing applications  in order to print the error messages. So link-parser.c will not need a rewrite in order to report such error messages.

### DIFF
--- a/bindings/python-examples/tests.py
+++ b/bindings/python-examples/tests.py
@@ -294,6 +294,13 @@ class DBasicParsingTestCase(unittest.TestCase):
         linkage = self.parse_sent("This is a _regex_ive regex test")[0]
         self.assertEqual(linkage.word(4), '_regex_ive[!].a')
 
+    def test_timer_exhausted_exception(self):
+        self.po = ParseOptions(max_parse_time=1)
+        self.assertRaises(LG_TimerExhausted,
+                          self.parse_sent,
+                          "This should take more than one second to parse! " * 20,
+                          self.po)
+
 class ESATsolverTestCase(unittest.TestCase):
     def setUp(self):
         self.d, self.po = Dictionary(lang='en'), ParseOptions()
--- a/bindings/python-examples/tests.py
+++ b/bindings/python-examples/tests.py
@@ -45,7 +45,7 @@ def setUpModule():
 # The tests are run in alphabetical order....
 #
 # First test: test the test framework itself ...
-class AAALinkTestCase(unittest.TestCase):
+class AALinkTestCase(unittest.TestCase):
     def test_link_display_with_identical_link_type(self):
         self.assertEqual(str(Link(None, 0, 'Left','Link','Link','Right')),
                          u'Left-Link-Right')
@@ -53,6 +53,10 @@ class AAALinkTestCase(unittest.TestCase):
     def test_link_display_with_identical_link_type2(self):
         self.assertEqual(str(Link(None, 0, 'Left','Link','Link*','Right')),
                          u'Left-Link-Link*-Right')
+
+class AAADictionaryTestCase(unittest.TestCase):
+    def test_open_nonexistent_dictionary(self):
+        self.assertRaises(LG_DictionaryError, Dictionary, 'No such language')
 
 class BParseOptionsTestCase(unittest.TestCase):
     def test_setting_verbosity(self):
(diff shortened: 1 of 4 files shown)


bindings/python-examples/tests.py
```diff
@@ -53,6 +54,10 @@ def test_link_display_with_identical_link_type2(self):
         self.assertEqual(str(Link(None, 0, 'Left','Link','Link*','Right')),
                          u'Left-Link-Link*-Right')
 
+class AAADictionaryTestCase(unittest.TestCase):
+    def test_open_nonexistent_dictionary(self):
+        self.assertRaises(LG_DictionaryError, Dictionary, 'No such language')
+
```
linas: Since tests seem to be run in alphabetical order, it would be best to keep the code in alphabetical order too.

linas:
Re exceptions -- there are several possibilities.
* use C++ to compile everything, and just throw C++ exceptions.
* Keep using C, and write a C wrapper for a C++ exception, and call that wrapper to throw. 
* Use the __cxa_allocate_exception() and __cxa_throw() calls directly.

It seems like option 2 is the best. 

Anything that is an assert should stay as an assert, since that indicates a program bug. Anything that prints an error and exist should throw an exception instead.  So, yes, there should be no calls to exit() in the library.

bindings/python-examples/tests.py
ampli: I will send a PR to fix that.

ampli:
My intention was to keep using C, and **not to use exceptions at all** for the library.

The idea is to:
1. Return error indications **exactly** like now (e.g. NULL for unable to create dict, negative number for unable to split, etc.).
2. Store the error message in a variable (per thread if compiled with thread support).
3. `link_grammar_error()` will return this error message (or NULL if no error).
4. To be compatible with existing programs, the error message will automagically printed if `link_grammar_error()` is not used. This means no changes will be needed in link-parser.c.

1-3 is similar to what is provided by other C libraries. They also generally supply numerical error codes  when an error happens, but providing it needs an API change and I don't find it too useful for now for the LG library.

Anything else seems to me a too big change, that needs too much efforts,  including heavy changes in all the language bindings to catch and translate such exceptions.

So I suggest that I implement it in the simple and easy way I described above.

linas:
Oh, OK.
In step 4, how does the automagic work?  Also, its OK to change link-parser.c.

I observed recently that exit() is called in a bunch of places, and this could be replaced by exceptions (wrapped c++ code).  However, that is low priority.

ampli:
> In step 4, how does the automagic work? Also, its OK to change link-parser.c.

It is possible to do it in several ways:
1. Call linkgrammar_error() once at the start of the program (or start of each thread). This will tell the library not to use the default internal error printing.
2. Provide an additional function, e.g. linkgrammar_error_autoprint() to indicate that auto-print of error messages (like now) is desired.
3. Provide an external variable, to be assigned the address of a printing function. If desired, the library can initialize it to a default printing function, and the user can assign it NULL if error auto-print is not desired.

(1) and (3) (with the indicated default) is compatible to the current library.
(2) and (3) can be combined, by providing linkgrammar_error_autoprint(print_function) (to be called with NULL if linkgrammar_error() is to be used).

There is a question what to do with the current auto-print **informational** messages. In an embedded application, it is desired to "catch" them too. I suggest to let the API user to collect them (of course this is needed only if the said auto-print is not in effect) by just calling linkgrammar_error() after the library calls:
```
    num_linkages = sentence_parse(sent, opts);
    error_message = linkgrammar_error();
    if (error_message) gui_status(error_message);
    if (num_linkages < 0)
    {
        sentence_delete(sent);
        sent = NULL;
        continue;
    }
```

Regarding **exit()**:
The simplest way be to call a configurable function of the user, defaulted to exit().
The user will be able to change it by something like:
linkgrammar_set_exit(myexit);
when myexit() is a user supplied function. This function can call linkgrammar_error() to get the error message, and then call, for example, pthread_exit().

linas:
seems reasonable to me

PR #410, opened 2016-09-18
Convert  exit  to error indication

ampli:
Main changes:
- 	"...a word that is too long": Change exit to error indication
- 	Word_file_struct: Change "file" to a pointer

### DIFF
--- a/link-grammar/dict-file/dictionary.c
+++ b/link-grammar/dict-file/dictionary.c
@@ -594,7 +594,9 @@ dictionary_six_str(const char * lang,
 	dict->empty_word_defined = boolean_dictionary_lookup(dict, EMPTY_WORD_MARK);
 
 	dict->base_knowledge  = pp_knowledge_open(pp_name);
+	if (NULL == dict->base_knowledge) goto failure;
 	dict->hpsg_knowledge  = pp_knowledge_open(cons_name);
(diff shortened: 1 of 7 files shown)


ampli:
Add exit-to-error conversion for pp_knowledge.

ampli:
> Add exit-to-error conversion for pp_knowledge.

I still need to complete it - plenty of exit() calls in pp_lexer.c.

ampli:
exit() calls that still exist, because they aren't worth the effort to convert or there is no need to, or I don't know what should be done with them:
1) In utilities.c:
```
static int wctomb_check(char *s, wchar_t wc)
...
	nr = wcrtomb(s, wc, &mbss);
	if (nr < 0) {
		prt_error("Fatal Error: unknown character set %s\n", nl_langinfo(CODESET));
		exit(1);
	}
```
Maybe just just notify and arrange to return with no effect from the functions in which it is called (upcase/downcase) so the words will remain unchanged.

2) In pp_lexer.c, when malloc(0 fails.

3) In the viterbi code.

4) In the corpus code.


In addition to the conversion to error indications, I also handled memory cleanup in order to prevent memory leaks when an error occurs.

5. In malloc-dbg.c.

6. in xalloc()/exalloc().

linas:
OK .. I'm not sure how to handle the cases above, either

PR #416, opened 2016-09-22
Python-bindings example of checking for possible grammar problems

ampli:
I wrote this example for an answer to a recent [stackoverflow question](http://stackoverflow.com/questions/38545726/how-to-use-the-link-grammar-parser-as-a-grammar-checker). However, the example is to big for including in the answer as a code snippet, so I am sending it as PR and will refer it later in my answer.

This example shows that the current library is still not appropriate for the task - there is no programmatic way to find the original sentence words to which the linkage words refer.
In simple cases there -s 1-1 correspondence. However, a problem happens if an original sentence word get split (like "it's", or a punctuation strip), or get spell-guessed.

My proposed solution is to create an API which exposes the Wordgraph. Using it, one can trace the result linkage words to the original sentence words. There is a question how to provide an API to access a graph, as exposing the bare data structure of the Wordgraph seems as a bad idea.
What is the right way to do that?

### DIFF
--- a/bindings/python-examples/sentence-check.py
+++ b/bindings/python-examples/sentence-check.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python
+"""
+Demo: Find unlinked or unknown words.
+These demo is extremely simplified.
+It can only work with link-grammar library version >= 5.3.10.
+Input: English sentences, one per line.
+Output: If there are any []-marked words in the linkage results,
+the output contains unique combinations of the input sentence with
+these works marked.  No attempt is done to handle the walls.
+Spell guesses are not handled in this demo.
+
+Example:
+This is a the test of bfgiuing and xxxvfrg
+Output:
+Sentence has 1 unlinked word:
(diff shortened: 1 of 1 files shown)


linas:
Huh. Interesting, never thought of that.  By "the word in the original sentence", I guess you mean  "that token in the original sentence that I would get, if I split the original sentence by using blank spaces as word-separators"  .. Phew. That's a mouth-full.

For the case of "it's" being split into two words, the user could write a utility themselves, its not hard. For spell-guessing, it gets hard, because string compares are no longer possible.

Yes, I guess something like this might work:
```
char * linkage_get_orig_word(const Linkage linkage, WordIdx word_num);
```  
and maybe also
```
int linkage_get_orig_index(const Linkage linkage, WordIdx word_num);
```

Can you think of a better name for "orig_word" and "orig_index"? There must be some nicer way of saying this...

linas:
opened issue #420 to track this idea.

PR #417, opened 2016-09-25
MiniSAT 2.2 upgrade

ampli:
Solving issue #415.

Main changes:
- Upgrade the bundled minisat code to version 2.2 (from the GitHUb of its authors).
- Fixed in it some problem (it is not maintained by its authors).
- Use the system minisat library by default.
- Add an option --enable-sat-solver=bundled .
- Update the README and Changelog.
- Update the RPM spec file to require minisat2-devel .
- Fix a test suite bug regarding pre-sorting of SAT linkages (it fixes a SAT test failure).

Tested on:
Fedora 23, with minisat2 version 2.2.0-12.fc23 and also the bundled version.
Ubuntu 16.04, with minisat2 version 2.2.1-5 and also the bundled version.
OS X Sierra, with the bundled minisat version only (no minisat library package is available).

It is not working on Windows yet (porting is needed).


linas:
Thanks Amir!

PR #426, opened 2016-09-28
Improve debug and error messages on a bad locale

ampli:
... to debug a problem on Ubuntu.

The problem was that tests.py failed badly in many places, as if the
locales are not defined in the dictionaries.

It turned out that only en_US.UTF-8 was on the system.  Installing all
the other locales that are in the current dictionaries solved the
problem.


linas:
Heh. Good one.

PR #429, opened 2016-09-30
Fixes for configuring the minisat bundled code

ampli:


ampli:
I tested these fixes on Debian (unstable), Fedora 23, and OS X Sierra.

configure.ac
```diff
@@ -333,6 +336,16 @@ if test -n "$use_minisat_bundled_library"; then
...
+	# use c++ to compile stdio.h and bomb if it fails.
+	AC_LANG([C++])
+	AC_CHECK_HEADER([stdio.h],,
+		[AC_MSG_ERROR([C++ compiler not found; it is needed for the SAT parser])])
+	if test x${apple_osx} != xyes; then
+		AC_CHECK_LIB(stdc++, main)
+	fi
+	AC_LANG([C])
+
 	CPPFLAGS="${CPPFLAGS} -DUSE_SAT_SOLVER=1"
```
linas: I think this check needs to be moved up higher in the file ... if a C++ compiler is not installed, then earlier parts of the configure script will crash.

Issue #419, opened 2016-09-26
Configuring Python2 and Python3 bindings at the same time


ampli:
I propose to add the ability to configure both Pyhon bindings at the same time.
The configure script will just use the version it finds, and if it finds both, it will use both.

The problem is that the provided autoconf Python detection macros cannot do that.
Other packages overcame that by providing their own autoconf code for that.

For example: [libpeas/autoconf.ac](https://github.com/gregier/libpeas/blob/8cae6abc21bb2ca18533a3f1806288be633cfc22/configure.ac#L361)

I propose to copy it (it is LGPL'ed).
If this is a good idea, I will send a PR.

BTW, I thought than then it can be a good idea to make the scripts run by Python3 if it is availble, but I don't know how to do that besides having tests.py.in etc., which is cumbersome and and makes developing harder.

linas:
Sigh. As if our autoconf file wasn't too big already.  Sure.

Also, we should enable a build of python bindings by default, if python headers/libs are installed.

ampli:
I still need to implement this:
>FIXME: If both versions cannot be configured, a configuration error should be generated unless --disable-python-bindings is given.

PR #432, opened 2016-10-01
Dual python by default

ampli:
See issue #419.

Main changes:
- Provide Python bindings by default, to both versions 2  & 3 (if instaleld).
- If one of them (or both) are not installed, no error happens.
- If it is desired to provide only bindings to one version, `--enable-python-bindings=N" can be used.

**FIXME**: If both versions cannot be configured, a configuration error should be generated unless `--disable-python-bindings` is given.

In order to configure both versions and one, **unusual** ways are used (maybe for the first time - I couldn't find any prior such implementations), involving these methods:
1. Since AM_PATH_PYTHON/ AX_PYTHON_DEVEL create cached variables and others that prevent using them again, a mechanism is introduced to forget them. This is a general idea for which I couldn't find any macro implementation, so I included it as general autoconf m4 macros in `m4/varcheckpoint.m4`.
2. Since AX_PYTHON_DEVEL hard fails if the Python version it tries to detect doesn't cannot be configured, AC_MSG_ERROR is temporarily converted to a "**soft error**".
3. Since AX_PYTHON_DEVEL defines HAVE_PYTHON, a redefinition problem happens if both versions are configured. The solution is to discard it from `confdefs.h`, since this definition is not used.

**FIXME?** `m4` was in `.gitignore`. But I have created `m4/varcheckpoint.m4` which cannot be ignored. So I ignored instead the other files in `m4`. However, some other files can be added there automagically and not be ignored. Maybe it is better to ignore `m4` and just unignore `m4/varcheckpoint.m4`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,12 @@ lg_*_wrap.cc
 Makefile.in
 Makefile
 missing
-m4
+m4/libtool.m4
+m4/ltoptions.m4
+m4/ltsugar.m4
+m4/ltversion.m4
+m4/lt~obsolete.m4
+m4/._*.m4
 .pc
 patches
 py-compile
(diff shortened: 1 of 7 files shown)


linas:
Wow. Amir, the consummate programmer.

PR #441, opened 2016-10-14
configure.ac: Improve minisat checks

ampli:
(The first 5 commits has already been applied. I don't know how to remove them from this PR - `git push -f` didn't help.)


linas:
Yeah, github did not used to behave that way, but they changed something about how repeated pulls from  a branch are displayed, which causes committed diffs to appear s if they are not committed. This started happen for me about 2-3 months ago, I think.

Issue #442, opened 2016-10-16
5.3.11 fails to build with system minisat


EvaSDK:
Hello there,

I am one of the Gentoo maintainers of link-grammar and following #415, I tried to build 5.3.11 against our system installation of minisat. We have two releases, basically the 2.2.0 release and the latest commit from their git tree from September 25th, 2013. The 2.2 release is simply not detected as valid, I guess some required features for link-grammar are missing so it falls back to bundled lib. With git HEAD though it fails with:

```
link-grammar-5.3.11  build configuration settings

    prefix:                         /usr
    C compiler:                     x86_64-pc-linux-gnu-gcc  -DUSE_PTHREADS=1 -DUSE_SAT_SOLVER=1 -DHAVE_SQLITE=1 -O2 -march=ivybridge -pipe -g -msse3 -mstackrealign -O3 -std=c11 -D_BSD_SOURCE -D_SVID_SOURCE -D_GNU_SOURCE -D_DEFAULT_SOURCE
    C++ compiler:                   x86_64-pc-linux-gnu-g++  -DUSE_PTHREADS=1 -DUSE_SAT_SOLVER=1 -DHAVE_SQLITE=1 -O2 -march=ivybridge -pipe -g -msse3 -mstackrealign -O3 -Wall -std=c++11
    Posix threads:                  yes
    Editline command-line history:  no
    UTF8 editline support:          no
    Java libraries:                 no
    Java interfaces:                
    Swig interfaces generator:      yes
    Perl interfaces:                no
    Perl install location:          
    Python2 interfaces:             no
    Python3 interfaces:             no
    ASpell spell checker:           no
    HunSpell spell checker:         yes
    HunSpell dictionary location:   /usr/share/myspell
    Boolean SAT parser:             yes
    SQLite-backed dictionary:       yes
    Viterbi algorithm parser:       no
    Corpus statistics database:     no
    RegEx tokenizer:                no

>>> Source configured.
>>> Compiling source in /var/tmp/portage/portage/dev-libs/link-grammar-5.3.11-r1/work/link-grammar-5.3.11 ...
make -j16 -l6.0 
Making all in link-grammar
make[1]: Entering directory '/var/tmp/portage/portage/dev-libs/link-grammar-5.3.11-r1/work/link-grammar-5.3.11-normal/link-grammar'
Making all in sat-solver
[...]

PR #444, opened 2016-10-19
configure minisat: Fix C++ function check problem on Gentoo

ampli:
(Solving issue #442.)

"AC_CHECK_DECL([mkLit], ..." uses the C++ code line "(void) mkLit;"
which causes the errro:
conftest.cpp:44:10: error: conversion to void cannot resolve address of overloaded function
   (void) mkLit;
(gcc version 4.9.3 (Gentoo 4.9.3 p1.5, pie-0.6.4))

Solve this by specifying its arguments.
(This function has only one definition. On higher gcc versions I didn't
encounter this problem.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -344,7 +344,7 @@ if test "x$enable_sat_solver" = xyes; then
 		dnl Adapt to slight variations in minisat2.2 code.
 
 		# The function for making a literal can be Lit() or mkLit()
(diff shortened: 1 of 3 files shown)


ampli:
I just pushed here a commit to statically  link the bundled minisat library.

After this change, libminisat is not installed in the system any more.
This solves the problem of clashing with the systems libminisat
or crashes when both are installed and the wrong one is used (as happened on Gentoo, #442).

Issue #449, opened 2016-10-26
make-check.py not in 5.3.11 distributable


errantlinguist:
On the AbiWord website, the current "stable" version is listed as 5.3.11 (see [listed download link](http://www.abisource.com/downloads/link-grammar/5.3.11/link-grammar-5.3.11.tar.gz)). In this version, the usage of the script file `make-check.py` is mentioned in the `README` for using Python bindings, but this file is not included in the distributable.

ampli:
Thanks.
I have just set a fix.
Meanwhile you can use [make-check.py](https://github.com/opencog/link-grammar/blob/83e5c9bfe07a3dd816d4df2fa2b7579658c85ef9/msvc14/make-check.py) from the GitHub repository (save the "Raw" version).

errantlinguist:
Should the script work without any modification?-- I can't test the setup on a Linux machine at the moment, but after building the parser using Visual Studio/C++, I get

    LINK_GRAMMAR_DATA=..
    Traceback (most recent call last):
      File "C:\Users\tcshore\src\link-grammar-5.3.11\msvc14\..\bindings\python-examples\tests.py", line 20, in <module>
        from linkgrammar import (Sentence, Linkage, ParseOptions, Link, Dictionary,
      File "C:\Users\tcshore\src\link-grammar-5.3.11\bindings\python\linkgrammar.py", line 13, in <module>
        import clinkgrammar as clg
      File "C:\Users\tcshore\src\link-grammar-5.3.11\bindings\python\clinkgrammar.py", line 28, in <module>
    _clinkgrammar = swig_import_helper()
      File "C:\Users\tcshore\src\link-grammar-5.3.11\bindings\python\clinkgrammar.py", line 20, in swig_import_helper
        import _clinkgrammar
    ImportError: DLL load failed: %1 is not a valid Win32 application.

Using the batch-file demo file `link-parser.bat` works without any problems, however, so must something not be wrong with SWIG and/or the Python bindings themselves?

ampli:
Yes, it should work without any modifications (I have just checked it on 5.3.11, with both Python2 & 3).
My tests environment:
- Visual Studio 2015.
- x64 machine.
- Link grammar compiled as x64 (I tried both Release and Debug).
- Python versions:
 - `Python 2.7.12 (v2.7.12:d33e0cf91556, Jun 27 2016, 15:24:40) [MSC v.1500 64 bit (AMD64)] on win32`
[...]

PR #451, opened 2016-10-27
Windows fixes

ampli:
Main Fixes:
1) Add LG_DLLPATH User Macro to solve issue #449.
2) Fix Python3 UTF-8 problem in `tests.py`.
The problem is that Windows locales don't have a UTF-8 codeset, and two of the "parses-*.txt" files contain UTF-8 characters which cause a Python3  decode problem (Python2 interprets the files as they are). On Linux this doesn't happen because the library takes care to ensure that the library locale's codeset is UTF-8.


linas:
Thanks

PR #453, opened 2016-11-12
Changelog: Update recent changes

ampli:
I propose to release 5.3.12 now.
The next changes that I would like to send are the improved error facility, that I propose to include in a yet new release (5.3.13).

---
* Fix locale_t use for the newly introduced Cygwin 2.6.0.
* Include in the distribution the missing make-check.py (for Windows).
* Minisat configuration improvements + fix a problem on Gentoo.
* When using the bundled minisat, link it statically and don't install it.


ampli:
I just sent PR #454 & PR #455, to also be included in 5.3.12.

PR #456, opened 2016-11-13
Fix a bogus enum Clang warning

ampli:
Also for 5.3.12.
build-disjuncts.c:356:29: warning: comparison of constant 4 with expression of
  type 'Exp_type' is always true [-Wtautological-constant-out-of-range-compare]


linas:
OK, I'll package 5.3.12 tomorrow (in about 18-24 hours)

PR #465, opened 2016-12-08
Add a missing library linkage dependency: link-grammar.def

ampli:
I added functions to link-grammar.def and noted that issuing `make` afterwards just did nothing.

This patch adds `link-grammar.def` as a library linkage dependency.

Check by:
```
make
touch link-grammar/link-grammar.def
make
```


linas:
thanks ... I see three other pull reqs, I will try to look at them tomorrow.

PR #466, opened 2016-12-11
Fix using stale stack memory for string command-line arguments

ampli:


link-parser/command-line.c
```diff
@@ -233,84 +230,80 @@ static int x_issue_special_command(const char * line, Command_Options *copts, Di
...
-		else
-		{
-			/* Found an abbreviated command, but it wasn't a boolean */
-			/* Replace the abbreviated command by the full one */
-			/* Basically, this just fixes !v and !h for use below. */
-			strcpy(s, user_command[k].s);
-		}
-	}
 
-	if (strcmp(s, "variables") == 0)
```
linas: I will merge this pull-req as is, but let me first mention a coding technique that I like that I call "indentation power reduction".  In the above, there would be `if (count == 0) return something;` and all of the rest of the code that follows can be indented one space less.  I find that that fewer indentations makes code easier to read, easier to understand. I can't quite tell if that would work here, but it seems like it might.   Anyway, its not that important, here, its just a nice trick.

Issue #468, opened 2016-12-24
Eliminating the empty-word device


ampli:
I succeeded to eliminate it.
It turned out the needed modification was rather small.
The main change is in do_count(). The rest is for supporting it.

**structures.h**:
```c
struct Word_struct
{
...
	bool optional;       /* Linkage is optional. */
}
```
**tokenize.c**:
```c
if (!empty_word_encountered)
{
#if 0
					if (!sent->dict->empty_word_defined)
						prt_error("Error: %s must be defined!\n", EMPTY_WORD_DOT);
					if (!determine_word_expressions(sent, empty_word(), &ZZZ_added))
						error_encountered = true;
#endif
					sent->word[sent->length - 1].optional = true;
...
}
```
**prune.c**:
```c
static bool optional_gap(Sentence sent, int w1, int w2)
{
	for (int w = w1+1; w < w2; w++)
		if (!sent->word[w].optional) return false;

	return true;
}

static bool possible_connection(...)
{
...
	if (no_null_links &&
	    (lc->next == NULL) &&
	    (rc->next == NULL) &&
	    (!lc->multi) && (!rc->multi) &&
	    !optional_gap(sent, lword, rword))   // <-----------
	{
		return false;
	}
...
}
```
**count.c**:
```c
/**
 * Return the number of optional words strictly between w1 and w2.
 */
static int num_optional_words(count_context_t *ctxt, int w1, int w2)
{
	int n = 0;

	for (int w = w1+1; w < w2; w++)
		if (ctxt->local_sent[w].optional) n++;

	return n;
}

static Count_bin do_count(...)
...
{
...
			if (null_count == (rw-lw-1) - num_optional_words(ctxt, lw, rw))
...
}

```

**Results**:
1). The run-time saving seems to be small (~1%, even for Russian).
This is strange since a test I did (reported in issue #250) shows that setting the `ZZZ` distance during expression-pruning causes a vast speedup.
2). This change automatically solves issue #241 (Disjunct strings sometimes have ZZZ in them).

**Adding `ZZZ+`**:
Note that a `ZZZ` connector (not related to the empty-word device) is still needed for handling quotes.
Currently, when a quote token appears (identified in the code by its `ZZZ-`), a `ZZZ+` is inserted on the previous word.
[...]

PR #469, opened 2016-12-25
Empty word elimination

ampli:
See issue #468.

**This is only for a demo**, as the SAT-parser malfunctions now for sentences that used to have an empty-word in them.

Notes:
1) It seems to fix the current bogus null count that happens when empty words are counted as null-words.
2) An original bug with LL link is now more apparent, as there are no empty-words "guards" to prevent malformed linkages between a root and a suffix of another word, in case the suffix of that root happens to not have a linkage.
E.g. (real example):
```
>     +--------------------------------------Xp-------------------------------------+
>     +-----------------------Wd----------------------+                             |
>     |                            +-------MVIag------+         +------Jd------+    |
>     |       +--------LLFHW-------+        +--LLFYZ--+----Ew---+      +-LLAAQ-+    |
>     |       |                    |        |         |         |      |       |    |
> LEFT-WALL али.= [са] [стрел] =ой.nlfsg взле.= =тела.vsndpfs по.jd трап.= =у.ndmsd . 
```
Possible solution:  Set LL links to length limit 1.
3) A current bug with empty-word handling is automatically corrected (because empty words are not used).
It may happens only for sentences which don't have a full linkage.
Here is how:
ZZZ link are set with length limit 1.
If a word is a null word, the empty word after it cannot be attached to it.
So it cannot be attached at all.
Even without setting length limit 1 it cannot be attached to the previous word (if there exists so not null word) due to an optimization of attaching ZZZ+ to the previous word on demand, so the word before the said null word doesn't have the needed ZZZ+. But even without this optimization there may be a bug in that case when the word before the said null word is also an empty-word, since it doesn't have ZZZ+.
Test sentence:
`In Xanadu did Kubla Kahn a stately pleasure dome decree.`
(There is currently an empty word in it due to did -> "d id", because of unit separation, that we also need to discuss.)

### DIFF
--- a/data/en/4.0.dict
+++ b/data/en/4.0.dict
@@ -56,7 +56,7 @@
 % FIXME: A better comment maybe.
 % See also EMPTY-WORD.x for the highly-unusual situation that EMPTY-WORD
 % appears in the input text.
-EMPTY-WORD.zzz: ZZZ-;
+%EMPTY-WORD.zzz: ZZZ-;
 
 % Capitalization handling (null effect for now- behave as empty words).
 1stCAP.zzz: ZZZ-;
--- a/data/en/4.0.dict.m4
+++ b/data/en/4.0.dict.m4
@@ -65,7 +65,7 @@ changecom(`%')
 % FIXME: A better comment maybe.
 % See also EMPTY-WORD.x for the highly-unusual situation that EMPTY-WORD
 % appears in the input text.
-EMPTY-WORD.zzz: ZZZ-;
+%EMPTY-WORD.zzz: ZZZ-;
 
 % Capitalization handling (null effect for now- behave as empty words).
 1stCAP.zzz: ZZZ-;
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -588,7 +588,7 @@ static void select_linkages(Sentence sent, fast_matcher_t* mchxt,
 }
 
 /* Partial, but not full initialization of the linakge struct ... */
-void partial_init_linkage(Linkage lkg, unsigned int N_words)
+void partial_init_linkage(Sentence sent, Linkage lkg, unsigned int N_words)
 {
 	lkg->num_links = 0;
 	lkg->lasz = 2 * N_words;
@@ -606,6 +606,7 @@ void partial_init_linkage(Linkage lkg, unsigned int N_words)
 #endif
 
 	lkg->pp_info = NULL;
+	lkg->sent = sent;
 }
 
 void check_link_size(Linkage lkg)
@@ -631,7 +632,7 @@ static void compute_chosen_disjuncts(Sentence sent)
 
 		if (lifo->discarded || lifo->N_violations) continue;
 
-		partial_init_linkage(lkg, pi->N_words);
+		partial_init_linkage(sent, lkg, pi->N_words);
 		extract_links(lkg, pi);
 		compute_link_names(lkg, sent->string_set);
 		/* Because the empty words are used only in the parsing stage, they are
--- a/link-grammar/count.c
+++ b/link-grammar/count.c
@@ -181,6 +181,19 @@ static bool pseudocount(count_context_t * ctxt,
 	return true;
 }
 
+/**
+ * Return the number of optional words strictly between w1 and w2.
+ */
+static int num_optional_words(count_context_t *ctxt, int w1, int w2)
+{
+	int n = 0;
+
+	for (int w = w1+1; w < w2; w++)
+		if (ctxt->local_sent[w].optional) n++;
+
+	return n;
+}
+
 static Count_bin do_count(fast_matcher_t *mchxt,
                           count_context_t *ctxt,
                           int lw, int rw,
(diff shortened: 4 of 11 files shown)


linas:
Wow. That is a very simple, elegant solution. I like it. Shame about the sat-solver.

linas:
Anyway, other than saying that this is nice code, is there anything more I should do right now?

ampli:
I'm trying now to fix the SAT solver parser to support "optional" words.

It seems it needs some small fixes in a few places, since it currently assumes all the words have linkages. However, it is not too easy, as it was not designed for partial linkages, and  also there are no detailed comments (but I already understood most of it).

It is also not clear to me why it works directly on expressions - this prevents it to get the speedup that results by omitting connectors by the "old" power_prune(). Its own `power_prune()` doesn't add real restrictions - the results are the same even when it is commented out, and the only difference is that the program runs **faster** without it... (maybe it was needed to handle FAT links).

One place of bad breakage when "optional" words have no links is in the **connectivity** functions (I finally understood how connectivity is enforced, and I'll need to correct a "why" comment I previously wrote). This is because they now cannot blindly enforce full connectivity, and instead need to account for the case of "optional" words that are allowed to have no links. I fixed maybe most of that (but I'm not sure yet).

I also fixed a  "bug" that added a dependency on the existence of words when it should not.
(It is not a real bug when all words should have linkages, just a redundancy.) Fixing this eliminated bogus linkages due to "missing" words.

After these 3 main fixes, it still not giving some linkages, so debugging it is in progress.
I'm not sure at that time that I will be able to fix it. It may also be a good idea to rewrite it anyway (working on the disjuncts after power-pruning, just like the chart-parser, and using the current fast-matcher).

If it is not fixed, I am not sure what to do with the empty-word-elimination fix.
**A possible solution in that case:**
Eliminate the empty-word as proposed, and move the empty-word adding code to the SAT parser so it will be able to continue to work like now.

ampli:
I fixed the SAT parser too.  For now - only for optional words. The fix is small, and its current empty-word handling code can just be removed.

BTW, it turned out the `generate_epsilon_definitions()` call in power_prune() is needed to reduce the number of duplicate linkages. The rest (that is still not commented out) seems to be damaging when using the fixed (with optional words) version (accounting in it the number of optional words didn't help).

So now it is possible to eliminate the empty-word from both parsers.
I can send a PR.

However, as I wrote above, inserting ZZZ+ is still needed for quotes.
It may be a good idea to fix that too, since this is a dummy linkage anyway (as opposed to QUd/QUc).
(Currently, an opening quote with ZZZ is attached to the previous word.)
**My proposal regarding quotes:**
Eliminate the ZZZ connector. Leave the unlinked quotes in the sentence without changing the null-count (by  making them an optional word too but not discarding them).
Example sentence: He "said": "the sky is blue"
Currently:
```
                     +-----------------QUc----------------+
                     +----------->WV----------->+         |
                     +---------Wd--------+      |         |
    +------>WV------>+----QUd---+        |      |         |
    |      +----Ss---+---Xe--+  |        |      |         |
    +--Wd--+ZZZ+     +-ZZZ+  |  |  +-Dmu-+--Ss--+--Pa-+   |
    |      |   |     |    |  |  |  |     |      |     |   |
LEFT-WALL he   " said.q-d " :.j " the sky.n-u is.v blue.a "
```
Proposed:
```
                     +-----------------QUc----------------+
                     +----------->WV----------->+         |
                     +---------Wd--------+      |         |
    +------>WV------>+----QUd---+        |      |         |
    |      +----Ss---+---Xe--+  |        |      |         |
    +--Wd--+         +       |  |  +-Dmu-+--Ss--+--Pa-+   |
    |      |         |       |  |  |     |      |     |   |
LEFT-WALL he   " said.q-d " :.j " the sky.n-u is.v blue.a "
```
This fix is simple, and that way all the ZZZ-related code can be eliminated altogether.

linas:
OK, Cool! The QU links are supposed to handle "grammatcally correct" usages of quotes, whereas the ZZZ links were handling "emphasis" quotes, which "probably" don't have any "syntacticic" structure, although I guess they have "semantic" content. So, yes, dropping ZZZ is OK.

Please push remaining commits, let me know when you're done.

p.s. I am restarting work on language learning.

ampli:
Here are the steps in my plan to eliminate ZZZ for quotes (and thus eliminate it altogether):

1) Rename the `optional` member of `Word_struct`  to `allow_null_link".
2) Add an additional `bool` member to `Word_struct`: `deletable` (maybe there is a better name).
3) For optional words, both of these members will be set to `true`.
4) remove_empty_word() will remove words where `deletable==true`.
5) Quotes (as defined by `QUOTES` affix-class) will be marked with `allow_null_link=true` (only).

ampli:
I think it's better to send the empty-word-elimination PR as a new one instead of overwriting this one and then close this one (or maybe it is better to overwrite it???).

**Summary of next PR's I would like to send separately:**
1) The said empty-word-elimination PR (regular parser).
2) SAT-parser adaptation for optional words.
3)  ZZZ-elimination.
4) Setting LL links to length 1 (to prevent bogus linkages with null-count>0, see example above).
5) Tokenization fixes for downcasing (a simplification).
6) The "linkage get word position" PR (depends on the downcasing tokenization fixes).
7) Spell bug fixes.
8) SAT-parser long sentence speed up (old branch - needs integration).
9) Extending the SAT parser to handle null-words (depends on (8)).

I don't want to send all at once, due to the dependencies between them - any needed fix in one of them may cause bad merge conflicts in that will make it necessary to resend them. Also, common commits then tend to appear as duplicates (harder to follow).

So I'm waiting for the existing "lasterror" PR to be applied first.

ampli:
Replaced by #477.

PR #470, opened 2017-01-09
Fix printing widths for Chinese ideographs.

linas:
The link-diagram was being printed incorrectly, when Chinese ideographs were involved.
This fixes the issue.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,5 @@
-Version 5.3.14 (?? 2016 ??)
- * Some sort of new work will be done.
+Version 5.3.14 (?? 2017)
+ * Fix printing widths for Chinese ideographs.
 
 Version 5.3.13 (19 Nov 2016)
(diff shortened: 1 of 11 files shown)


ampli:
CC       word-utils.lo
../../../../usr/local/src/link-grammar-devel/link-grammar/link-grammar/wcwidth.c:122:5: warning: no previous prototype for ‘mk_wcwidth’ [-Wmissing-prototypes]
 int mk_wcwidth(wchar_t ucs)
     ^~~~~~~~~~
../../../../usr/local/src/link-grammar-devel/link-grammar/link-grammar/wcwidth.c:207:5: warning: no previous prototype for ‘mk_wcswidth’ [-Wmissing-prototypes]
 int mk_wcswidth(const wchar_t *pwcs, size_t n)
     ^~~~~~~~~~~

PR #472, opened 2017-01-12
Add utf8 support to the random morpheme splitter.

linas:
Add utf8 support to the random morpheme splitter. 
Also, create a second language, "ady", which only splits into two morphemes: stem+suffix.
The "amy" language will be reserved for splits into 3 or more morphemes. (which is 
currently broken, anyway).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,8 @@
 Version 5.3.14 (?? 2017)
  * Fix printing widths for Chinese ideographs.
  * Fix broken randomization in the "any" language.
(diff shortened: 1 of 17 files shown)


ampli:
$ make
...
Making all in ady
gmake[2]: *** No rule to make target '4.0.regex', needed by 'all-am'.  Stop.
Makefile:414: recipe for target 'all-recursive' failed
gmake[1]: *** [all-recursive] Error 1
Makefile:531: recipe for target 'all-recursive' failed
gmake: *** [all-recursive] Error 1

Issue #414, opened 2016-09-21
Improve error reporting


ampli:
Hi Lians,
The discussion started at PR #400. Here is the relevant section:
## 

In order that the library will be able to get embedded seamlessly in applications (like grammar checkers or GUIs) it should not print such error messages, but instead only provide an error indication, and put the error in a variable (a per-thread variable if multi-threaded), to be retrieved by a new function link-grammar_error(). Such an arrangement is found in other libraries.
## 

I thought of two different ways to implement such a scheme:
1. Putting the error messages in a **thread-local variable**. A function like get_linkgrammar_lasterror() will be called after each library API call and will retrieve the messages from this variable.
2. Putting the error messages in the "object". This means adding a "message" member to the Dictionary, Sentence, Linkage and ParseOptions structures.
There will be several function to retrieve the errors, one per object (or one function can be used if the "messages" member is always the first one in the corresponding structure).

But option (2) has a problem: If the dictionary couldn't be opened, there is no object in which to store the error due to which it couldn't be opened (there can be many).  My proposed solution for that is **use NULL as the object** when the returned dict is NULL (and in that case retrieve  the error from a global variable).

Another thing to note is that most of the API functions, including each of the parse_option_get_*() functions, will have to **start with a call to reset_lasterror()**, so errors will not get accumulated over API calls - there can be several error messages from the same API call, so they are allowed to be accumulated per API call, and to be returned at once by get_linkgrammar_lasterror().
# 

I already implemented most of (1), but then thought again on (2), so **I need you guidance** regarding what to actually implement.

Since I intend to initially make it completely compatible with the current way of error reporting (to the extent that no change will be needed in existing applications) there is no need to do things totally the "right" way from the start. We can just declare the new error reporting way as "**experimental**" for several releases until it get stabilized.

linas:
option 1 seems OK. I understand the appeal of option 2 but it seems to add complexity, with no real benefit - our system is just not that error-complex.

I dislike the idea of an automatic reset_lasterror(), except maybe in only one case only: sentence_create(). The idea is that the user could make a whole bunch of API calls, and then check if there was an error, at the end.  That error code would only get reset when starting with a new sentence.

ampli:
After trying to implement both,  option (2) is indeed much more complicated, and has no added functionality.

Regarding resetting messages:
There are currently plenty of  notification messages on things which are not considered as errors.
I mainly thought of discarding them if not gathered "in time".
[...]

PR #473, opened 2017-01-14
Improved error facility

ampli:
Per issue #414.

**N.B.**
The branches for eliminating the empty word from the regular and SAT parsers are also ready,
and I'm waiting for merging this PR in order that I will be able to rebase on the result (may be tedious because of common touched areas). After sending them I will prepare the ZZZ elimination PR.

I propose to issue a new version after this PR (5.3.14). I already included this version number in `link-grammar/README` (please change if needed). Before that, I can fix any problem you find in your code review.


linas:
sorry for delay

PR #477, opened 2017-01-22
Empty word elimination (replacement PR)

ampli:
Replaces PR #469.
The complete ZZZ elimination is to be done in a separate PR.


linas:
same comment as the other pull req: is this ready to merge?

ampli:
> same comment as the other pull req: is this ready to merge?

Indeed it is ready.

It should apply before PR #478, even though theoretically applying #478 could be enough since it is based on #477 and I was supposed to keep the common commits identical.

PR #478, opened 2017-01-22
SAT optional word

ampli:
This is a continuation to PR #477 (that replaces #469).

The batch run times for English are now somewhat slower (~8%), most probably due to added constraints, but maybe also due to more overhead in generate_disconnectivity_prohibited() for handling optional words.

These optional words are a an artifact of extra unit splitting that I think is not really needed, and it introduces significant overhead from the time it has been added (at the start of version 5). For example,
the word "As" gets split to "A s", "mall" to "ma l l", "sin" to "s in", etc. (However, this adds interesting tests cases for the parsers...).  BTW, the current English dict cannot handle such unit splits anyway (it never cold) so introducing them was an incomplete work.

The runtime of Russian is not affected.

I ran detailed comparisons on on the languages (in English also the fixes batch), of producing up to 50K linkages.  The large number of linkages is needed for long sentences, because the SAT parser tends (old problem) to produce more duplicate linkages, especially on very long sentences.

The disconnectivity prohibition is subtle, and I don't have a proof that it is completely right (however, I don't have counter examples).

Most of the power-pruning code is disabled in this PR. It seems to do nothing, besides of adding overhead (checked on the current code - before this PR). Now it even malfunctions (most probably because it doesn't account for optional words).

There are still vast speedup opportunities in the SAT parser code (future work, some of it has already been done).

### DIFF
--- a/data/amy/4.0.dict
+++ b/data/amy/4.0.dict
@@ -10,8 +10,6 @@
 % Dictionary version number is 5.3.14 (formatted as V5v3v14+)
 <dictionary-version-number>: V5v3v14+;
 
-EMPTY-WORD.zzz: ZZZ-;
-
 LEFT-WALL: ANY+;
 ANY-WORD: {@ANY-} & {@ANY+};
 UNKNOWN-WORD: {@ANY-} & {@ANY+};
--- a/data/any/4.0.dict
+++ b/data/any/4.0.dict
@@ -16,6 +16,3 @@ ANY-PUNCT: {@ANY-} & {@ANY+};
 UNKNOWN-WORD: {@ANY-} & {@ANY+};
 
 JUNK: {@ANY-} & {@ANY+};
-
-% Needed to get single-quotes to parse: e.g. "'relex' is the best"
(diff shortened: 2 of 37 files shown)


linas:
Wow. OK, so is this ready to be merged? Or should I wait some more?

I did a quick review, without thinking a lot about it, and it looks reasonable, so I can merge this, if you think that it is ready.

ampli:
>Wow. OK, so is this ready to be merged? Or should I wait some more?

It passes my parsing tests (that compare it to the regular parser) **so it is ready to merge**.

It is said to include in it PR #477, but I think it is better to apply #477 first and then this (PR #478)
in case some difference exist somehow due to a later fix.

I.e. please apply PR #477 first, and then this one, unless GitHUB then indicates from some reason this one cannot apply cleanly (in such a case I will fix it).

I intend to make the tuning for getting the lost 8% performance for English (and much more than that) at or after the version that parses with null words (it already working but needs a performance solution for postprocessing).

PR #481, opened 2017-01-24
anysplit: Fix split to more than 2 parts

ampli:
Main fixes:
1. General: Contraction dict validation should not be done in the tokenizer if anysplit is used.
2. Fix a problem in the `MOR-` regexes.
3. Add issuing a prefix if more than 2 tokens.

The number of suffixes is still variable.
This can be fixed, supposing  a definition of how to perform the marking in
case of more than 3 tokens.

---

I noted a bad interaction with the random splitting and the current way sane-morphism is done:
A sentence may have many millions of parses, but only a few ones are displayed (say even 1-3).

This happens due to a combination of several things:
1.   Every word is broken to many alternatives.
2. The parser is currently doing "mix&match" of the alternatives.
3. The density of good" linkages w/o alternatives mix is very low.
4. The linkage array (for the regular parser) is getting fill before sane-morphism is enforced.

Due to that, only very few "sane" linkages remain.

Possible fix:
Make sane-morphism on the fly when filling the linkage array. This will also simplify the program.
Am lower **!limit** can be used then to prevent issuing of the default 1000 linkages.

A similar problem currently happens also in English for some sentences with many linkages
(when a large portion of them is insane due to bogus unit splitting tries).

I can send a PR if this is a good fix.

### DIFF
--- a/data/amy/4.0.regex
+++ b/data/amy/4.0.regex
@@ -13,9 +13,9 @@ ANY-PUNCT:  /^[[:punct:]]+$/
 
 % Multi-part random morphology: match any string as prefix, stem, or
 % suffix.
-MOR-PREF: /^[[:alnum:]_'-]+.=/
-MOR-STEM: /=[[:alnum:]_'-]+.=/
-MOR-SUFF: /[[:alnum:]_'-]+$/
+MOR-PREF: /^[[:alnum:]_'-]+=$/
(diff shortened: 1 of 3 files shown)


linas:
thanks!

linas:
this seems to mostly fix, but .. well, let me open a new issue..

linas:
Issue #482 describes the problem

PR #485, opened 2017-01-27
Rework flow of linkage processing.

linas:
Previous code processed the linkages in batches. The new code does them to completion, in one loop, more or less, except for post_processing.

This was driven by a need to search for more cases for the "amy" language; unfortunately, that still does not work, and I cannot see why.  So, for now, save of partially finished, but working code (no regressions from before)


ampli:
Why not **always** immediately discard bad-morphisme in `process_linkages()`?
This way there is no need of a special handling for any/amy.

linas:
the final version does this; I couldn't quite get it working last night.

PR #498, opened 2017-02-02
Main README.md: Fix truncated names of files/directories

ampli:
N.B. Note that currently most directories are marked with `/*` (certain ones are not marked at all).
I propose to mark all directories **only**with a trailing `/` .
E.g: `bindings/python/` instead of `bindings/python/*' .
If this is fine I will send a PR for that.


linas:
yes, the removal of the trailing * would be fine

PR #499, opened 2017-02-02
Add alternatives check in the fast-matcher

ampli:
This PR also contains (ifdef'ed out) the power-prune alt-pruning code.

BTW, here is a convenient way to produce a list of linkages:
`link-parser amy -m -test=auto-next-linkage < file`
when file contains one or more sentences. Note that you need to add **-batch** in case the file
sets batch mode (to negate it).
OR
`echo 'This is a sentence' | link-parser amy -m -test=auto-next-linkage`

As far as I have checked, this PR removes all the duplicates.

(A similar check can be added in the SAT-parser.)

### DIFF
--- a/link-grammar/disjunct-utils.c
+++ b/link-grammar/disjunct-utils.c
@@ -97,15 +97,15 @@ static inline unsigned int old_hash_disjunct(disjunct_dup_table *dt, Disjunct * 
 }
 
 /* FIXME? Can the same word get appended again? If so - prevent it. */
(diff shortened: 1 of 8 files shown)


linas:
The README and/or other documentation needs a section on "convenient ways to do things".  For example, I am perenially confused about the right way to turn on debugging, and lose a lot of time trying to remember.

ampli:
I will send a PR for that.

Issue #500, opened 2017-02-02
continuing crashes in wordgraph, "ady" language


linas:
I currently cannot reproduce this, but just saw this:

```
*** glibc detected *** java: free(): invalid pointer: 0x0000000059551940
***
======= Backtrace: =========
/lib64/libc.so.6[0x3e296714af]
/lib64/libc.so.6(cfree+0x4b)[0x3e296757ab]
/usr/local/lib/liblink-grammar.so.5[0x2b3a2392a7ce]
/usr/local/lib/liblink-grammar.so.5[0x2b3a2392b396]
/usr/local/lib/liblink-grammar.so.5(sentence_parse+0x99)[0x2b3a2392bb69]
/usr/local/lib/liblink-grammar-java.so.5.3.15(Java_org_linkgrammar_LinkGrammar_parse+0xf5)[0x2b3a236f8995]
[0x2b3a1690eeee]

```
 Examining teh core file:
```
#5  0x00002b3a2392a7ce in wordgraph_path_free (free_final_path=true, wp=0x59551860)
    at ../../link-grammar/api.c:937
```
which is currently:
```
static void wordgraph_path_free(Wordgraph_pathpos *wp, bool free_final_path)
{
   Wordgraph_pathpos *twp;

   if (NULL == wp) return;
   for (twp = wp; NULL != twp->word; twp++)
   {
      if (free_final_path || (MT_INFRASTRUCTURE != twp->word->morpheme_type))
         free(twp->path);        <<<<<<<<<<<<<< line 937
   }
   free(wp);
}
```
```
#6  sane_linkage_morphism (sent=sent@entry=0x586bb130, lkg=lkg@entry=0x5877b850, 
    opts=opts@entry=0x586baef0) at ../../link-grammar/api.c:1021


(gdb) print twp
$1 = (Wordgraph_pathpos *) 0x59551890
(gdb) print *twp
$3 = {word = 0x587d7f40, same_word = 56, next_ok = 74, used = 149, path = 0x59551940}

(gdb) x/16g 0x59551940
0x59551940: 0x00000000587d6ef0   0x0000000000000000
0x59551950: 0x0000000000001270   0x0000000000000021
0x59551960: 0x000000005894f650   0x0000003e29954a68
0x59551970: 0x0000000000000020   0x0000000000000040
0x59551980: 0x00000000595516f0   0x000000005a5c82a0

				free(twp->path);
(gdb) print  free_final_path
$5 = true

[...]

PR #501, opened 2017-02-02
Bugfix for bug #500

linas:
Terminating null entry was not moved.


linas:
OK this doesn't work as written.

linas:
OK, so I fixed a rather confusing off-by-one. It is slightly less efficient than the old code, because it requires two additions instead of one, but .. wow, it made the old code hard to understand.

linas:
After further review, this patch is a no-op, it doesn't actually change any function at all. It was created during the hunt to fix #500

PR #508, opened 2017-02-05
Revert "Bugfix for bug #500"

ampli:
For some reason reverting it caused a conflict (which I solved).
I don't know if it will conflict again when you apply it...

### DIFF
--- a/link-grammar/wordgraph.c
+++ b/link-grammar/wordgraph.c
@@ -197,15 +197,11 @@ bool wordgraph_pathpos_add(Wordgraph_pathpos **wp, Gword *p, bool used,
 		}
 	}
 
-	*wp = wordgraph_pathpos_resize(*wp, n+1);
-
+	*wp = wordgraph_pathpos_resize(*wp, n);
 	if (insert_here < n)
 	{
-		/* n+1 because n is the length of the array, not including the
-		 * terminating null entry. We need to protect the terminating null.
-		 */
 		memmove(&(*wp)[insert_here+1], &(*wp)[insert_here],
-		        (n+1 - insert_here) * sizeof (*wpt));
+		        (n - insert_here) * sizeof (*wpt));
 	}
 
 	(*wp)[insert_here].word = p;


link-grammar/wordgraph.c
```diff
@@ -197,15 +197,11 @@ bool wordgraph_pathpos_add(Wordgraph_pathpos **wp, Gword *p, bool used,
 		}
 	}
 
-	*wp = wordgraph_pathpos_resize(*wp, n+1);
-
+	*wp = wordgraph_pathpos_resize(*wp, n);
```
linas: the memove fix below is correct, but I don't believe that the above is. That's because I also changed the argument to `wordgraph_pathpos_resize` so that it expects the new length, not counting the terminating null. by contrast, the `n` is the old length.  You need to add one to make room for the new element.

ampli:
You are right. I neglected to note you changed it.
But I'm not sure it was a good idea to change it, since now it is inconsistent.
It (and not only it) is modeled after the original resize_alt() from 4.8.6:
```c
static const char ** resize_alts(const char **arr, size_t len)
{
        arr = (const char **)realloc(arr, (len+2) * sizeof(const char *));
        arr[len+1] = NULL;
        return arr;
}
```
These family of resize functions add a new element at index n, not a new array length of n.
Note that wordgraph_pathpos_resize() is used also in another place, that you have not changed.
So I suggest that I change it back and document it.

Please tell me what you prefer.
In any case a new PR is needed.

linas:
Talking about resize_alts right now will confuse the issue, so let me ignore it; it's used in only one localized place, so that's OK.

Now, `wordgraph_pathpos_resize()` is used in two places, and when I grep:
```
api.c:	*nwp = wordgraph_pathpos_resize(*nwp, n+1);
wordgraph.c:	*wp = wordgraph_pathpos_resize(*wp, n+1);
```

so I think that both places are using it correctly, in the new code.  The reason I changed it to add one, instead of adding two, is that I'm used to thinking about null-terminated strings in a certain way: thier length, and the location of the terminating null.  Iwould prefer to be able to think about `wordgraph_pathpos` in the same way.  So, although technically, there was nothing incorrect with adding 2 and then writing a null into length+1, this just ended up being more and more confusing to think about, while I was trying to debug the crash.  It was just one more "non-standard" way of thinking, it confused me, it seemed likely to lead to new bugs ... and so I changed it. 

In short, the original code was technically correct, but psychologically confusing, and an extra cognitive burden. The new code is still technically correct (since I changed both places, right?) and is more string-like in it's behavior. 

Now, `resize_alts` still uses the +2 behavior, but that's OK, since it's use is just in one very limited location, and so no confusion arises.  We can leave that alone.

ampli:
Is there anything more to change here, or should this PR just be closed?

Issue #509, opened 2017-02-06
pp_info memory leak


ampli:
At appeared after at the `process_linkage()` version.
It didn't exist at ae7e8eff2a296ecf8d924d59f843926571137c4a.

The leak (at 1b2f201d530fa836b9c1b30f655d995d9e1b8ab9):
```gdb
Direct leak of 16 byte(s) in 2 object(s) allocated from:
    #0 0x7f6215d5de60 in malloc (/lib64/libasan.so.3+0xc6e60)
    #1 0x7f6215701980 in exalloc ../../../../usr/local/src/link-grammar-devel/link-grammar/link-grammar/utilities.c:441
    #2 0x7f62156b02db in linkage_set_domain_names ../../../../usr/local/src/link-grammar-devel/link-grammar/link-grammar/post-process.c:358
    #3 0x7f621565900f in post_process_lkgs ../../../../usr/local/src/link-grammar-devel/link-grammar/link-grammar/api.c:584
    #4 0x7f6215660786 in classic_parse ../../../../usr/local/src/link-grammar-devel/link-grammar/link-grammar/api.c:1470
    #5 0x7f6215660cc9 in sentence_parse ../../../../usr/local/src/link-grammar-devel/link-grammar/link-grammar/api.c:1529
    #6 0x4070b0 in main ../../../../usr/local/src/link-grammar-devel/link-grammar/link-parser/link-parser.c:786
    #7 0x7f6212d06400 in __libc_start_main (/lib64/libc.so.6+0x20400)
```

Example sentence (ru):
`Пять дней`

It happens only with `!morphology=0` apparently due to shortening of `num_links` from 5 to 3 (by omitting the LL links). However, it just worked well before the `process_linkage()` reorganization, that was not supposed to make any related changes. So I'm confused regarding the possible reason of this problem.

P.S. I got a similar problem due to `!morphology=0` in the sat-no-empty-word version, which I fixed,
but now I see there remained another memleak in on ``!morphology=0` (which I will have to fix).
(As said above, the pp_info memleak in the classical parser didn't exist just after the sat-no-empty-word and appeared only at the recent `process_linkage()` version.)

ampli:
Fixed in #510.
The memleak in SAT parser also happen only in case of `!morphology=0`, but is not due to pp_info.
I'm going to fix it in an case, so I'm closing this issue.

PR #510, opened 2017-02-07
remap_linkages(): Restore pp_info handling

ampli:
Fix issue #509.

Commit 0cf8c60f removed remapping/freeng of pp_info.
This causes a memleak in case of HIDE_MORPHO, because then the pp_info
related to the removed links is not freed at linkage_free_pp_info().

Restore the relevant part of the removed code.

The part that zeros out the rest of the domain name array is not restored
because it seems it is not needed, as anything beyond pp_info[j] is not
going to be accessed any more.

### DIFF
--- a/link-grammar/linkage.c
+++ b/link-grammar/linkage.c
@@ -185,7 +185,17 @@ static void remap_linkages(Linkage lkg, const int *remap)
 
 			new_lnk->link_name = old_lnk->link_name;
 
+			/* Remap the pp_info, too. */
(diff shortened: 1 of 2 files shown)


linas:
github is acting weird again: it shows almost the same diffs as pull req #511 -- since this is earlier, I merged it first, and expected #511 to be a merge conflict .. but it wasn't.  Reviwing the diffs in both pull reqs, and the code looks good, its just weird that github is showing me these .. confusing diffs.

PR #512, opened 2017-02-07
Don't invoke remap_linkages() if no word got removed

ampli:
If remove_empty_words() did nothing, there is no need to invoke remap linkages.
This saves non-negligible CPU on English, with the cost of (maybe) adding negligible overhead on other languages.


linas:
ongoing github weirdness -- this is showing a diff that is very similar to that of #510 and #511 -- I see the once place where the diffs are different, and that does indeed seem to be your patch.  It's just .. weir

PR #513, opened 2017-02-07
Fix a crash in alt_connection_possible()

ampli:
The crash can be repeated by:
tests.py ZENLangTestCase

Problem explanation:
gword_record_in_connector() is called once, in prepare_to_parse().
It puts a reference in each connector to the gword array of its
disjunct.

In case a parse with possible nulls is done in one step (i.e. using
min_null_count=0 and max_null_count>0), some disjuncts may get freed
by power_prune() along with their gword array, leading to a stale
reference in the connectors in case a parse with null_count>0 is needed
(such as in the test that crashes).

The solution that is implemented here is to use the disjuncts_copy that
is done in that case, since it is valid until the end of the parse.

This doesn't look nice. Hence the FIXME.
It can be done better by using a gword linked-list in the disjuncts that
is not getting freed along with the disjuncts but separately in
sentence_delete(). This is also lighter in CPU and memory.

### DIFF
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -1355,6 +1355,24 @@ static void process_linkages(Sentence sent, bool overflowed, Parse_Options opts)
 	}
 }
 
+#ifdef ALT_CONNECTION_POSSIBLE
+/**
+ * Record the wordgraph word in each of its connectors.
+ * It is used for checking alternatives consistency.
+ */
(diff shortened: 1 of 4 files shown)


ampli:
Replaced by #514.

linas:
closing, because I merged #514, and so that makes this pull req obsolete, right?

ampli:
Yes,  I actually forgot to close it.

PR #516, opened 2017-02-14
Configure

ampli:


ampli:
(Opened with empty comment by pressing an extra ENTER by mistake...)
Currently the configure script doesn't use "best practice" for CFLAGS etc.
For example, it sets -O3 by default, and doesn't allow to change it to -O2 by CFLAGS environment or argument configure (because both are prepended). This has not been changed here, because we have to decide about the desired setup of the optimization flag (the autoconf default is -O2).

This PR allows to append a CFLAGS/CPPFLAGS when using --allow-debug and it also append these variables to other compiler flag setup, effectively allowing to override them.

See the commit message for an example of a useful debug flag that can be set now by CFLAGS/CPPFLAGS when configuring.

PR #525, opened 2017-02-23
Account for optional words when islands_ok=true

ampli:
This is a bug fix.
In b10861b78e0d57ff017babea955be1599fbfe43f I neglected to support the case of `ilands_ok=1`.
This PR supports it too.
The original comments above the code changes in these both PR's are now not accurate and need changing.

---

BTW, I encountered  a problem in this idea of eliminating the empty word (may happen in certain cases when parsing with null_count>0), for which I will open an issue soon so we will be able to discuss it.


linas:
thanks... sorry for delay.

Issue #532, opened 2017-04-28
Syntax error


liliburnett:
From ./autogen.sh:

../configure: line 17259: syntax error near unexpected token `,:'
../configure: line 17259: `			AX_TLS(,:)'

from autogen.err
`configure.ac:327: error: possibly undefined macro: AC_MSG_NOTICE
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
`

I have pkg-config, this is Arch Linux.

ampli:
It looks like you need to install `autoconf-archive`.

linas:
assuming that installing autconf-archive fixed this.

But really, you should NOT work out of git, you should use the released versions; only those ar guaranteeed to be bug free, and actually work. The stuff in git might not compile.

PR #551, opened 2017-06-06
Compilation fixes

ampli:
- Fix issue #532.
- Gitignore additional files.
- Fix compilation in an external directory (finding pp_lexer.h).
- Fix SAT-parser compilation (include file paths).
- Fix setting wordgraph flags (avoid the possibility of out-of-range shifts, especially negative ones).


linas:
Thanks!

PR #553, opened 2017-06-08
Updates for Windows

ampli:
1. Adapt to the new source structure/files.
2. Add optional Flex support (defined by default).
3. Update the README about Flex usage and fix the broken table for Definitions for Python bindings.
4. Fix compilation breakage due to PR #535.
5. Add utilities.h in constituents.c to get the strtok_r redefinition if needed.
6. Fix 2 improper definitions that got detected by MSVC.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -57,7 +57,7 @@ AC_LANG([C])
 
 AC_C_CONST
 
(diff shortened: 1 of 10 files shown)


linas:
thanks you for the attention to detail!

PR #556, opened 2017-06-10
Convert dict_display_word_info()/expr() to return (char *)

ampli:
... instead of directly printing the results.

This solves the following problems:
1. It is not a good idea to print from the library code (it is not flexible, and is not appropriate for a GUI).
2. It is not consistent with other LG function API.
3. If a dictionary entry is not found, link-parser prints 2 errors.

This is an API change. However, this seems to me fine since these functions are not externally documented and n addition link-includes.h says they are internal functions.

**NOTES:**
1. This pull request depend on the previous one.
2. SUBSCRIPT_MARK to SUBSCRIPT_DOT conversion is omitted here! This will be handled in a general automatic way in the next pull request I'm about to send.


linas:
I reviewed this quickly, it looks reasonable.

PR #557, opened 2017-06-10
Convert append_string() of fixed strings to dyn_strcat()

ampli:


ampli:
Should we now rename append_string() to something that start with **dyn_str**?
For example: **dyn_str_printf()**

PR #558, opened 2017-06-10
"Automatic" conversion of SUBSCRIPT_MARK to SUBSCRIPT_DOT

ampli:
**EDIT: THIS PR SLIGHTLY CONFLICTS...***
If you apply  anyway, it is easy to fix the 2 created conflicts.
But you can apply only the previous ones, and I will  pull and fix the conflicts and then push this PR again in-place (with a fix of a warning on an unused variable).

---
PR #556 doesn't include this needed conversion. Instead, it depends on the change in this PR.

After this PR, the said conversion can be removed from another place (I will send an additional PR for that when all the recent PRs are applied).

**NOTE:**
(Edit) In order to prevent conflicts, the previous pending PRs should be applied in their order.

### DIFF
--- a/link-grammar/linkage/linkage.c
+++ b/link-grammar/linkage/linkage.c
@@ -224,11 +224,7 @@ static void print_chosen_disjuncts_words(const Linkage lkg)
 		else
 			djw = cdj->string;
 
-		char *djw_tmp = strdupa(djw);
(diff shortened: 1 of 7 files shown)


ampli:
Merged into #561.

Issue #420, opened 2016-09-26
Get words in original sentence


linas:
See pull req #416 for discussion.
To summarize:   [stackoverflow question](http://stackoverflow.com/questions/38545726/how-to-use-the-link-grammar-parser-as-a-grammar-checker) - how to use LG as a grammar checker - asks for an API to find which words where not included in a parse. The issue is non-trivial due to the following: 
* word-numbering is tokenization-dependent, e.g. if a word is split  it's -> it 's
* spell-guesser can change length of words
* utf8 langs have word begin/end counts differ from byte counts.

ampli:
>I guess you mean "that token in the original sentence that I would get, if I split the original sentence by using blank spaces as word-separators"

Approximately but not exactly - see below.

>For the case of "it's" being split into two words, the user could write a utility themselves, its not >hard. For spell-guessing, it gets hard, because string compares are no longer possible.

It may get complicated - there are also punctuation split, morpheme splits and unit splits.

>Yes, I guess something like this might work:
>
>char * linkage_get_orig_word(const Linkage linkage, WordIdx word_num);
>and maybe also

>int linkage_get_orig_index(const Linkage linkage, WordIdx word_num);

It is more complicated than that. Take for example the following sentence:
```
Linkparser> This is a, test
...
LEFT-WALL this.p is.v a [,] test.n
```

For "`[,]`", you don't want to return "`a,`".
There are similar and also more complicated other examples.

To remove any ambiguity, it seems there is maybe a need to return (start,end) string positions. This will also enable the application to mark (underline, color, etc.) the offending tokens, and more.

Only after trying to implement it, including writing a demo using it, we can see what is actually needed. I can try to do that.

>Can you think of a better name for "orig_word" and "orig_index"? There must be some nicer way of saying this...

Maybe use the word **sentence** instead of **orig**.

linas:
yes, start,end string positions are exactly the right thing!  then `linkage_get_word_start()` and `linkage_get_word_end()` would work.

ampli:
I will implement it and add it to the sentence-check.py example program.

ampli:
Hi Linas,
I implemented it. However, I encountered a problem.
I need you comments on these things before I finish the code polishing (including test cases) and send a PR.

Currently, my implementation returns start and end offsets **in bytes**.
When I wanted to use that in a demo program, it looked fine for a simple English sentences.
But when I tried Rusian I realized that I really would like **character offsets**, not byte offsets.
[...]

Issue #404, opened 2016-09-17
spell-guessing mis-handles capitalized words


linas:
Even with spell-guessing enabled, the sentence " Therefo, I declare I love Computer Science" is not fixed.  However, the lower-case version of this works fine.  I suggest that spell-guessing should down-case words before trying to correct them.  (for aspell, I did not attempt Hunspell)

ampli:
It even doesn't arrive the spell-guess step.

By design (per the original one), the lookup order of the tokenizer is as follows, and it **stops** on the first successful one (from my memory):
1. Dictionary lookup of the whole word, as is and also downcased.
2. Morpheme split (as is and also downcased).
3. Regex lookup (capital words also match).
4. Spell guess.

It means that a word that matches a regex doesn't pass a spell correction.

Similar problems happen with these sentences:

**therfor, I declare ...**
therfor[!C-NOUN-WORDS].n , I.p declare.v ....y

elefants are big
elefants[!S-WORDS].n are.v big.a

From tokenizer.c:
```
	 * ??? Should we add spell guesses as alternatives in case:
	 * 1. The word if not in the main dict but matches a regex.
	 * 2. The word an unknown capitalized word.
```
If the answer to the above is "yes", I can implement that.

Also note that words that match a regex loss the opportunity to be handled by the **unknown-word** device!  For handling such things we may need an additional kind of post-processing, plus maybe programmatically costs.

linas:
Hmm. 

If a word is capitalized and unknown, we should spell-guess the lower-case version.
If a word ends with an s and is unknown,  we should spell-guess it.

In other cases, we should not do so?

The point is, I don't want to run roman numerals through the spell-guesser, since I already think I know what they are.

Maybe in some ideal world, we should allow the dictionary to contain something like
```
foo: BAr+ or <spell-guess>;
```
and even assign costs:
```
foo: BAr+ or [<spell-guess>]3.14159;
```
so that the BAr+ connector gets priority, else the spell-cheeked version.  .. Something like that.  This would allow common typos to be automaitcally handled: e.g.
```
there: ... or [[<spell-guess>]];
```
which would add their, they're as possibilities, but at a large cost.

ampli:
> If a word is capitalized and unknown, we should spell-guess the lower-case version.
> If a word ends with an s and is unknown, we should spell-guess it.
> 
> In other cases, we should not do so?

[...]

PR #564, opened 2017-07-11
Implement linkage_get_word_*() (issue #420)

ampli:
We can leave linkage_get_word() as is (subscripted words) for compatibility, since these functions can be used to retrieve the bare words.

For the linkage words we can add functions to separately retrieve the bare word, the subscript, and the guess information. If this is a good idea I can open an issue for that.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,7 @@ Version 5.4.0 (XXX 2017)
  * Major reorganization of code base into more modular directories.
  * Revive the sqlite3 dictionary into operational form.
  * Add double-quotes to splittable punctuation for the "any" langauge.
+ * Add API functions to get linkage word positions in the sentence.
 
 Version 5.3.16 (15 April 2017)
  * Fix python3 unit tests.
--- a/bindings/python-examples/sentence-check.py
+++ b/bindings/python-examples/sentence-check.py
@@ -34,19 +34,31 @@ def nsuffix(q):
 
 DISPLAY_GUESSES = True   # Display regex and POS guesses
 DEBUG_POSITION = True    # Debug word position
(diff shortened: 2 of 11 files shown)


linas:
thank you!

linas:
My gut impression is that information about subscripts and guessing is inappropriate for users to look at and make use of.  I used to believe that the subscripts could be used in "meaningful" ways, but eventually came to realize that they are bogus.  it's really the case that the disjuncts carry the important information, and the other stuff (subscripts, guess marks) are "algorithm noise", best ignored.

linas:
I'm vaguely thinking of having a version 5.4.0 possibly soon. Anything you want to do before then?

ampli:
> My gut impression is that information about subscripts and guessing is inappropriate for users to look at and make use of. I used to believe that the subscripts could be used in "meaningful" ways, but eventually came to realize that they are bogus. it's really the case that the disjuncts carry the important information, and the other stuff (subscripts, guess marks) are "algorithm noise", best ignored.

Maybe guesses can have some metric, different than the current ones, which is sorted last.
This may solve some of the problems such as in #404. It will also hint that the linkage is actually not of the exact original sentence.

ampli:
> I'm vaguely thinking of having a version 5.4.0 possibly soon. Anything you want to do before then?

Maybe I can send PR for a python binding for my recently introduced lg_error_flush(), that I forgot to implement.
(I have many other fixes and enhancements, but they need porting to the current code, and also some more developement and checks, so this will be for the next versions.)

ampli:
I have just sent PR #565 for completing lg_error_flush().
After you merge it and before you issue version 5.4.0, I would like to send a general cleanup PR.

PR #566, opened 2017-07-14
Cleanup

ampli:
1. Fix typos.
2. Fix include file list in Windows.


linas:
thanks

ampli:
Is expand.h still needed?

linas:
No. It seems like a good idea, somehow, but in it's naive form, it never worked well. Its obsolete.

PR #568, opened 2017-07-18
Bare word getters

glicerico:
Added function
'''
char * linkage_get_bare_word(const Linkage linkage, WordIdx w)
'''
to return the w-th word of a parsed sentence without the tags "[?]", "[!]" for unknow words or the word tag specifying the part of speech it belongs to (e.g. withough ".n" in "dog.n").
Note: one needs to be careful to free the memory allocated in the function by freeing the pointer where its return value is stored.

### DIFF
--- a/link-grammar/linkage/linkage.c
+++ b/link-grammar/linkage/linkage.c
@@ -875,6 +875,29 @@ const char * linkage_get_word(const Linkage linkage, WordIdx w)
 	return linkage->word[w];
 }
 
+// Need to free the returned value of this function later:
+// the memory pointed to by bare_word
+const char * linkage_get_bare_word(const Linkage linkage, WordIdx w)
+{
+	if (!linkage) return NULL;
+	if (linkage->num_words <= w) return NULL; /* bounds-check */
+
+	const char delimiter = '.';
+	char *separator_ptr = strchr(linkage->word[w], delimiter);
+
+	if(separator_ptr == NULL)
+	{
+	  return linkage->word[w]; // if there is no separator in the word
+	}
+	// copy and return word without the suffix
+	int position = separator_ptr - linkage->word[w];
+	char* bare_word = (char*) malloc((position + 1) * sizeof(char));
+	memcpy(bare_word, linkage->word[w], position);
+	bare_word[position] = '\0';
+	return bare_word;
+}
+
+
 int linkage_unused_word_cost(const Linkage linkage)
 {
 	/* The sat solver (currently) fails to fill in info */
--- a/link-grammar/linkage/linkage.c
+++ b/link-grammar/linkage/linkage.c
@@ -877,17 +877,26 @@ const char * linkage_get_word(const Linkage linkage, WordIdx w)
 
 // Need to free the returned value of this function later:
 // the memory pointed to by bare_word
-const char * linkage_get_bare_word(const Linkage linkage, WordIdx w)
+char * linkage_get_bare_word(const Linkage linkage, WordIdx w)
 {
 	if (!linkage) return NULL;
 	if (linkage->num_words <= w) return NULL; /* bounds-check */
 
-	const char delimiter = '.';
-	char *separator_ptr = strchr(linkage->word[w], delimiter);
+	// first look for unknown word tag
+	char delimiter = '[';
+	char *separator_ptr = strrchr(linkage->word[w], delimiter);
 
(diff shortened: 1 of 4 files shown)


link-grammar/linkage/linkage.c
```diff
@@ -875,6 +875,38 @@ const char * linkage_get_word(const Linkage linkage, WordIdx w)
...
 
+// Need to free the returned value of this function later:
+// the memory pointed to by bare_word
+char * linkage_get_bare_word(const Linkage linkage, WordIdx w)
+{
+	if (!linkage) return NULL;
+	if (linkage->num_words <= w) return NULL; /* bounds-check */
+
+	// first look for unknown word tag
+	char delimiter = '[';
```
linas: the square bracket is also used for spell guesing and regexes, this will not work.

link-grammar/linkage/linkage.c
```diff
@@ -875,6 +875,38 @@ const char * linkage_get_word(const Linkage linkage, WordIdx w)
...
+	char *separator_ptr = strrchr(linkage->word[w], delimiter);
+
+	if(separator_ptr == NULL)
+	{
+	    // look for beginning of word tag
+		delimiter = '.';
```
linas: this can't work, because abbreviations have periods in them: e.g. Jr. Sr. Corp. Inc. Dr. Mr.  that is why we used 0x3 as the separator.

link-grammar/linkage/linkage.c
```diff
@@ -875,6 +875,38 @@ const char * linkage_get_word(const Linkage linkage, WordIdx w)
...
+	    separator_ptr = strrchr(linkage->word[w], delimiter);
+	    if(separator_ptr == NULL)
+		{
```
linas: This is incorrectly indented.

link-grammar/linkage/linkage.c
```diff
@@ -875,6 +875,38 @@ const char * linkage_get_word(const Linkage linkage, WordIdx w)
...
+			// if no tags are aded, point to the end of the word
+			delimiter = '\0';
+	  		separator_ptr = strrchr(linkage->word[w], delimiter); // if there is no separator in the word
+		}
+	}
+	// copy and return word without the suffix
+	int position = separator_ptr - linkage->word[w];
+	char* bare_word = (char*) malloc((position + 1) * sizeof(char));
+	memcpy(bare_word, linkage->word[w], position);
```
linas: this causes a memory leak.

glicerico: Hi @linas, I didn't notice Amir's update. Last time I read was his comment in google groups that his changes were still for an older version of link-grammar, so I decided to give it a try myself. My bad for not checking further commits.
Thanks for pointing out my mistakes, clearly my function is not addressing all possibilities. I will cancel my pull request.
About the memory leak part, I see that Amir uses alloca, which I didn't know before. So, in general will you have a memory leak if you malloc memory in a function even if you are careful to free it in the calling function? Valgrind didn't detect a memory leak the way I used my function.

glicerico: I just realized that 
    linkage->wg_path_display[word_index]->subword
contains the indexed bare word, without extra tags, so that's probably a way to get them directly, unless I'm missing some possible suffix I have not found so far.

linas: alloca allocates on stack. When the stack frame disappears, the pointer is no longer valid.  

malloc is dangerous, because users have a habit of forgetting to free. the LG API tries hard to avoid returning malloced memory.

LG subwords are used to hold morphemes and spell-guessed word-splits. They probably don't do what you think they do, unless you carefully studied the Russian link-grammar, where they are used heavily.  The simplest English example would be "seasond" which might be one word: "seasoned" or it might be two: "seas and"  "sea sand".

The functions that Amir added provide both byte and character offsets. In English, these are the same thing, for non-English, they are not.

linas:
I will publish a version 5.4.0 in a few days or a week, it will have the new word API and many other fixes.   Please verify that the API that Amir provided is sufficient for your needs before then ... I'd rather fix tuff before, not after the new version comes out.

glicerico:
Hi Linas, I confirm that the word API added by Amir is good for my
purposes. Thanks to both of you!
a.

linas:
FYI, it was #420 that described the original issue and  #564 and #565 that added the code.

PR #574, opened 2017-07-29
Fix man page processing and installation breakage

ampli:
(It got broken in 5.3.8.)

- Fix sed command to use "," separator, needed since datadir usually contains slashes (broken in commit 7b5d29ef "Remove obsolete and unused BINRELOC support").

- Fix man page filename (broken in commit 6a516c7a).

P.S. Using autoconf for the replacement doesn't seem straightforward because $(datadir) remains unexpanded.


linas:
thanks!

PR #575, opened 2017-08-06
Add MPUNC

ampli:
1. Add the ability to split words at MPUNC punctuation list.
2. Add/adjust the random splitters definitions to  split hyphenated words and words with underbars, and also words with em-dash (`--` and `—`).
3. Add MPUNC to `en/4.affix`. Currently it contains only ASCII em-dash (`--`). Add also an example sentence (from the old defect list).
Most probably adding UNICODE dashes to the English dict is a good idea (at least em-dash) since LPUNC/RPUNC already includes them they are currently unknown-words.
4. Implement "negative-regex" for REGPRE/REGMID/REGSUF (not actually used for now).
5. Fix some bugs/problems as described in the commit messages, especially ones that interfere with these changes.


linas:
will look at this "really soon now"

linas:
thanks!

linas:
No comments, I looked at it and it seemed reasonable.

PR #580, opened 2017-09-24
Fix a crash when there is no PP info

ampli:
Return an error indication in the following API calls when there is no PP info:
linkage_get_link_num_domains()
linkage_get_link_domain_names()

Repeat by:
$ link-parser any -links
...
linkparser> test
...
Segmentation fault (core dumped)

(It also happens with he, lt, etc.)


linas:
thanks!

Issue #582, opened 2017-09-30
MAX_LINE is too low


ampli:
In a recent commit  0af2d1d281f521074f0939233e273e3e9cad9f16  (from May 18), MAX_LINE got changed from 2500 to 500. As a result, English sentences with more than ~60 words cannot be displayed any more. With Russian this happens with much less words.

I reviewed the sources and it seems it is easy to dynamically allocate MAX_LINE as needed.
I will send a PR for that.

linas:
It was this commit ... https://github.com/opencog/link-grammar/commit/23e567a3ce94cbfc61c562f67465479303d18fe1

yes, OK, make it dynamic

ampli:
Just now I sent PR #584, that removes both MAX_LINE and MAX_HEIGHT.

PR #584, opened 2017-10-01
 Diagram printing: Eliminate size hard-wired limits

ampli:
Fix issue #582.

Changes:
1) Diagram printing: Eliminate the MAX_HEIGHT and MAX_LINE limits.
2) utf8_strwidth(): Eliminate MAX_LINE limit.
3) print_one_disjunct(): Remove the MAX_LINE limitation. This change was not really needed....
4) Fix a typo.
5) Remove unneeded includes from the touched files. I guess that there is much more "include file rot" elsewhere.
6) Change vappend_string()/append_string() to return the number of appended chars, to be more similar to the *printf functions. Not used for now. In the same occassion I added I added function doc comments.
7) The stack buffer overflow is from the previous PR, and included here so I can update the ChangeLog without creating a conflict.

Also, a possible crash in linkage_print_diagram_ctxt(), due too a double-free when the diagram is too high, gets "automatically" fixed by this change. The reason of the bug was this code:
``` c
			if (2*row+2 > MAX_HEIGHT-1) {
				dyn_strcat(string, "The diagram is too high.\n");
				dyn_str_take(string);
			}
```

It should have been `return dyn_str_take(string)`.
If MAX_HEIGHT is reached, a C library crash results:
`*** Error in `/tmp/link-grammar/master/link-parser/.libs/lt-link-parser': double free or corruption (fasttop): 0x00000000024ec4e0 ***`

### DIFF
--- a/link-grammar/print/print-util.c
+++ b/link-grammar/print/print-util.c
@@ -63,8 +63,14 @@ size_t utf8_strwidth(const char *s)
 
 /* ============================================================= */
 
-/* Note: As in the rest of the LG library, we assume here C99 compliance. */
-void vappend_string(dyn_str * string, const char *fmt, va_list args)
+/**
+ * Append to a dynamic string with vprintf-like formatting.
+ * @return The number of appended characters.
+ *
+ * Note: As in the rest of the LG library, we assume here C99 library
+ * compliance (without it, this code would be buggy).
+ */
+int vappend_string(dyn_str * string, const char *fmt, va_list args)
 {
 #define TMPLEN 1024 /* Big enough for a possible error message, see below */
 	char temp_buffer[TMPLEN];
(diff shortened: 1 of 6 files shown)


link-grammar/print/print-util.c
```diff
@@ -50,8 +63,14 @@ size_t utf8_strwidth(const char *s)
 
 /* ============================================================= */
 
-/* Note: As in the rest of the LG library, we assume here C99 compliance. */
-void vappend_string(dyn_str * string, const char *fmt, va_list args)
+/**
+ * Append to a dynamic string with vprintf-like formatting.
+ * @return The number of appended characters.
```
linas: Number of appended bytes, or number of appended glyphs?

ampli: It is the number of bytes.
However, the vsnprintf() Linux manual use the term "characters" for that.
The BSD manual also uses the term "characters".

---
RETURN VALUE
Upon successful return, these functions return the number of characters printed (excluding the null byte used to end  output to strings).

---
However, I neglected to mention that a negative value is returned in case of error.
So I will add that and change it to bytes too.

ampli:
> So I will add that and change it to bytes too.

Addressed in #587.

PR #594, opened 2017-10-20
Fix man page installation

ampli:
Hopefully this fixes the man page installation that got broken long ago.


linas:
Hah. Well, I suppose no one noticed :-(

PR #601, opened 2017-10-29
Windows fixes

ampli:
I wanted to test the long-word split code on Windows, but encountered a compilation error because the `ssize_t` definitions that I recently added were not recognized (it is `SSIZE_T` on MSVC).
I fixed that and in the same occasion also several new (bit rot) compilation warnings.
Also included some other minor fixes for Windows.
An interesting fix for MSVC relates to stderr full buffering when the output is not a console. I noted that when running an MSVC binary under Cygwin X11. This may also cause (again on MSVC) unsync stdout/stderr when the output is a file.

### DIFF
--- a/link-grammar/utilities.c
+++ b/link-grammar/utilities.c
@@ -88,9 +88,7 @@ char *
 strndup (const char *str, size_t size)
 {
 	size_t len;
-	char *result = (char *) NULL;
-
-	if ((char *) NULL == str) return (char *) NULL;
+	char *result;
 
 	len = strlen (str);
 	if (!len) return strdup ("");
--- a/link-grammar/utilities.c
+++ b/link-grammar/utilities.c
@@ -199,7 +199,7 @@ static int wctomb_check(char *s, wchar_t wc)
  * because the byte-counts might not match up, e.g. German ß and SS.
  * The correct long-term fix is to use ICU or glib g_utf8_strup(), etc.
  */
-void downcase_utf8_str(char *to, const char * from, size_t usize, locale_t locale_t)
+void downcase_utf8_str(char *to, const char * from, size_t usize, locale_t locale)
 {
 	wchar_t c;
 	int i, nbl, nbh;
@@ -216,7 +216,7 @@ void downcase_utf8_str(char *to, const char * from, size_t usize, locale_t local
 		prt_error("Error: Invalid UTF-8 string!\n");
 		return;
 	}
-	c = towlower_l(c, locale_t);
+	c = towlower_l(c, locale);
 	nbl = wctomb_check(low, c);
 
 	/* Check for error on an in-place copy */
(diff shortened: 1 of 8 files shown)


link-grammar/utilities.c
```diff
@@ -88,9 +88,7 @@ char *
 strndup (const char *str, size_t size)
 {
 	size_t len;
-	char *result = (char *) NULL;
-
-	if ((char *) NULL == str) return (char *) NULL;
+	char *result;
 
 	len = strlen (str);
```
linas: are you sure that strlen doesn't crash, if passed a null pointer?  I'm pretty sure that it used to...

linas:
You probably want to have a Changelog entry that tells users that the windows code has been cleaned up.

I'm assuming it will be easier for you if I don't merge immediately, and you push that update onto this branch?

link-grammar/utilities.c
ampli: [glibc strdup code](https://sourceware.org/git/?p=glibc.git;a=blob;f=string/strndup.c;h=fa0e7dcda1189a352411131ac5604cae1d06ad4c;hb=HEAD) also doesn't check for null there, so not checking that in our private copy (which is not used on Linux)  is in line with it, and I guess the LG code never provides a NULL there...

Also, Linux (and other) header files define the fisrt argument to be non-null:
```
extern char *strndup (const char *__string, size_t __n)
     __THROW __attribute_malloc__ __nonnull ((1));
#endif
```
And the compilers uses this `__nonnull` for optimizations, so NULL there is undefined anyway.

So I propose to leave it as is (without the check).

> You probably want to have a Changelog entry that tells users that the windows code has been cleaned up.

I pushed a ChangeLog commit.

BTW, a lot of Windows code cleanup is needed, in two areas:
1) Shared with non-Windows: Integer/ truncation and signedness-mix handling.
The GCC warning for that are just not enabled, but the equivalent ones are enabled by default on Windows and I didn't silenced them in the Project files.
If you enable them on GCC, you get really plenty of such warnings (in the order of 1000),
on things like assignment of int to short or char, and mixing unsigned and signed.
They seem not to be harmful but I think it will be a good idea to inspect and fix them all.
If we decide not to fix that, maybe we should disable these warnings on Windows too.

2) Also shared with non-Windows: Some functions get `signed char`argument when they are not defined for negative value, and we don't currently check the input (or cast the arguments to unsigned) so theoretically a problem can happen there.

3) The MSVC (and maybe MinGW - I need to check that) utf-8 to wchar_t code conversion doesn't handle characters out of the BMP, like emoticons. In the good case the handling is incorrect (like a wrong width), but in some cases this causes a buffer overflow. This can be fixed.

linas:
sgined vs unsigned, char vs short vs int. -- I tried "fixing" this several times, and might have made it worse. Its a bit of a mess that's hard to clean up.  The primary issue is that there are a few places where we want to use the smaller sizes, for space conservation, and this then trickles through the code.

It might be useful to `typedef size_t WordIdx;` (its already there in link-includes) and then try to use that consistently whereever a word index is used. Ditto for LinkIdx.

re: unicode emoticons. Don't know what to say. emoticons only got added to unicode a year or two ago.  Even though they've long been on the phones, perhaps they are only now coming to windows. Watch and wait is my knee-jerk reaction.  Buffer overflows are bad security-wise, but is anyone using link-grammar in a secure environment?

PR #603, opened 2017-10-29
Diagram generation: Fold long words

ampli:
- Fold long words.
- Add a FIXME for the PS generation problems.

Note that the right margin is actually 2 characters (like it was before this fix), because the screen width is set to one less than the real one. Is there a reason for that, or can we set screen_width to the exact real one? (This way it will also match a setting by `!width`.)

FIXME:
1) The last part of the diagram is still sometimes printed in a new row when it actually fits in the previous row (a more complex code is needed to fix that).
2) Maybe the trailing blanks in the diagram should be suppressed.

[ In the same occasion, something about the `!width=X` command. It is now ignored unless writing to file.
It can be changed to always obey a manual setup (when manually setting to 0 will return the current behaviour). This will enable manual adapting of the diagram width.
The change is simple, but it is not elegant in that it needs a special handling of the screen_width option (internally I would represent a manual fixed width by a negative value). It also has a limited usage. ]

### DIFF
--- a/link-grammar/print/print.c
+++ b/link-grammar/print/print.c
@@ -539,8 +539,11 @@ linkage_print_diagram_ctxt(const Linkage linkage,
 				if (k == cr) break;
 			}
 
-			/* We know it fits, so put it in this row */
-			pctx->link_heights[j] = row;
+			if (NULL != pctx->link_heights)
+			{
+				/* We know it fits, so put it in this row */
+				pctx->link_heights[j] = row;
+			}
 
 			if (2*row+2 > max_height-1) {
 				lgdebug(+9, "Extending rows up to %d.\n", (2*row+2)+HEIGHT_INC);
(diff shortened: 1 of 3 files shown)


link-grammar/print/print.c
```diff
@@ -743,8 +769,8 @@ char * linkage_print_diagram(const Linkage linkage, bool display_walls, size_t s
 	ps_ctxt_t ctx;
 	if (!linkage) return NULL;
 
-	ctx.link_heights = (int *) alloca(linkage->num_links * sizeof(int));
-	ctx.row_starts = (int *) alloca((linkage->num_words + 1) * sizeof(int));
+	ctx.link_heights = NULL;
```
linas: ??? Cough cough. How can this possibly be correct?

linas:
> screen_width
I know of no technical reason that its off by two. In fact, I've always found this kind-of annoying.

> trailing blanks
This too has been annoying, e.g when cut-n-pasting the diagrams.

> !width=X 
Not worth fixing.

ampli:
>>screen_width

>I know of no technical reason that its off by two. In fact, I've always found this kind-of annoying.

Maybe originally it was a defensive programming against an off-by-one bug. I will make screen_width the exact width, and if there will be an of-by-one bug (in the ASCII diagram generation, the only place it is used for now) I will fix it.
>>trailing blanks

>This too has been annoying, e.g when cut-n-pasting the diagrams.

I will remove them.
>>!width=X

>Not worth fixing.

OK.

link-grammar/print/print.c
ampli: `link_heights` and `row_starts` are need only for the postscript generation. So for the ASCII diagram they can just be ignored.

linas: ah I see now.

PR #617, opened 2017-11-17
Fix unit tokenization to remove ambiguity

ampli:
This PR issues units with their subscript. This fixes problems due to a previous ambiguous interpretations of unsubscripted unit strings.

I compared all the linkages of the sentences in data/en/corpus-{basic,fixes}.batch produced
from the current and fixed code. Here my notes on the differences due to this fix.

1) `As` is no more interpreted as `A s`.
Preventing bogus linkages for these sentences from en/corpus-{basic,fixes}.batch:
```
As had been expected, the party was a big success
As guessed, he was lying.
As mentioned before, we are creating a business service.
As promised, she left
```

The prevented bogus linkages:
```
LEFT-WALL a  s.u guessed.q-d , he was.v-d lying .
LEFT-WALL a  s.u mentioned.v before , we are.v creating.v a business.s service.s .
LEFT-WALL a  s.u promised.v-d ,  she left.v-d
```

2) The following sentence and all similar ones now have less linkages, a thing that indicates a shortcoming in the current separation algo:

```
linkparser> We bought 200ft. of lumber.
...
LEFT-WALL we bought.v-d 200[!] ft..u of lumber.n-u .
...
LEFT-WALL we bought.v-d 200[!] ft.u . of lumber.n-u . <-- missing
...
LEFT-WALL we bought.v-d 200[!] ft..u of lumber.n-u .
...
LEFT-WALL we bought.v-d 200[!] ft.u . of lumber.n-u . <-- missing
```

The reason is that without issuing with subscript, `ft.` gets issued, and then at a next tokenization round it get split again in a split_right() of RPUNC to `ft .`. With this new code, after `ft.u` is issued, it cannot get split farther.

In all the cases in en/corpus-fixes.batch it is not harmful since the dict handles both  `ft .` and `ft.` . Moreover, we get half the number of linkages for every unit which has such two versions (with and w/o a dot). However, this indicates that the separation algorithm is not general enough.

Possible FIXME: Use backtracking + remove the units versions with the dot (supposing they are supported with a following dot).

3) In these sentence and additional similar ones:
`the weight was 7grams`

Currently (without this fix) we get these 2 main linkages:
```
LEFT-WALL the weight.n-u was.v-d 7 grams.n
LEFT-WALL the weight.s was.v-d 7 grams.u
```
With this fix, we get only the linkage with `grams.u`. This seems to me better.
(Similarly: `The ferrite core is wrapped with 24gauge wire.` Only `gauge.u` with this fix.)

Similarly for (and additional similar ones):
`It was 10:30AM in the morning.`

Now (without the unit fix) - note the linkage marked with <---:
```
LEFT-WALL it was.v-d 10:30[!HMS-TIME] AM.ti in.r the morning.n-u .
LEFT-WALL it was.v-d 10:30AM[!HMS-TIME] in.r the morning.s .
LEFT-WALL it was.v-d 10:30[!HMS-TIME] AM.l in.r the morning.n-u . <---
```
With this fix, we don't get the `AM.l` one which is clearly wrong.

Here, with this fix we don't get this bogus full linkage:
`It was 10:555:88AM in the morning.`
Before: `LEFT-WALL it was.v-d 10:555:88[?].v AM.l in.r the morning.s .`
With this fix: `LEFT-WALL it was.v-d 10:555:88[?].a [AM] in.r the morning.n-u .`

BTW, maybe including AM etc. in the regex is not needed, as the dict handles it.

4) These words from en/corpus-fixes.batch still get separated to "units", but this time with a unit subscript, so in addition to reducing bogus linkages, the linkages are faster because there are less disjuncts/connectors in the sentence, and less linkages are with bad-morphism.
```
As Ash Kids SAN Sam Shh Shhh TNT TV add all call did lamp lid mall mind minds
mold odd sin sins slid small smidgin (s.u m.u id.u g.u in.u)
bases (base s)
```
The last one is due to `base_pairs` in `units.1`, which only its first word gets to UNITS (as designed).
With this fix it is broken to `base s.u` which has less chase to get misinterpreted.

5) This sentence from en/corpus-fixes.batch still is not handled by the dict:
`Apply 12 ft.lbs. of torque.`
Results in:
`LEFT-WALL apply.v 12 ft..u [lbs.] of torque.n-u .`

6) This sentence has no full parse with this fix, and demonstrates a problem in the dict:
`I can't decide between the 7:30AM and the 9:30AM appointment.`
Before: `LEFT-WALL I.p can't decide.v between the 7:30[!] AM.l and.j-n the 9:30AM[!] appointment.n-u .` 
With this fix: `LEFT-WALL I.p can't decide.v between the 7:30AM[!HMS-TIME] and.j-ru [the] 9:30AM[!HMS-TIME] appointment.n-u .`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,7 +7,7 @@ Version 5.4.3 (XXX 2017)
  * Fix square area and cubic volume measurements (English dict).
  * Fix assorted exclamations and responses (English dict).
  * Fix displaying random linkages on Windows.
-
+ * Fix unit tokenization to remove ambiguity.
 
 Version 5.4.2 (19 October 2017)
  * Fix man page build (broken in 5.4.1)
--- a/data/en/4.0.affix
+++ b/data/en/4.0.affix
@@ -60,3 +60,5 @@ y' w/: PRE+;
 % units.6 contains just a single, sole slash in it. This allows units
 % such as mL/s to be split at the slash.
 /en/words/units.6: UNITS+;
+%
+/en/words/units.a: UNITS+;
--- a/data/en/4.0.affix
+++ b/data/en/4.0.affix
@@ -3,18 +3,14 @@
 % i.e. spaces are inserted between the affix and the word itself.
 %
 % Some of the funky UTF-8 parenthesis are used in Asian texts.
-% In order to allow single straight quote ' and double straight quote ''
-% to be stripped off from both the left and the right, they are
-% distinguished by the suffix .x and .y (as as Mr.x Mrs.x or Jr.y Sr.y)
-%
 % 。is an end-of-sentence marker used in Japanese texts.
 
 % Punctuation appearing on the right-side of words.
-% Note: the ellipsis ....y must appear *before* the dot ".", else the
+% Note: the ellipsis "..." must appear *before* the dot ".", else the
 % splitting won't work right.
-")" "}" "]" ">" "".y" » 〉 ） 〕 》 】 ］ 』 」 "’’" "’" ” ''.y '.y `.y
-"%" "," ....y "." 。 ‧ ":" ";" "?" "!" ‽ ؟ ？ ！
-_ ‐ ‑ ‒ – — ― ….y ━.y –.y ー.y ‐.y 、.y
+")" "}" "]" ">" """ » 〉 ） 〕 》 】 ］ 』 」 "’’" "’" ” '' ' `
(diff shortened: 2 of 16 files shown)


linas:
> maybe including AM etc. in the regex is not needed, as the dict handles it.

Yes.

linas:
> Apply 12 ft.lbs. of torque.

I just applied a cheesy fix for this. Its too permissive, but it works.

PR #618, opened 2017-11-18
Windows: Fix a Unicode linkage graph rendering problem

ampli:
The main fix in this PR is for a bug in the recently new change in `utf8_next()` for Windows: It just always returns 1 for valid utf-8 characters (on invalid chars it skips to their end).  By chance the rendering is still done reasonably well on Windows even with this bug.
Recently I refactored the character width computation to its own function and added there error checking. I then noticed that this error actually happens.

In the same occasion I added some minor changes.
I also added a FIXME proposal for fixing character width finding on Windows (badly fails for code points that need 2 wchar_t to represent them - as mentioned in needed fix#3 mentioned in PR #601 ): Just use char32_t with wcstoc32() (POSIX, added on C11).

(Pushed again to refix.)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,7 @@ Version 5.4.3 (XXX 2017)
  * Fix assorted exclamations and responses (English dict).
  * Fix displaying random linkages on Windows.
  * Fix unit tokenization to remove ambiguity.
+ * Fix utf8-related bug on Windows that could affect printing.
 
 Version 5.4.2 (19 October 2017)
  * Fix man page build (broken in 5.4.1)
--- a/link-grammar/print/print-util.c
+++ b/link-grammar/print/print-util.c
@@ -186,7 +186,7 @@ size_t append_utf8_char(dyn_str * string, const char * mbs)
 {
 	/* Copy exactly one multi-byte character to buf */
 	char buf[10];
-	size_t n = utf8_next(mbs);
+	size_t n = utf8_charlen(mbs);
 
 	assert(n<10, "Multi-byte character is too long!");
 	strncpy(buf, mbs, n);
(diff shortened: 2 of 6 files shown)


linas:
Meta-question -- what happens if we pass in invalid utf8?  If someone creates a custom-craffted string where the first byte makes it look like a 4-byte code point, the next byte is null, so that it is designed to cause a buffer overflow?   You don't have to answer this question.

I used to be sure that there were no buffer overflows, but now, with all this utf8 code, I am no longer sure that we do accurate bounds-checking.

ampli:
In  `separate_sentence()` we have a check by mbrtowc() that is said to detect that.
For non-Windows system it is from clib.
For Windows it is actually `lg_mbrtowc()` which now uses the unsafe `utf8_charlen()`. However, the rest of the code in it is also buggy (and unsafe) because code-points outside of the BMP are said to take two `wchar_t`, while only one is allocated by the caller.

From reading on Unicode portability it is clear that using `wchar_t` is not portable at all. On most POSIX systems it is 32 bit, but on Windows and some other systems it is 16 bit, and uses 2 `wchar_t` elements for code-points out of the BMP (i.e. UTF-16 is a variable-length encoding, a thing that the LG code doesn't take into account at all).

A portable solution can be to use `char32_t`, which is in C11 and is supported by recent versions of MSVC. I found that Windows doesn't support outside-of-the-BMP values in its isw*() functions (they don't fault even in debug mode, and return different incorrect values for different argument values which are outside-of-the-BMP). However, our code can just pretend that isw*() returns false for outside-of-the-BMP values. Such values are supported by `mk_wcwidth()`. We can use `char32_t` in all the places that don't interface with the Windows (e.g besides locales and maybe filesystem - even though non-ASCII for files/directories is not yet supported in the LG lib on Windows).

I can send PR's, if desired, for:
1. Use `char32_t` instead of `wchar_t` (including fixing `utf8_charlen()` to use it).
2. Use the libc `wcwidth()` (POSIX 2001,2008) (with a configuration switch if desired).
3. Simplify locale support (since I learned some related new things from the time I implemented it).

linas:
Not sure what to say. If this bugs you, then send PRs.   I suppose its easier now, while this is still fresh in your head.  I'm up to my neck in other urgent issues.

PR #621, opened 2017-12-16
do_count() speed improvements

ampli:
It turns out do_count() doesn't use its pseudocounting in the most efficient way.
For  example:
1. If `Lmatch=false` but `Rmatch=true`, there is no need to even make pseudocounting unless `le==NULL`.
2. If `pseudototal=true` and  `leftcount=true', there is sometimes still no need to perform LHS counting.
3. If `leftcount=true' and `rightcount=true` (but the "but not" countings are false), there is no need to perform the RHS counting if the actual LHS counting turns put to be false.
4. The pseudocouning table lookups can be cached and used in the real countig if needed.
5. ...

Several more  significant optimizations are be done.
See the commits:
- Optimize the use of the pseudocounting
- Optimize further

Implementing these optimizations and several more smaller ones  makes the `en/corpus-fixes.batch` run 10+% faster.
However, these optimizations affect longer sentences even more, and especially when parsing with null words. For example, in the "The problem is: sentence, the CPU saving is 40% (when `!null=T`).
These optimizations doesn't have much effect on Russian (it seems its execution path doesn't involve the pseudocounting inefficiencies that got corrected here).

Regretfully had to modify some original comments in order to match them to the new code.

In WIP branches I have several more modifications that speed-up the classic parsing.
They all modify the master branch as it is now, so I will need to reimplement/rebase them after each PR (lake this one) is accepted.

### DIFF
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -41,7 +41,7 @@ struct count_context_s
 	bool    islands_ok;
 	bool    null_links;
 	bool    exhausted;
-	int     checktimer;  /* Avoid excess system calls */
+	unsigned checktimer;  /* Avoid excess system calls */
 	int     table_size;
 	int     log2_table_size;
 	Table_connector ** table;
@@ -133,11 +133,11 @@ find_table_pointer(count_context_t *ctxt,
 	/* Create a new connector only if resources are exhausted.
 	 * (???) Huh? I guess we're in panic parse mode in that case.
 	 * checktimer is a device to avoid a gazillion system calls
-	 * to get the timer value. On circa-2009 machines, it results
-	 * in maybe 5-10 timer calls per second.
+	 * to get the timer value. On circa-2017 machines, it results
+	 * in about 0.5-1 timer calls per second.
 	 */
 	ctxt->checktimer ++;
-	if (ctxt->exhausted || ((0 == ctxt->checktimer%1450100) &&
+	if (ctxt->exhausted || ((0 == ctxt->checktimer%(1<<21)) &&
 	                       (ctxt->current_resources != NULL) &&
 	                       resources_exhausted(ctxt->current_resources)))
 	{
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -159,6 +159,13 @@ Count_bin* table_lookup(count_context_t * ctxt,
 	if (t == NULL) return NULL; else return &t->count;
 }
 
+#define NO_COUNT -1
+#ifdef PERFORM_COUNT_HISTOGRAMMING
+Count_bin count_unknown = {.total = NO_COUNT};
+#else
+Count_bin count_unknown = NO_COUNT;
+#endif
+
 /**
  * psuedocount is used to check to see if a parse is even possible,
  * so that we don't waste cpu time performing an actual count, only
@@ -171,14 +178,13 @@ Count_bin* table_lookup(count_context_t * ctxt,
  * in the hash table, and its zero, then we know, for sure, that the
  * count is zero.
  */
-static bool pseudocount(count_context_t * ctxt,
+static Count_bin pseudocount(count_context_t * ctxt,
                        int lw, int rw, Connector *le, Connector *re,
                        unsigned int null_count)
 {
 	Count_bin * count = table_lookup(ctxt, lw, rw, le, re, null_count);
-	if (NULL == count) return true;
-	if (hist_total(count) == 0) return false;
-	return true;
+	if (NULL == count) return count_unknown;
+	return *count;
 }
 
 /**
@@ -374,9 +380,17 @@ static Count_bin do_count(fast_matcher_t *mchxt,
 			{
 				bool leftpcount = false;
 				bool rightpcount = false;
-				bool pseudototal = false;
-				bool lmbnrptotal = false;
-				bool rmbnlptotal = false;
+
+				Count_bin l_any;
+				Count_bin l_cmulti = NO_COUNT;
+				Count_bin l_dmulti = NO_COUNT;
+				Count_bin l_dcmulti = NO_COUNT;
+				Count_bin l_bnr = zero;
+				Count_bin r_any = NO_COUNT; /* gcc: may be used uninitialized */
+				Count_bin r_cmulti = NO_COUNT;
+				Count_bin r_dmulti = NO_COUNT;
+				Count_bin r_dcmulti = NO_COUNT;
+				Count_bin r_bnl = zero;
 
 				rnull_cnt = null_count - lnull_cnt;
 				/* Now lnull_cnt and rnull_cnt are the costs we're assigning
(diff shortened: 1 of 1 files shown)


linas:
thanks!

linas:
BTW, the "histogramming" code was an attempt to understand what was happening inside of here., and maybe find a way of restructuring the algorithm to be faster or more direct.

Out of idle curiosity, where does this leave the SAT solver? is it still faster on long sentences, or are they tied?

ampli:
> BTW, the "histogramming" code was an attempt to understand what was happening inside of here., and maybe find a way of restructuring the algorithm to be faster or more direct.

How can it be used, especially the info about disjunct cost?

In an unpublished modification the `do_count()` tracing also traces the exact recursive `do_count()` calls (by labels that I gave them). This allows to easily see, among other stats, how many times each one counts 0 and non-0, or find the result in the table. For example, is the result for en/corpus-fixes.batch - `F` is 0 and `T` is >0 (it may be clear from the source code which label refers to which `do_count()` call):
```
1116239 F DC_L-0
  68947 F DC_L-RBNL
 407777 F DC_L-cmulti
  11759 F DC_L-dmulti
 401665 F DC_R-0
1350715 F DC_R-LBNR
 137137 F DC_R-cmulti
  41218 F DC_R-dmulti
1469430 T DC_L-0
  38609 T DC_L-RBNL
  37388 T DC_L-cmulti
    939 T DC_L-dmulti
 178705 T DC_R-0
 153854 T DC_R-LBNR
   1450 T DC_R-cmulti
   1248 T DC_R-dmulti
```
The Russian one has totally different ratios:
```  21952 F DC_L-0
  11473 F DC_L-RBNL
    284 F DC_L-cmulti
    859 F DC_L-dmulti
  14641 F DC_R-0
  28753 F DC_R-LBNR
   6054 F DC_R-cmulti
     60 F DC_R-dmulti
  43664 T DC_L-0
  20393 T DC_L-RBNL
     12 T DC_L-cmulti
     17 T DC_L-dmulti
  31841 T DC_R-0
  14931 T DC_R-LBNR
    160 T DC_R-cmulti
```

**I noted something strange** - the condition `le->multi && d->left->multi` is never true!
I don't know yet this is just because there are no such dictionary definitions, or this is a check that cannot be true (If it cannot be true, the code can be sped up in 2 ways).

(I didn't include the code for these stats in the PR because it adds a `do_count()` optional argument (enabled in DEBUG mode) that clutters the source code.


> Out of idle curiosity, where does this leave the SAT solver? is it still faster on long sentences, or are they tied?

I didn't try, but I believe it is still much faster for long sentences. I also have a fix that makes it twice faster for very long sentences (it uses a more efficient XOR encoding). The SAT parser can also be made very much more efficient, even for short sentences.

I think that the main problem with the current SAT parser, that makes it unuseful, is that it cannot parse with null links.
Several months ago I found how to do that in apparent efficient way, but stopped to work on it.
Since this topic seems to me very interesting, and since I think the SAT parser has more potential than the classic parser,  I intend to return to it after I convert to PRs most of the (many) WIP branches I have and implement some top things in my fix list for LG (most of it is not published).

BTW, I also once started to investigate the SAT-encoding of post-processing - now most of it is done per solution. Making it with SAT-encoding has a potential to vastly speed up handling linkages with null-links.
Sane-morphism can also be SAT-encoded, so all the solutions will automatically be valid
(I didn't investigate yet the possibility of encoding the wordgraph directly).

Some WIP branches I now would like to convert to PRs are:
- optional 32 bit counter (intended to speed up very long sentences).
- connector enumeration + 32 bit connectors (it speeds up even short sentences).
- very short connectors - I found how to shrink them to 4 bytes (only partially done).
- new table_connector hash table for do_count() (more efficient and extendable). The table size and its current lack of hit locality is currently a major overhead for very long sentences.
- connector/disjunct packing - putting them in an array instead of a list speeds up do_count() more than the overhead of constructing such an array even for short sentences.
- implementing things we discussed in the issues.
- some more things that I need to check how much work is needed to finish them, like link-parser rewrite in Python (mostly implemented but needs debugging) and dict handling of capitalization.

I also found that more caching can be done in the fast-matcher, and would like to try it (it is a major overhead).
In addition I'm also investigating extracting more info from pruning in order to speed up do_count(), especially find block of words that can be skipped in its main loop - currently only blocks at the edges can be skipped, but not in the middle. If middle blocks can be skipped, long sentences can be "shortened".
(And I didn't forget a new tokenizer and a real Hebrew support...).

linas:
I no longer remember the details of what I was trying to do with the histogramming; however, I had two general principles I wanted to promote: 

(1) if every parse always has a certain link, then we are "certain" that the link is "correct" (maybe set its cost to zero)

(2) if a certain sequence of words can be linked in many different ways, but the rest of the sentence can only be linked one way, then indicate that sequence as being "variable" or "uncertain". 

(3) if a given link occurs only in a very small number of parses, maybe its "wrong".

Or something like that.  I was trying to match up what the algorithm is doing with some crude model of human psychology or "reasonableness" in a parse interpretation.

I run a bit cold on the SAT solver mostly because the algorithm for it doesn't feel "natural". in the sense of preferring certain types of links, or believing that some links are better than others.  If a sentence doesn't parse, it doesn't tell you *why* it doesn't parse.

For example: I meant to say "this is a test" but wrote "this is test" by accident.  How can I get link-grammar to guess that the word "a" is missing, and where it should go?

ampli:
> (1) if every parse always has a certain link, then we are "certain" that the link is "correct" (maybe set its cost to zero)
> 
> (2) if a certain sequence of words can be linked in many different ways, but the rest of the sentence can only be linked one way, then indicate that sequence as being "variable" or "uncertain".
> 
> (3) if a given link occurs only in a very small number of parses, maybe its "wrong".
> 
> Or something like that. I was trying to match up what the algorithm is doing with some crude model of human psychology or "reasonableness" in a parse interpretation.

Maybe we can post-process the linkages (say first 1000) of the current copus batches to find examples of each of the above, and then see if these heuristics are right.



> I run a bit cold on the SAT solver mostly because the algorithm for it doesn't feel "natural". in the sense of preferring certain types of links, or believing that some links are better than others. If a sentence doesn't parse, it doesn't tell you why it doesn't parse.
> 
> For example: I meant to say "this is a test" but wrote "this is test" by accident. How can I get link-grammar to guess that the word "a" is missing, and where it should go?
> 
>  Delete branch

Is the "cllasic parse" better in that respect?

Regarding SAT, it can be used in a different mode than the current SAT parser uses. We can define constraints that must be true, like "no cross links", "no more than one selected disjunct per word", etc. and ask it to minimize the number of unsatisfied (rest of) constraints. The idea is to get these way "partial best linkages" in which some disjuncts don't have full connectivity and find words that can complete these connectivity. This is an extension of the way in which I would like to implement a SAT parser with null links - it minimizes the number of unsatisfied clauses that state that each word participates in the linkage (currently all of these clauses must be true unless a word is marked as "optional").

How this can be implemented with the classic parser?

linas:
> Maybe we can post-process the linkages (say first 1000) 

Although it would be nice to solve 1), 2) and 3) "in practice", I'm more interested in solving them "in principle" -- understanding what kind of parse algorithm works well, with "human-like" characteristics.  Does it the way that a human brain might be doing it. I have no scientific way of acheiving this, except to wonder if the algo "feels natural".  I used to think that the viterbi stuff "felt natural", but I no longer believe that.  Some insight can be gained from papers in psychological linguistics. Some insight can be gained from neural nets. But this is all very squishy, very seat-of-the-pants work.

linas:
On a related note: a possibly/probably important thing to do would be to replace the no-links-cross constraint by the landmark transitivity constraint mentioned in the README file.  Some initial work on this was done, but I do not recall the status.

ampli:
> On a related note: a possibly/probably important thing to do would be to replace the no-links-cross constraint by the landmark transitivity constraint mentioned in the README file. Some initial work on this was done, but I do not recall the status.

The classic parse algo is based on the no-links-cross constraint, so I cannot imagine how it can be replaced there. However, in the SAT parser it is easy to disable the no-links-cross constraint (just comment it out) and to define any other additional constraint (if it is well-umderstood).

linas:
> (if it is well-umderstood)

Somewhere this is clearly explained. Maybe even in the README. Its not that bad, its fairly direct.

PR #622, opened 2017-12-17
Fix memory leaks

ampli:
1. In pp_lexer (accumulated - every time a dict is opened/closed).
2. In the same occasion I fixed also the one in `dictionary_set_data_dir()` (minor - not accumulated).

BTW, now we have again the "problem" of pp_lexer exiting on file format errors.
It can be fixed if desired (but maybe it is not important enough).


linas:
thanks

PR #624, opened 2017-12-22
!file fixes

ampli:
While playing the the new !file autocomplete, I found that the ability to press return to get more linkages using `!file` when not in `batch` mode, got lost. This happened in 5.3.8. Since it is not hard to restore this ability, I did it here. Just because it is trivial, I added the ability to also interpret `!` commands in that case.

In the same occasion I also fixed 2 problems that always existed in `!file`:
1. Notifying if the file is actually a directory (instead of silently doing nothing).
2. Reporting file read error.

### DIFF
--- a/link-parser/link-parser.c
+++ b/link-parser/link-parser.c
@@ -37,6 +37,7 @@
 #include <locale.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


linas:
thanks!

PR #625, opened 2017-12-23
link-parser: !file: Check stat() returned value

ampli:
Strange things may happen when not checking the result of system calls...
In that case a nonexistent file may be considered a directory...


linas:
:-)

linas:
just checkout errno when a syscall succeeds...

PR #628, opened 2018-01-03
Upload the SAT link parser PDF

leungmanhin:
for anyone who is interested


linas:
Do you have the source for this?   I'm not sure I've ever seen this document before!  Maybe I have but if so, I certainly forgot.  Some of the details have changed; for example, fat links no longer exist; post-processing has been redesigned.

leungmanhin:
Yes it's 10 years ago, should still be an interesting thing to read though. I got this from Ben this morning.

PR #636, opened 2018-01-14
Fix idiom string incrementing

ampli:
Without this fix, the idiom names start as:
BUFF=ID
BUFF=ID
BUFF=ID^A
BUFF=ID^A
BUFF=ID^B
BUFF=ID^B
BUFF=ID^C
BUFF=ID^C
because the '\0' at the end is incremented.
At some early point there are even 16 A's and a buffer access violation happens:
idiom.c:210:31: runtime error: index 16 out of bounds for type 'char [16]'

### DIFF
--- a/link-grammar/dict-common/idiom.c
+++ b/link-grammar/dict-common/idiom.c
@@ -185,7 +185,7 @@ static Dict_node * make_idiom_Dict_nodes(Dictionary dict, const char * string)
 
 static void increment_current_name(Dictionary dict)
 {
-	int i = IDIOM_LINK_SZ-1;
+	int i = IDIOM_LINK_SZ-2;
 
 	do
 	{


linas:
16 A's -- when?  we can make this number larger -- 16 already seemed absurdly large... that would be 26^16 = 43608742899428874059776 idioms .. right?

ampli:
16 A's are already far too much.
The original code that you recently modified contained only 8 A's.
` (Dom Lachowicz  2005-02-15 13:55:43 +0000 187) static char current_name[] = "AAAAAAAA";`
Since this is also more than enough ( 208827064576), I wondered why you increased it to 15 A's...
```
#define IDIOM_LINK_SZ 16
char            current_idiom[IDIOM_LINK_SZ];
...
memset(dict->current_idiom, 'A', IDIOM_LINK_SZ-1
```
Anyway, the buggy code increments the `'\0'` at the end of the 15 A's string, and eventually it becomes
an `A`,  and then the following line gets out of the `current_idiom[16]` array:
`for (i=0; dict->current_idiom[i] == 'A'; i++)`

Note the extra work need to scan a string of so many A's. **I propose to reduce their number to only 4**, which is enough for 456976 idioms. (If there is a possibility that an initial unsupervised language learning program may initially treats many combinations as idioms before it refines, then maybe 5 A's...)

linas:
yes, reducing to 4 or 5 is OK. ... or just set it to 8, that way it it fills a full byte.  

> extra work need to scan a string of so many A

where?

ampli:
>>extra work need to scan a string of so many A

>where?

`for (i=0; dict->current_idiom[i] == 'A'; i++)`

linas:
oh. OK.  changed it to 4 just now

PR #639, opened 2018-01-19
Main README fixes

ampli:
1. Add `flex` and `ant` to the list of packages needed for development.
2. Unify the two Java Bindings sections.

I also noted that **all** the links regarding **landmark transitivity** don't work any more.
I wanted to replace them by working ones, but could not find any. I would like to read them in order to have a better insight on implementing landmark transitivity (discussion at PR [#621](https://github.com/opencog/link-grammar/pull/621#issuecomment-352299657)).

BTW, the info on SAT performance (A performance diary) seem to be bogus. The reported times are the times needed to load the dictionary. I tested the mentioned version (5.1.0) and it indeed doesn't actually perform the linkage. I propose to update this info.
In any case, comparing the performance of different library versions needs these considerations:
- Use the same "short" option (older libraries use a smaller value, which cause faster linkages). 
- As the English dictionary becomes bigger and more complex over versions, this increases the linkage times even if the same library version is used.
- The sentence `In vivo studies...` is not good for such a benchmark because it contains "spelling errors". Depending on the spelling package version which is used, there may be different numbers of spelling corrections, which make the linkage time vary. Especially, it becomes very big if no spelling corrections are found, due to the need of linkages with nulls (this what happens when I tested it with the current library version).

All of these need to be considered when comparing the performance.

An  additional proposals: Remove the comment on `libedit` UTF-8 support on Ubuntu 14.04 and Mint 17 after their End of Life (in a few months).

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,6 +356,7 @@ gcc-c++ (for the SAT solver)<br>
 autoconf<br>
 autoconf-archive<br>
 swig (for language bindings)<br>
+flex<br>
 graphviz (if you like to use the word-graph display feature)
 
 The GitHub version doesn't include a `configure` script.
--- a/README.md
+++ b/README.md
@@ -357,6 +357,7 @@ autoconf<br>
 autoconf-archive<br>
 swig (for language bindings)<br>
 flex<br>
+ant (for Java bindings)<br>
 graphviz (if you like to use the word-graph display feature)
 
 The GitHub version doesn't include a `configure` script.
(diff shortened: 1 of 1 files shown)


linas:
> A performance diary

Best to just remove this section entirely. It was meant to be a a personal note-to-self to make sure accidental performance-damaging bugs were not introduced.  For this test, it doesn't much matter if the sentences are grammatical or not...  I was also hoping to see an obvious time vs. length of sentence behavior, but no such luck.   Having a graph of average sent length vs time would be .. interesting to see.

>libedit

OK

PR #643, opened 2018-01-22
Move pp data from PP struct to linkage.

linas:
The goal of this change was to gain a minor amount of performance by avoiding the computation of domain names, unless they are printed. However, to accomplish this, one hs to create a per-linkage copy of PP_data, which creates a lot of memory-pool churn, almost surely increasing RAM usage significantly, and, in general making things slower rather than faster.

I will probably reject this pull request.  Its here temporarily, for further review.

### DIFF
--- a/link-grammar/api-types.h
+++ b/link-grammar/api-types.h
@@ -24,7 +24,6 @@ typedef struct Exp_struct Exp;
 typedef struct Connector_struct Connector;
 typedef struct Linkage_info_struct Linkage_info;
 typedef struct Postprocessor_s Postprocessor;
-typedef struct PP_info_s PP_info;
 typedef struct Resources_s * Resources;
 
 /* Some of the more obscure typedefs */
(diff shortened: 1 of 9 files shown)


linas:
Amir, Can you look this over and comment?  This is a larger chunk of work that failed, wonder if you have any bright ideas.

Yes, I know you asked me some me some questins that I have not answered. I've been procrastinating, they were hard questions. I will answer tomorrow for sure :-)

linas:
p.s. its about 6% slower according to one (brutal) measurement - parsing `corpus-failures.batch`

ampli:
Since making and processing a linkage is memory-access bound, and for long sentences the CPU is mostly waiting for memory reads, it is indeed not a good idea to increase the memory footage per linkage.

I would propose another approach that reduces as much as possible resource use on the main execution path on the expense of adding overhead to the rare use of `linkage_set_domain_names()`, of doing these two things:
- In the normal linkage post-processing just don't call `build_type_array()`.
- When `linkage_print_links_and_domains()` is called, repeat the post-processing for the particular linkage (or the minimal needed subset of it) , which in this time (only) will also call `build_type_array()` (by checking an added flag).

The benefits of this approach is that all the extra type-array processing is prevented when the linkage is done - no need even to create anything in the linkage struct. On the other hand the extra overhead in the case of calling `linkage_print_links_and_domains()` doesn't matter. Also, the needed change for this approach seems to me much smaller.

linas:
thank you, yes, quite right!

PR #655, opened 2018-01-23
bugfix wordgraph crash on invalid data

linas:
Also make aspell thread-safe; it wasn't, before.

Cleanup: re-organize the wordgraph code, so that the delete functions are not in api.c

Move wordgraph display code to it's own file.


linas:
OK. yeah

PR #658, opened 2018-01-24
 Fix test case EErrorFacilityTestCase

ampli:
Problem description: When the default library error handler is called by Python code, its data argument points to a variable from a frame that has just got stale.

### DIFF
--- a/bindings/swig/link_grammar.i
+++ b/bindings/swig/link_grammar.i
@@ -296,7 +296,9 @@ static void PythonCallBack(lg_errinfo *lge, void *func_and_data)
       }
 
       if (error) return NULL;
-      $1 = &arg;
+
+      $1 = (int*)malloc(sizeof(int));
+      *$1 = arg;
    }
 }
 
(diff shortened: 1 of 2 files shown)


linas:
Huh Interesting.  Did the static analyzer catch this?

ampli:
I found it because suddenly the tests started to trigger ASAN in debug mode (I always compile with ASAN/UBSAN in debug mode - maybe this should be officially added to `configure.ac`, and I guess a  Python update or the like made some change that uncovered it).
I then looked in `lg_python_wrap.c`c at the point the default handler is invoked and the problem was obvious.
But because of your question I tried `cppcheck lg_python_wrap.c`. And indeed it clearly points out this error (the one at 5998):
```
[lg_python_wrap.cc:3653]: (error) Memory leak: result
[lg_python_wrap.cc:6001]: (error) Dead pointer usage. Pointer 'arg2' is dead if it has been assigned '&arg' at line 5998.
```
It also pointed out another problem (at 3653) that I will investigate - the call is intended to create a new `lg_errinfo` object, so it needs to allocate memory for it, that I supposed will get deallocated when the object gets out of scope. I don't know yet how `cppcheck` may know this cannot happen (or maybe this is just a false positive).

Issue #652, opened 2018-01-23
Static Analysis Report


gy741:
Hello,

I do not know if this information will help you.

I checked the source code using cppcheck.

I received the following report.

There may be a false here.

Thanks.

```
[bindings/java-jni/jni-client.c:86]: (error) Memory leak: msg
[bindings/lisp/link-interface.cl:28]: (error) No pair for character ('). Can't process file. File is either invalid or unicode, which is currently not supported.
[link-grammar/dict-common/dict-impl.c:148]: (warning) Obsolete function 'alloca' called. In C99 and later it is recommended to use a variable length array instead.
[link-grammar/dict-common/dict-impl.c:87]: (warning) Obsolete function 'alloca' called. In C99 and later it is recommended to use a variable length array instead.
[link-grammar/dict-common/dict-utils.c:121] -> [link-grammar/dict-common/dict-utils.c:119]: (warning) Either the condition 'el1!=NULL' is redundant or there is possible null pointer dereference: el1.
[link-grammar/dict-common/dict-utils.c:123] -> [link-grammar/dict-common/dict-utils.c:119]: (warning) Either the condition 'el2!=NULL' is redundant or there is possible null pointer dereference: el2.
[link-grammar/dict-common/print-dict.c:125] -> [link-grammar/dict-common/print-dict.c:137]: (warning) Either the condition 'el' is redundant or there is possible null pointer dereference: el.
[link-grammar/error.c:235]: (warning) Obsolete function 'alloca' called. In C99 and later it is recommended to use a variable length array instead.
[link-grammar/error.c:276]: (warning) Obsolete function 'alloca' called. In C99 and later it is recommended to use a variable length array instead.
[link-grammar/error.c:371]: (warning) Obsolete function 'alloca' called. In C99 and later it is recommended to use a variable length array instead.
[link-grammar/error.c:387]: (error) va_list 'given_features' was opened but not closed by va_end().
[link-grammar/linkage/linkage.c:77]: (warning) Obsolete function 'alloca' called. In C99 and later it is recommended to use a variable length array instead.
[link-grammar/linkage/linkage.c:204]: (warning) Obsolete function 'alloca' called. In C99 and later it is recommended to use a variable length array instead.
[link-grammar/linkage/linkage.c:264]: (warning) Obsolete function 'alloca' called. In C99 and later it is recommended to use a variable length array instead.
[...]

PR #660, opened 2018-01-25
Handling the PVS-Studio static analysis from #652 ++

ampli:
Fix problems detected in the PVS-Studio static analysis from #652.

Bug fixed:
- Fix a bug in my previous fix (5d513a0).
- Buffer overrun on long regex names.
- Correct X-node allocation size in the SAt parser.
- Make sure EOF can be represented in dict read
- Fix end location reporting of spell-corrected input word (linkage_get_word_*_end()).

Also:
- Remove redundant duplicate condition (typo).
- Comment out a null-effect loop break and mark iit for review (most probably it remained from a code rewrite).

In this occasion of inspecting the code. I made efficiency fixes in:
- read_entry()
- make_idiom_Dict_nodes()

### DIFF
--- a/link-grammar/dict-common/dict-common.h
+++ b/link-grammar/dict-common/dict-common.h
@@ -123,7 +123,7 @@ struct Dictionary_s
 	const char    * pin;
 	bool            recursive_error;
 	bool            is_special;
-	char            already_got_it;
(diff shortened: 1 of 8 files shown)


linas:
Thanks

Issue #665, opened 2018-02-01
heap-buffer-overflow in compute_chosen_words


gy741:
Hello.

I found a heap-buffer-overflow bug in link-grammar.

This problem is detected by valgrind or AddressSanitizer.

Please confirm.

Thanks.

OS: CentOS 7 64bit
Version: commit 44ac8950df91dc91911d2cbea7512670923fe408
PoC Download: [Ov_compute_chosen_words.zip](https://github.com/opencog/link-grammar/files/1684283/Ov_compute_chosen_words.zip)

Steps to reproduce:
1.Download the .POC files.
2.Compile the source code with ASan.
3.Execute the following command
     : ./link-parser < $PoC

```
karas@karas:~/link-grammar/run/bin$ xxd Ov_compute_chosen_words 
00000000: 2847 033d 03                             (G.=.
```

```
=================================================================
==21046==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60300025aea4 at pc 0x000000518679 bp 0x7ffd0ca44640 sp 0x7ffd0ca44638
READ of size 1 at 0x60300025aea4 thread T0
    #0 0x518678 in compute_chosen_words /home/karas/link-grammar/link-grammar/linkage/linkage.c:480:22
    #1 0x518f5a in linkage_create /home/karas/link-grammar/link-grammar/linkage/linkage.c:681:2
    #2 0x4ea6f2 in process_some_linkages /home/karas/link-grammar/link-parser/link-parser.c:320:13
    #3 0x4ea6f2 in main /home/karas/link-grammar/link-parser/link-parser.c:917
    #4 0x7f47841f782f in __libc_start_main /build/glibc-Cl5G7W/glibc-2.23/csu/../csu/libc-start.c:291
    #5 0x441c08 in _start (/home/karas/link-grammar/run/bin/link-parser+0x441c08)

0x60300025aea4 is located 0 bytes to the right of 20-byte region [0x60300025ae90,0x60300025aea4)
allocated by thread T0 here:
    #0 0x4c8bd2 in __interceptor_malloc (/home/karas/link-grammar/run/bin/link-parser+0x4c8bd2)
    #1 0x56b0b2 in string_set_add /home/karas/link-grammar/link-grammar/string-set.c:156:17

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/karas/link-grammar/link-grammar/linkage/linkage.c:480 compute_chosen_words
Shadow bytes around the buggy address:
[...]

PR #666, opened 2018-02-01
Fix buffer overrun due to multiple SUBSCRIPT_MARK

ampli:
Per issue #665.
Test input: (G^C=^C
(^C is a SUBSCRIPT_MARK character 0x03.)

The code supposes the last subscript-mark in a stem belong to the stem-mark. This is not correct if the original word already contains a stem-mark that has a subscript-mark after it. Since the code assumes
there is "=\0" after a subscript-mark of a stem, a buffer overrun results.

To prevent this overrun, the fix here is to search for the first subscript mark of the stem. (The result of 
 such a nonsense word can still be nonsense.)

I inspected the other uses of strrchr(..., SUBSCRIPT_MARK), and fixed it in another place to strchr(), this time for consistency (another one in compute_chosen_words() and also one  in add_morpheme_unmarked(), seem harmless).

---
Proposals:
- Maybe we should use word marks access-methods  for manipulating SUBSCRIPT_MARK and other word marks.
- We may consider to filter-out control characters in the input, or maybe better, to convert them to ^X form.
In that case it will be fine to assume there is no more than one subscript-mark in a word, and it is a valid.

### DIFF
--- a/link-grammar/linkage/linkage.c
+++ b/link-grammar/linkage/linkage.c
@@ -469,7 +469,7 @@ void compute_chosen_words(Sentence sent, Linkage linkage, Parse_Options opts)
 							/* Cannot NULLify the word - we may have links to it. */
 							if (m != mcnt-1) chosen_words[i+m] = "";
 
-							sm =  strrchr(cdjp[i+m]->word_string, SUBSCRIPT_MARK);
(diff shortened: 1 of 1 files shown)


linas:
thanks

Issue #671, opened 2018-02-13
Actual dict order is not as described for UTF-8 characters


ampli:
In a comment in read-dict.c:
```
The use of suffixes means that the ordering of the words is not
 * exactly the order given by strcmp.  The order must be such that, for
 * example, "make" < "make.n" < "make-up" -- suffixed words come after
 * the bare words, but before any other other words with non-alphabetic
 * characters (such as the hyphen in "make-up", or possibly UTF8
 * characters). Thus, plain "strcmp" can't be used to determine
 * dictionary order.
```

However, since the comparison in `dict_order_strict()` is done using char, which is signed on x86, the above is not correct regarding `makeøup`, because the UTF-8 leading char turns out then to be negative.  Here is a dict order test I did with some fake words (using `link-parser '-\!a*'`):
```
    aaø.test 
    aa.u
    aab.test
```

**This doesn't look right. Are there any bad implications due to that?**

Note that the `dict_order_strict()` indeed takes care to make the subscript separator the lowest value character for comparisons, to ensure **make.n" < "make-up"**, from the time the subscript mark was `.`:
```
        return ((*s == SUBSCRIPT_MARK)?(1):(*s))  -  ((*t == SUBSCRIPT_MARK)?(1):(*t));
```

There are only very few such "mixed" in the dict, so this is not actually tested.
I could not create test "mixed" words that cause problems, but changing the above "return" to not preserve the **make.n" < "make-up"** causes dict errors.

BTW, I tried to enforce the description above on UTF-8 characters too by replacing the arguments of the `dict_order_*()` functions to be "signed char", in a hope that the overhead will be the same and this question will be redundant, but from some reason this generates much overhead.
(Also , compiling with `CFLAGS=-funsigned-char` results in ~50% slower runs.).

linas:
I believe (not quite sure) that the only word-ordering that matters is that the only ordering that matters is that the subscripted words must come after the non-subscripted words. I believe (not sure) that this is required for word-lookup to work correctly.  Thus, there would be bugs if this could happen:

```
aa
aa-foo
aa.u
```
or maybe 
```
aa.u
aaø
aa
```
Unclear about
```
aa.u
aa
```
The reason for this is that some sort of binary-tree lookup is used, and it halts when one gets too high or low -- thus a word, and word+subscripts need to be next to each-other.

ampli:
The check result that I posted indeed doesn't show the problem.
Here it is again, using the words from your post.

I added in the dict:
```
[...]

Issue #667, opened 2018-02-10
A dict word once with a subscript and once without


ampli:
From http://www.link.cs.cmu.edu/link/dict/introduction.html:
> If a word is listed more than once with the same subscript, or if it listed once with a subscript and once without, the parser will generate a warning message and will ignore one of the entries.

Also, in a comment in `tokenize.c`:
>         The dictionary reader does not allow you to have two words that
>         match according to the criterion below.  (so you can't have
>        "dog.n" and "dog")

Currently this is enforced (in `insert_list()`) **only** if the subscripted word appear before the unsubscripted one. This is because `abridged_lookup_list()` doesn't fetch the unsubscripted word is a subscripted one is given.

There are now several words in the dict that doesn't obey this rule.
For example, `a` appears before `a.eq`, and thus is not detected. (You commented there: `% XXX why doesn't this clash with a.eq ??`)

Other words with this kind of clash: . me lying if thy seriously

**Question**: What is the reason of this rule?
If it is needed, then the check (and the dict) should be fixed.
If it is unneeded, then the check should be removed, and the dict read will then become much faster (30% for **en**).

linas:
I cannot think of a good reason for this rule. I can think of a mediocre reason: it prevents the dictionary maintainer from accidentally creating multiple entries for the same word, and the resulting confusion from that situation.

For example, having two entries for "dog" might be confusing. By contrast having "dog.n" and "dog.v" makes it clear that one entry is meant to be a noun, the other a verb.  This makes it a bit more fool-proof for the dictionary maintainer.

I dunno. I kind-of don't care. Currently, I'm the only fool that needs to be fool-proofed - no one else is authoring these dicts.  It might be nice to have this check being optional, but its hard for me to imagine that I'd use this check very much.

linas:
I fixed some but not all of the trouble spots in #668 I couldn't get motivated to fix  ". me seriously"

linas:
I also removed the text "If a word is listed more than once with the same subscript, or if it listed once with a subscript and once without, the parser will generate a warning message and will ignore one of the entries."

ampli:
> I also removed the text "If a word is listed more than once with the same subscript, or if it listed once with a subscript and once without, the parser will generate a warning message and will ignore one of the entries."

So, should we keep checking this rule in every dict read?

[...]

PR #672, opened 2018-02-14
Dictionary read speedup

ampli:
See issue #667.

Main fixes:
- Avoid checking (by default), for each dict word, bare identical words.
  Instead, make this check at verbosity 10 (basic dict debug).
- Speed up `get_file_contents()`. Add read error message.
- Speed up dict_match() and is_number().
- Fix a bug in dict_error2() that prevented a further dict parsing (due to location lost).
- Fix error messages, and reword error messages to make them clearer.
- Fix a potential stack overflow when using !! with a wildcard.

In addition:
- Add general dict warning suppression, to be used (also) by a future dict checking program.
- Suppress warnings for know subscripted dict words that also have bare words.
- Add debug level 103 to show all these words without this suppression.
- Document the added check/verbosity level.
- Get rid of  BOM in `en/words/words.n.2.x`.
- Some code cleanup.
- Comment cleanup and reword.
- Jumbo typo fix of en dict comments.

Some speed up (maybe) can be done in dict_order_strict(), but I need to first fully understand the apparent problem I raised in issue #671.

### DIFF
--- a/data/en/4.0.dict
+++ b/data/en/4.0.dict
@@ -277,7 +277,7 @@ nonCAP.zzz: ZZZ-;
 
 % The use of COa here needs to be carefully re-examined; it is used much too freely.
 % COa+ is used to block links to COd-
-% Xc+ & Ic+: connect to imperatives (infinitve verbs): "Anyhow, don't"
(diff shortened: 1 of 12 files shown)


linas:
thanks!

Issue #632, opened 2018-01-11
Implementing connector length_limit per connector name


ampli:
I mostly finished the connector-enumeration modification, and it, as expected, maks a significant speedup. Currently it uses 32-byte Connector struct (instead of 64). 

Letter it may allow further reducing of the Connector struct (even 4 bytes is possible if desired - the connector descriptor number, bool:1 multi, and length_limit). It will also allow to reduce the size of the Table_connector struct, which its memory access is a major bottleneck.

Since I changed parts of the library to use connector descriptors instead of connectors (their ordinal number serves as a perfect hash), I also so change the expression pruning. In the same occasion I added a length_limit check (as you wrote [here](https://github.com/opencog/link-grammar/issues/250#issuecomment-165948913)). This significantly speeds up the power-pruning, which results in speed-up per (not trivially short) sentence.

Now I would like to add morphology-connectors length limits to solve problems like issue #537 (Russian sentences that are not very long take very much memory and CPU time). This may also solve some strange cases of linkages with null words, in which a stem of one word is connected to a suffix of another one.

(BTW I suppose that `-` and `+` connectors never need different length_limit, the original code had a partial implementation for that which I once removed).

There are several questions regarding it, among them:
1. Whether to initially implement length_limit weighting, and if so, how.
2. Which file format to use. It will be a good idea to use a flexible and extendable format, so it can be extended with easy backward compatibility (so old dicts can still be used with new libraries, for e.g. speed comparisons).
3. Where to implement length_limit caching . A caching is needed in order  not to parse and recompute the length_limit definitions again for each sentence, and not computing it twice (expression pruning and power pruning).

The previous discussion of defining length limit per connector type is [here](https://github.com/opencog/link-grammar/issues/537#issuecomment-333967987).

**1.** Whether to implement length_limit weighting, and if so, how.
You said there:
>Even if LL-style connectors were weighted, their length would probably still be restricted to 2 or 3, and so pruning would still mostly work. But maybe we should ignore wights until later; I'm not ready yet with complex morphology.

I didn't understand how you intend to make such weighting, and responded with a suggestion to make a length_limit table (per connector type) which depends on the `cost` option, e.g.
>```
>% For the connector PLx, the length limit depends on the cost, as follows:
>% length limit=1 for cost<=2.7
>% length_limit=3 for 2.7<cost<4
>% length_limit=5 for cost>4
>PLx  1 2.7   3 4   5: LENGTH_LIMIT+;
>```

[...]

PR #673, opened 2018-02-14
Connector enumeration + length-limit setup

ampli:
I could not find (in order to link to it here)  the issue in which I originally proposed the connector enumeration idea (you named this idea "enumeration" there).

For length-limit setup, see issue #632.

Main changes in this PR:
- Connector enumeration (significant speedup).
  While reading the dictionary, put the fixed properties of each connector in a central table, and
  precompute all properties which their value is now computed for each parsed sentence.
  In the connector structure, use a pointer to this table (called in the code the connector descriptor table).
  This also enables using the slot number of a connector in this table as a "perfect hash".
- Compute an encoded value for the lower-case part of connectors, for faster matching.
- Add a facility to define specific length limit for connectors.
- More efficient connector length-limit setup (also for UNLIMITED-CONNECTORS, and the 
  short_length/all_short_connectors parse options).
- Define length limit for LL* in the Russian dict (big speedup - especially for long sentences) and for 
 some connectors in the English dict.
- Add length-limit check to the expression pruning.
  I didn't check if it may improve the result of the power pruning, but it reduces its CPU time.
- Pack the connectors and disjuncts before parsing (notable speedup).
- Code infrastructure: Add connector accessors.

Some additional changes:
- Improve expression pruning debug.
- Add debug print for the parsing connector table at verbosity=102 (in a preparation for reimplementing it 
  \- a WIP).
- List all the connectors,  and their enumeration and length limits at verbosity=101.

Additional possible improvements using the new infrastructure:
- Improve post-processing speed. Deep changes are need to add an argument to many functions, unless 
  a **thread_local** pointer is added, to  point to the connector enumeration table.
- Improve the speed of the fast-matcher by sorting the connector lists according to upper-case strings (to 
  improve the match cache hit ratio).
- Reduce the size of the Connector struct (I intend to try even 4 bytes).
- Reduce the size of a Table_connector element in count.c (made possible by packing the connectors - the connector index can  used instead of its address).

### DIFF
--- a/link-grammar/connectors.c
+++ b/link-grammar/connectors.c
@@ -62,7 +62,6 @@ Connector * connector_new(const condesc_t *desc, Parse_Options opts)
 	c->desc = desc;
 	c->nearest_word = 0;
 	c->multi = false;
-	c->tableNext = NULL;
 	set_connector_length_limit(c, opts);
 	//assert(0 != c->length_limit, "Connector_new(): Zero length_limit");
 
--- a/link-grammar/connectors.c
+++ b/link-grammar/connectors.c
@@ -17,7 +17,9 @@
 #include <math.h>
 
 #include "dict-common/dict-utils.h" // for size_of_expression()
+#include "api-structures.h"         // for Parse_Options_s
 #include "connectors.h"
+#include "link-includes.h"          // for Parse_Options
 
 /**
  * free_connectors() -- free the list of connectors pointed to by e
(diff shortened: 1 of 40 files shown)


link-grammar/connectors.c
```diff
@@ -33,116 +35,221 @@ void free_connectors(Connector *e)
...
+		if (NULL != conlist)
+		{
+			conlist[*cl_size] = e->u.condesc;
+		}
+		(*cl_size)++;
 	} else {
 		for (l=e->u.l; l!=NULL; l=l->next) {
-			build_connector_set_from_expression(conset, l->e);
+			get_connectors_from_expression(conlist, cl_size, l->e);
 		}
```
linas: I think it would be slightly faster to return cl_size as a return value, rather than passing a pointer.  When you increment a value at a location, the compiler has to assume a worst-case of possible aliasing, and constantly write it out to memory (cache). By contrast, a return value will usually stay in registers, maybe be written to stack occasionally.

link-grammar/connectors.c
```diff
@@ -33,116 +35,221 @@ void free_connectors(Connector *e)
...
+			printf("%5zu %6d %3d %s\n", n, ct->sdesc[n]->uc_num,
+			       ct->sdesc[n]->length_limit, ct->sdesc[n]->string);
+		}
+	}
+
 }
 
 /* ======================================================== */
 
+static bool connector_encode_lc(const char *lc_string, condesc_t *desc)
```
linas: Can you provide a description of what this does?  It seems to be taking the lower-case part of a connector, normally an 8-bit quantity, and packing it into a 7-bit thing .. that seems like a lot of work, for not much savings.  Since it is lower-case, you could, in principle, pack down to 5 bits, for more savings ... but is it really worth the effort? or am I missing something?

linas: OK, never mind. I see what you are doing .. you have to copy the LC chars into the u64 anyway, and you have to compute the mask, anyway, and, so at almost no extra cost, you can shrink to 7 bits.  But then, at almost no extra cost, you can subtract 0x60 and pack in 5 bits. On the other hand, I do not believe we have have more than maybe 5 or 6 lower-case letters... so this is not all that necessary...

link-grammar/connectors.h
```diff
@@ -27,73 +30,122 @@
...
+		connector_hash_size uc_num;
+	};
+	uint8_t length_limit;
+	                      /* If not 0, it gives the limit of the length of the
+	                       * link that can be used on this connector type. The
+	                       * value UNLIMITED_LEN specifies no limit.
+	                       * If 0, short_length (a Parse_Option) is used. If
+	                       * all_short==true (a Parse_Option), length_limit
+	                       * is clipped to short_length. */
+	char head_depended;   /* 'h' for head, 'd' for depended, or '\0' if none */
```
linas: typoe: "dependent", not "depended"

link-grammar/connectors.h
```diff
@@ -198,4 +309,92 @@ static inline unsigned int pair_hash(unsigned int table_size,
...
+	ct->hdesc = (condesc_t **)malloc(size * sizeof(condesc_t *));
+	memset(ct->hdesc, 0, size * sizeof(condesc_t *));
+	ct->size = size;
+}
+
+static inline bool condesc_insert(ConTable *ct, condesc_t **h,
+                                  const char *constring, int hash)
+{
+	*h = (condesc_t *)malloc(sizeof(condesc_t));
+	memset(*h, 0, sizeof(condesc_t));
```
linas: I'm sort-of surprised by this malloc. Why not just change `condesc_table_alloc` so that it mallocs an array of `condesc_t` instead of an array of pointers to `condesc_t` ?  That way, you can skip this smaller alloc entirely.

link-grammar/dict-file/read-dict.c
```diff
@@ -947,14 +954,16 @@ static Exp * make_connector(Dictionary dict)
...
 			min = make_dir_connector(dict, i);
+			if (NULL == min) return NULL;
 
 			n = make_or_node(&dict->exp_list, plu, min);
 		}
 		else
 		{
-			dict_error(dict, "Unknown connector direction type.");
+			dict_error(dict, "Unknown connector direction type '%c'.");
 			return NULL;
```
linas: The connector direction is never iven.

link-grammar/parse/parse.c
```diff
@@ -247,6 +246,55 @@ static void sort_linkages(Sentence sent, Parse_Options opts)
...
+		{
+			dcnt++;
+			for (Connector *c = d->right; c!=NULL; c = c->next) ccnt++;
+			for (Connector *c = d->left; c != NULL; c = c->next) ccnt++;
+		}
+	}
+
+#define CACHELINE 64
+	size_t dsize = dcnt * sizeof(Disjunct);
+	dsize = (dsize+CACHELINE)&~(CACHELINE-1); /* Align connector block. */
```
linas: Here, you align, but then below, you add csize of unknown alignment.   I'm sort-of confused, because you then use what malloc provided, without actually aligning -- malloc might provide an address that is not 64-byte aligned.

link-grammar/connectors.c
ampli: > But then, at almost no extra cost, you can subtract 0x60 and pack in 5 bits. On the other hand, I do not believe we have have more than maybe 5 or 6 lower-case letters... so this is not all that necessary...

The LC part is not really only lower case letters. It may contain also numbers. So we have 36 possible characters, that can be packed into 6 bits. However, this is more complex than just removing the most significant bit since there is a gape between the values if the lower case letters and the numbers.

Not doing this packing means maximum 8 characters in the LC part, and I agree 9 is not a big improvement and may not even worth the slight overhead of doing it once. The reason I included it is copying from an old code I wrote, that included packing the UC part in 6 bits (the same packing code can be used). Packing the UC part is not needed now because the connector enumeration idea is better.

So I will remove this packing (or ifdef it out).

Also note that all[*] the connector related computations, including LC packing, are now done only once - at the dictionary read time, instead of per sentence (or even per connector) as before.
[*] Besides short_length/all_short_connectors, which can also be computed and cached only when changed, but I didn't implement it.

link-grammar/connectors.c
ampli: You are right. I will fix that.

link-grammar/connectors.h
ampli: I will fix it,

link-grammar/connectors.h
ampli: I don't know in advance how many different connectors are in the dict, so this hash table can grow. I didn't want to pre-allocate a an array that cannot grow.

However, I now see it can be done in another,  more efficient, way:
Make a big malloc for the initial guessed number of connectors, as set in dictionary_six_str().
If exceeded, realloc double space. Use it as a `condesc_t` pool. Optionally, at the end, realloc to the exact used amount, in order to free the extra space. Another possibility is allocate by groups of `condesc_t` elements, say 256 and in that way absolutely most of the malloc overhead is saved.

The above realloc way is simpler.  I can implement it.

link-grammar/dict-file/read-dict.c
ampli: Indeed.
Since the second argument of dict_error2() is never used, I can change this call to dict_error2(), and change dict_error2() to print an additional value nicely.

A better alternative can be to change dict_error() to a printf-like function.

link-grammar/parse/parse.c
ampli: The memory arrangement in this allocation is:
DISJUNCT_ARRAY **ALIGNEMNT_PAD** CONNECTOR_ARRAY
The Connector struct is 32 bytes (and can be shrinked to 16, 8 and even 4 bytes - I would like to test all).
In order not to split connectors over two cache lines, I want CONNECTOR_ARRAY to start at 64-byte boundary. I get this here by adding an alignment pad at the **end** of the disjunct allocation.
(As far as I remember the Disjunct struc size is something like 44 bytes so there is no point to align its start and it doesn't usually end at the desired boundary.)

So it seems fine to me.
If desired, I can add a description like the above to the function comment.

ampli: > As far as I remember the Disjunct struc size is something like 44

(It is actually 56.)

link-grammar/connectors.c
linas: OK. Maybe its not that important.

link-grammar/connectors.h
linas: sorry, I did not realize it was a hash table.  It seemed to be getting used as a sequential list.  I suppose I misread the code

link-grammar/parse/parse.c
linas: OK. If the code likes right to you, then OK, it looked strange to me but I did not try very hard to figure it uout.

link-grammar/connectors.h
linas: Um, hang on. `condesc_t` doesn't have a `next` pointer in it, so it cannot be chained, so it cannot be used in a hash table, unless you can guarantee zero collisions..... A hash table with zero-or-one items in each bucket ... Hmmm.  Are you using some trick to spill to the next empty bucket? There was some old code that did that.

link-grammar/parse/parse.c
ampli: I hope it will be clearer after I add comments to it.

link-grammar/dict-file/read-dict.c
ampli: > Um, hang on. condesc_t doesn't have a next pointer in it, so it cannot be chained, so it cannot be used in a hash table, unless you can guarantee zero collisions..... A hash table with zero-or-one items in each bucket ... Hmmm. Are you using some trick to spill to the next empty bucket? There was some old code that did that.

It is similar to the string_set code, in that it uses empty bucket (but in the case of my code I use a constant stride of 1, since it is more cache friendly, see below).  It is also how it is done in the fast-matcher connector lists per word.

This can be improved by using Robin Hood placement (that I use in a WIP to replace the memoizing table of do_count()), and intend to try it here too.

The problem of using a next pointer for chaining is that it is very cache unfriendly, due to random memory access (the chained element has a very low chance to be in a close memory location). On the other hand, searching in an "in place" hash table accesses sequentially a relatively small memory region (very few elements are needed to be inspected by average and the worst case is small, if the table is kept, e.g, under 50% used), and due to the sequential access the CPUs knows to prefetch the next memory to be accessed.

link-grammar/connectors.h
linas: OK, so you are saying that condesc_t is large enough that it is better to have an array of pointers to them, rather than an array of them?

ampli: Drawbacks of not using array of pointers:
1. The size of condesc_t is 32, so a table of say 20K items (i think this what it may end up for "ru") will have a significant memory overhead. Since it is too large for an L2 CPU cache, searching in it may cause several times more CPU cache misses (with 8 pointers per cache line you may normally have no more than 3 misses).
2. The sorting, which is essential here, is more costly (need to swap the elements content during sort).

However, I didn't made a benchmark that compares the different ways of doing it.

There is an additional way of doing it that maybe worth exploring: Use the dict creating code.
The idea is just to put the connector strings into their own dict, and enumerate them by listing this dict words (similarly to what `!!*`command does).

linas: OK, that's OK. You can leave this alone; I was just reading through the code, got a little confused, and needed to ask a few questions. but this seems OK as is,  No change needed.

On a different note, the `tests/memleak` unit test crashes with a memory double-free. That's important.

link-grammar/connectors.c
ampli: > But then, at almost no extra cost, you can subtract 0x60 and pack in 5 bits. On the other hand, I do not believe we have have more than maybe 5 or 6 lower-case letters... so this is not all that necessary...

With such a sustruction method (e.g substructing 64 and taking the the absolute) it is possible to pack in 6 bits instead of 7 (I was wrong in saying this is more complex due to the gape), and also the longest LC part is 5 - the version (and domain) encoding. So with this packing 32-bits can hold 5 characters. However, this will prevent versions like `V5v4v10`...

BTW, I once hacked the dict reading and added `#define` for dict definitions (a very small change) (`#` is already a special character for the dict).

Another BTW, for not opening an issue for that, I noted that reading the `id` dict fails
 with: `Connectors beginning with "ID" are forbidden`. The easiest fix for that is to forbid only connectors in the format IDX when X is a  capital letter string (as only that may clash with idiom connectors).

Is such a fix is fine?

linas: ID: apparently, its needed, so that we can set the indonesian locale. So yes.  

We do need version strings like `V5v4v10` ...

Note that the statistically-generated dictionaries currently use only upper-case letters, and have zero lower-case letters in them. And there are a lot of them:  I haven't counted recently, but maybe hundreds of thousands (or more). The goal of learning is to shrink this number, but ...

PR #676, opened 2018-02-17
 Fix a crash of double-free of sentence disjuncts

ampli:
See issue [#632](https://github.com/opencog/link-grammar/issues/632#issuecomment-366310669).

At the same opportunity:
- Refactor the touched code.
- Remove unneeded include files from the touched files.
- Avoid trying to free disjuncts from a fresh sentence.


linas:
Thanks!

PR #681, opened 2018-02-21
do_count() minor changes

ampli:
This branch is called "do_count-speedup" but the speedup here (saving multiplications) is minor.
(The great speedup (hopefully) for which I started this branch is still a WIP.)
However, I am sending the current changes in this branch in order not to accumulate too many changes over too much time (for some old changes it becomes very hard to merge/rebase due to other PRs that are sent meanwhile, and I have tens of WIP branches...).

The major change here is the improved do_count() trace, that gives hints on a possible speedup improvement.


linas:
Yes, smaller, shorter changes are also easier to review

PR #685, opened 2018-03-03
Changes for MSYS2, Russian speedup, and more

ampli:
In order to test the memory pool changes, I also liked to test them under MinGW.
However, I encountered configuration and compilation problems (not related to the new code), some of them are new but others always existed.

The fixes also introduced general improvements.
Here is a list:
- Detect `libtre` and prefer it for regex.  It was needed because MinGW  doesn't
support backslash escapes like `\w` (because it uses the BSD regex library). This fixes the same problem for Cygwin.
As a side effect, this change speeds the Russian batch by more than 10%.
- Use `regexc` option `REG_NOSUB`, since capture references are not currently used (if needed, and option like  `//s` can be easily added). This has a slight speed benefit for Russian.
- Update the tools prerequisites in the main README.
- `configure.ac`: Use `AX_JNI_INCLUDE_DIR`  for detecting `jni.h`.
- `configure.ac`:  Remove unneeded `AC_PROG_LIBTOOL`.
- `configure.ac`:  Check `$host_os` instead of `$host` and set `OS_WIN32` for all the WIN32 POSIX environment emulations.
- Adapt Java handling for Windows (several changes) to enable automatic bindings.
- Update the MinGW documentation (new `README.MSYS2`. The old one was retained, but can be removed if desired.
- Remove the performance section from `README.md`.
- Fix python3 uninstall problem (it tried to rmdir system directories instead of the installed ones).
- Fix compiling  `jni-client.c` with `--enable-debug`.
- Add a workaround for a MinGW/MSYS2 problem that prevents wrapper compilation.

In a preparation to the memory pool changes:
- Change the CFLAGS on WIN32 to  `-std=c11`  and add `-D_ISOC11_SOURCE`.
- Check for an aligned malloc.

EDIT: Pushed again to:
- Fix a commit.
- Add commits for README renaming and README reference markdown fixes, and typo fixing.


linas:
thanks!

PR #697, opened 2018-03-11
Fix the optimized pseudocounting code for PERFORM_COUNT_HISTOGRAMMING

ampli:
Commit 3cf78a7b for optimizing do_count() broke the histogramming code.
This PR fixes that.


linas:
The histogramming code was intended as a way of finding parser optimizations, and learning about what it is doing....

ampli:
Are there any insights from it?

linas:
It was years ago. I got some insights, but not as clean-cut as I'd hoped. I wanted to do one particular  optimization, and was going to use this as evidence for it, but the evidence was cloudy.

I understand now that one needs to average over thousands of sentences to get a clear picture, but building technology that can do that is non-trivial.

PR #701, opened 2018-03-14
Fix memleak of length-limit definitions on bad dict + handle ID4id+

ampli:


ampli:
Note that the `id` dict cannot work because every work there is defined multiple times instead of using OR.
Propose fix: Convert to use OR.

Also note that the `vn` dict cannot work due to connector names with a Unicode character Đ and an underbar.
Proposed fix: Convert `Đ` to `DD` and omit the underbars.

linas:
Yeah, I keep writing to the guy in Vietnam but he doesn't answer. The indonesian dict is also proof-of-concept. Yes, a fix would be nice.

PR #702, opened 2018-03-14
Fix dict syntax of vn & id

ampli:
(Ref. PR #701.)
For id, the solution I finally implemented is different, in order to preserve the info in the dict.
I supposed the duplicate words are different usages of the word, so I just subscripted them with a number.
(Regretfully, there is no corpus to make checks.)

For vn, I just converted the invalid connector characters to something valid.
I noted that it parses "fully" only one sentence (bside a word that has an empty expression in the dict for some reason).


ampli:
You may also want to upgrade the dict versions.

PR #704, opened 2018-03-17
Fix errors detected by Coverity

ampli:
Regarding the action on bad UTF-8: For now it is silently "bypassed" or ignored.
Maybe it will be a good idea to add an error message in DEBUG mode, or check the UTF-8 input and add an "internal error" message on every bad UTF-8 afterwards.

### DIFF
--- a/link-grammar/dict-common/dict-api.h
+++ b/link-grammar/dict-common/dict-api.h
@@ -25,7 +25,9 @@ LINK_BEGIN_DECLS
  * the public API to the link-parser system.
  */
 
+#ifdef USE_VITERBI
 Dictionary dictionary_create_from_utf8(const char * input);
+#endif
 
 bool boolean_dictionary_lookup(const Dictionary, const char *);
 
--- a/link-grammar/dict-file/dictionary.c
+++ b/link-grammar/dict-file/dictionary.c
@@ -307,10 +307,15 @@ Dictionary dictionary_create_from_file(const char * lang)
 }
 
 
+#ifdef USE_VITERBI
 /**
  * Use "string" as the input dictionary. All of the other parts,
  * including post-processing, affix table, etc, are NULL.
  * This routine is intended for unit-testing ONLY.
+ *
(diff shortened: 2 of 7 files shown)


link-grammar/dict-common/dict-api.h
```diff
@@ -25,7 +25,9 @@ LINK_BEGIN_DECLS
  * the public API to the link-parser system.
  */
 
+#ifdef USE_VITERBI
 Dictionary dictionary_create_from_utf8(const char * input);
+#endif
```
linas: Some variant of this should almost certainly be in the public API.  Viz, a way to provide dictionary contents, without requiring files.  Not clear how affix and regex files could be handled, though... maybe more string routines.

ampli: Maybe:
`Dictionary dictionary_create_from_string(const char *dict, const char *affix, const char *regex, const char *knowledge , const char *constituent_knowledge)`
when the one that are not used can be NULL.
The returned value should probably not be `const`, so API for manipulated the dict can be later added.

linas: Yeah, seems reasonable.   C still doesn't have default argument values, like C++, right?

ampli: > C still doesn't have default argument values, like C++, right?

No such thing yet in C, but maybe it can be emulated by vararg or sentinel values, like:
``` c
# The rest assumed to be in files
dictionary_create_from_string(dictstr, NULL);
dictionary_create_from_string(dictstr, , affixstr, NULL);
# DICT_FROM_FILE means read from standard location
dictionary_create_from_string(dictstr, DICT_FROM_FILE, NULL, DICT_FROM_FILE, DICT_FROM_FILE);
```

Most probably the language should also be added - as first argument.

It can even be made more generic - to read from any source,.e.g. using URL-like prefixes like `file:` or `sqlite:` (default is string), and an optional additional argument can be a search path. In that case the function name can be just `dictionary_create()`.

PR #705, opened 2018-03-18
Fix memory pool leak in prasing with null links

ampli:
Somehow I neglected to check that.
This leak happen when parsing with null links, and there is no full linkage.
Just running the batches doesn't trigger it.

**I strongly recommend to make a new release soon.**
However, before that I would like to add a test of 2-step parse, see below.

---

The test suit doesn't include a 2-step parsing with null links, like `link-parser` does.
Such a test would reveal this leak (because I run it with ASAN/UBSAN).
The only parse with nulls it includes is 1-step.

However, a 2-step test cannot be added just now because the Python binding of `parse()` doesn't include arguments, so a new `Parse_Options` cannot be used (the `Parse_Options` of `Sentence()` is reused).
This is a deficiency of the Python bindings, because the ability to change the `Parse_Options` without  the overhead of calling `Sentence()` or `sent.Split()` again is desired (for example to change `short_length`).

(BTW, Linkage() already includes `Pasre_Options`. A test for using it to change the `morphology` option should be most probably eventually added too.)

To sum up:
1. I will send a PR for an optional `Parse_options` argument of `parse() and test for a 2-step parse with nulls.
2. After that it will be a good idea to release a new version.

### DIFF
--- a/bindings/python-examples/lg_testutils.py
+++ b/bindings/python-examples/lg_testutils.py
@@ -68,9 +68,11 @@ def add_eqcost_linkage_order(original_class):
 
     original_class.original_parse = original_class.parse
 
-    def parse(self):
+    def parse(self, parse_options=None):
         """A decoration for the original Sentence.parse"""
-        linkages = self.original_parse()
+        # parse() has an optional single argument for parse options. If it is not given,
+        # call original_parse() also without arguments in order to test it that way.
+        linkages = self.original_parse() if parse_options is None else self.original_parse(parse_options)
         return eqcost_soretd_parse(linkages)
 
     original_class.parse = parse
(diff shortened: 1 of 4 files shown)


ampli:
It turned out the current test suit also shows the memory leak , but my LSAN suppression `leak:*python2.7*` suppressed too much (will need to check how not to suppress LG library leaks too).
In any case:
1. I added a flag to my test script to avoid this suppression, to be used for memory leak checks before I send PRs...
2. In any case I added an optional Paerse_Option to parse() and added a test of a 2-step parse.

I pushed again the PR so now it includes these fixes too.
In addition to running the tests, including the Python ones, I also produced detailed linkages for all the languages, all of that with ASAN.

So now hopefully there is no so trivial leakage, **and a new version can be released**.

linas:
OK. When you do this, update the Changelog too, so that we know why there is a new release.

ampli:
Submitted the ChangeLog, but please wait to the next PR (will send in a few hours), as I ran Coverity again and have some more fixes.

PR #706, opened 2018-03-18
Clang static analyzer round 2

ampli:


ampli:
Need to fix it...

ampli:
Fixed.

PR #707, opened 2018-03-19
Update the change log (fix memleak & Python API)

ampli:
Changelog for PR #705


ampli:
It seems I deleted the branch before this has been applied.

linas:
Yeah, I was wondering ....  So, are things ready to go, from your end?  I need to fix bug #710 and then want to publish.

ampli:
> So, are things ready to go, from your end?

Yes.

PR #709, opened 2018-03-19
Cleanup8

ampli:
Misc. cleanup.

### DIFF
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -86,9 +86,10 @@ static void init_table(count_context_t *ctxt, size_t sent_len)
 
 }
 
-#ifdef DEBUG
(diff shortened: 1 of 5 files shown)


link-grammar/parse/count.c
```diff
@@ -86,7 +86,7 @@ static void init_table(count_context_t *ctxt, size_t sent_len)
 
 }
 
-#ifdef DEBUG
+#if 1// def DEBUG
 static int hit;
```
linas: not `#if 1`

ampli: In a try to prevent such things I will change my bad practice of "temporary" enabling DEBUG sections (when I don't want to use --enable-debug due to check slowness). Instead,  in cases it is needed I will add another macro definitions for that (to be used when configuring or compiling), like what I did in the fix I have just force-pushed.

Issue #710, opened 2018-03-22
missing condesc_t in dict-structures.h


mjsduncan:
this is on a fresh install of ubuntu 17.10 and link-grammar 5.4.4 tarball
```
[ 49%] Building CXX object opencog/nlp/lg-dict/CMakeFiles/lg-dict-entry.dir/LGDictExpContainer.cc.o
In file included from /usr/local/include/link-grammar/dict-api.h:17:0,
                 from /home/mozi/oc/opencog/opencog/nlp/lg-dict/LGDictExpContainer.h:27,
                 from /home/mozi/oc/opencog/opencog/nlp/lg-dict/LGDictExpContainer.cc:27:
/usr/local/include/link-grammar/dict-structures.h:18:10: fatal error: connectors.h: No such file or directory
 #include "connectors.h"
          ^~~~~~~~~~~~~~
```

ampli:
I added it so editing `dict-structures.h` with an IDE will not flag errors due to the recently added `condesc_t` struct member of Exp_struct. But when used from `LGDictExpContainer.h`, it is not found since it is not installed in the system.

There is a worse problem:
In `LGDictExpContainer.cc`, `exp->u.string` is used. But now the Exp struct doesn't contain this field.

@linas, please advise...

ampli:
As a work around, try to use link-grammar 5.4.3 instead of 5.4.4.

linas:
I will fix `exp->u.string`  ...   mostly just have to move `conndesc_t` to `dict-structures.h`

ampli:
Instead, I can implement #577.
I will continue there.

linas:
Amir -- both the pull reqs #711 and #715 provide fixes for this bug. Can you look and see which is better? Or if there is a third way?

I don't like #711 because it puts "private" things into the public API.
For this reason, #715 was supposed to be better, except that now there is a string pointer in both `struct Conn` and in `condesc_t`, which is kind-of-ugly.  

Is there a better way?

ampli:
Readding the string to the connector struct doesn't seem "right", due to duplicate information
and making the struct bigger (on the other hand it makes the access to the string faster).

I will try to propose another way, but I need to understand the following:

What is wrong with adding a public API `const char* get_expression_string(Exp*)`, leaving `condesc*` as an opaque pointer?
[...]

Issue #577, opened 2017-08-12
dictionary_lookup_list does not look up regexes!


linas:
The `dictionary_lookup_list()`  was added to the public dictionary API, so that other users (specifically, the sureal (surface realization) and microplanning modules for sentence-generation) could look up words in the LG dictionary. And that's mostly fine, except that it does not lookup regexes.  So:

-- the LG internals need to be jiggered around, so that the public API does lookup regexes.

A related issue is that the public API exposes the Exp structure, which has a bizarre design. It needs to be reworked so that it's cleaner, nicer for the ordinary user. Unfortunately, this is a lot of work.

ampli:
Do you mean something like that (and the same for `db_lookup_list()`):
``` c
Dict_node * file_lookup_list(const Dictionary dict, const char *s)
{
	Dict_node * llist =
		rdictionary_lookup(NULL, dict->root, s, true, dict_order_bare);
	llist = prune_lookup_list(llist, s);
	if (NULL != llist) return llist;

	const char *regex_name = match_regex(dict->regex_root, s);
	if (regex_name) return file_lookup_list(dict, regex_name);

	return NULL;
}
```

Regarding expression format, what would be considered a nice format?
Is ASCII representation like `expression_stringify()` better than the current C structure?

linas:
> Do you mean something like that

Yes.

> stringify()

Yes, that would probably be best.  Returning a string that resembles the current ascii dictionary format would be best.  I don't recall exactly what `expression_stringify()` prints.

ampli:
> I don't recall exactly what expression_stringify() prints.

It prints the expression in the current dictionary format.
So the question now is how to implement it.

I guess we will need a new API.
The easiest way is seems to have 2 API  functions, something like:
`dictionary_lookup_words()` # Return list of words
`dictionary_lookup_exp()` # Return list of corresponding expressions
Or maybe we can have one function `dictionary_lookup()` that returns a list word1, exp1, word2, exp2, ... etc...
One question is whether we need to add a third component that indicates if the word has been resolved through a regex.

Or maybe we can use a JSON format, which can be extendable in a compatible way for any future need, and API users can use a JSON library function to decode it, if desired. We can use such a JSON API as a future model for some additional APIs we still need to add.

linas:
> JSON format

Yes, I like that best!  Some of the current API's could/should be provided in json.

ampli:
Proposal:
```
[...]

PR #711, opened 2018-03-22
Fix for bug #710

linas:
Move condesc structures to "public" API .... 

This is not a good solution.... it moves too much internal stuff into an external header file.

### DIFF
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -20,6 +20,7 @@
 #include <stdint.h>  // for uint8_t
 
 #include "api-types.h"
+#include "dict-common/dict-structures.h"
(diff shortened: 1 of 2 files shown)


ampli:
I propose two alternatives:
1. Implementing `linkgrammar_get_dict_word()` (issue #577), and use it in `LGDictExpContainer.cc` .
But this may take too much time for the next version, that needs to be issued ASAP.
2. As an immediate work around, define the following (maybe temporary) API:
const char* get_expression_string(Exp*)
char get_expression_dir(Exp*)
bool get_expression_multi(Exp*)
This way there is even no need for the current `Exp_struct` exposure.

If (2) is fine I can send a PR.

PR #722, opened 2018-04-01
Use "bool" instead of "int" for functions returning T/F

ampli:


link-grammar/prepare/exprune.c
```diff
@@ -160,25 +160,25 @@ static E_list * or_purge_E_list(E_list * l)
...
-	if (l == NULL) return 1;
+	if (l == NULL) return true;
 	if ((l->e = purge_Exp(l->e)) == NULL)
 	{
 		free_E_list(l->next);
 		xfree((char *)l, sizeof(E_list));
-		return 0;
+		return false;
 	}
 	if (and_purge_E_list(l->next) == 0)
```
linas: 0 should be `false`

Issue #729, opened 2018-04-06
A proposal for link-parser -help and !help


ampli:
On some issue (I could not locate it) I talked about adding a linkage output legend.
Here is a proposal:
```
Output legend (!graphics=1):

 [word]            Null-linked word
 word[!]           Word classified by a regex
 word[!REGEX_NAME] Word classified by REGEX_NAME (turn on by !morphology=1)
 word[~]           Word generated by a spell guess (unknown original word)
 word[&]           Word run-on separated by a spell guess
 word[?]           Word is unknown (looked up in the dict as UNKNOWN-WORD)
 word.POS       Word found in the dictionary as word.POS
 word.#CORRECTION  Word is probably a typo - got linked as CORRECTION

For dictionaries that support morphology (turn on by !morphology=1):
 word=             A prefix morpheme
 =word             A suffix morpheme
 word.=            A stem

For more details see:
	http://www.abisource.com/projects/link-grammar/dict/introduction.html
```
I also noted the following current problem:
`link-parser -help` does not show the command line usage (it is shown only when a bad argument is used, e.g. `link-parser -x`. Instead it shows the internal help only.

**Proposal** (according to [GNU Coding Standards 4.7.2](https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dhelp)):
  - Do not show the internal help (show it only on the internal `!help`).
  - Show the command line usage on `-help`.
  - Show package web links.
  - Then **exit**.

On an incorrect command line usage as `link-parser -x`, I propose to mention `-help` and not `!help`.

The proposed output legend can be shown on !help.
However, a better approach can be to add arguments to `!help`, and show it on `help graphics`. This way help texts can be added for other commands too. In principle, it is possible to extract them from a file (e.g. `data/help.txt`) that will be installed in `PREFIX/share/link-grammar`.
(It can be a major change to allow `!help` arguments, but if we use the fact that we don't have commands starting with `h`, then it becomes a trivial change.)

EDIT:
From the `link-parser(1)'  manual (see the <--- marked line):
```
OPTIONS
       --help Print usage and exit.                  <---

       --version
              Print version number and exit.
```
Somehow my proposed action (as opposed to the current behaviour) is already documented.

ampli:
> In principle, it is possible to extract them from a file (e.g. data/help.txt) that will be installed in PREFIX/share/link-grammar.

[...]

PR #738, opened 2018-04-15
Link-parser: Add per command help and command completion

ampli:
The per-command help implements issue #729.
The "!" command line completion implements a FIXME I added when I implemented file completion for `!file.`.

This PR contains also some non-related fixes in the thouched code:

- link-parser: Avoid duplicate history entries
This can also be done in `.inputrc`, but since they are not useful in `link-parser` I added that in the code.
- link-parser: Remember in the command history 100 events
Int can be increased in `inputec`, but the original 20 seems to be too low.
- Handle missing !file argument
Currently the code tries to open an empty-string file name.
- link-parser --help: Print the command-line usage and exit 0
I proposed in #729 but didn't implement (may implement later):
- List Web reference
- In the help message, refer variable setting using '-' and not '!' (since it is a help of the command line usage).
x_issue_special_command(): Fix checking float assignment
- Clarify the path_found comment
- command-line.h: Remove MSVC6 support
- Several source code reorganizations.

**Related changes:**
- Add API linkgrammar_get_data_dir()
- link-parser: Add per-command help
 -- The help text starts by displaying the current and default values (for variables).
 -- I included a preliminary help file `command-help-en.txt` with detailed help on most of the commands.
I need help in making it better...
For example I used `!varname=True` to denote assigning to a Boolean variable, but the `!variable` command uses the terms On and Off (which cannot be actually assigned).
 -- The new code contains searching for the language part of this filename using `LANGUAGE` and `LANG`.  
I didn't find a reference code so I wrote it according to the specifications.
- The `!help` output now document `!exit` (I preferred it over `!quit`, which is also recognized).
-  link-parser: Use UNDOC for clusters, cost-model, senses
This is a source code infrastructure features I added for command which are not listed or being completed. `!quit` is also included. I later added "memory".  The idea was not to clutter the user interface with unneeded command. Maybe all of them (but `!quit`) should be ifdef`ed out instead.
- Support shortening of the `!file`, `!exit` and `!quit` commands (by adding them to the command table).
- Add "!help var" to !variables
Now that I write this I think maybe it should have been `!help <var>` or `!help varname`.
- lg_readline.c: Add command/variable completion
 -- It handles only lines that start with `!`.
 -- There is some code bloat for handling optional whitespace after the initial  '!' (to be compatible to the current code).
 -- Support also completing `!help` argument.
 -- In case muti-completions are possible, instead of just list the choice, show 1-line help on the possible completions.
 -- I added printing the command /variable type. In retrospect maybe this is unneeded.
 -- Even though the existing command decoding code supports case-ignore on commands, the completion code uses the commands as they are (lowercase), as I found it unneeded to support uppercase in what must be an interactive usage (in files uppercase can still be used).

I could not add a `ChangeLog` since it will cause a conflict with the pending PR #737, which changes `ChangeLog` too.

I can send additional PR after these PRs are applied.
Anyway, here is the ChangeLog:
```
* Add a per-command help in link-parser.
* Add a command line completion in link-parser.
```


link-grammar/link-includes.h
```diff
@@ -113,6 +113,8 @@ link_public_api(void)
      dictionary_set_data_dir(const char * path);
 link_public_api(char *)
      dictionary_get_data_dir(void);
+link_public_api(char *)
+     linkgrammar_get_data_dir(void);
```
linas: Hmm. I merged this pull req, but I'm thinking that this routine should probably be merged with `dictionary_get_data_dir()` -- that is, once the location of the dictionary has been determined, that is what `dictionary_get_data_dir` should return.

ampli:
> 1. Add a new dictionary_data_dir(Dictionary *).

OK, so it will return the data dir that will be kept in the Dictionary struct.

> 2. Maybe remove dictiionary_get_data_dir()?

I agree that it is not too useful since it returns what `dictionary_set_data_dir()` has set, and the programmer can memoize that and use it later if needed, but I don't see any harm to leave it as is...
(It still can be useful in cases that I can think of.)

> 3.  Rename command-help-en.txt to command-help.txt and install it in the root data directory.

This is what I proposed in [this post](https://github.com/opencog/link-grammar/issues/729#issuecomment-379815448), but I added in its end "EDIT: Not really - see below".

> 4. When searching for command-help.txt, first look for it in the dict locale, and use that if it exists; if it does not exist, then use the one in the root data directory (i.e. the English one)

There are several problems with that, here are the main 2 ones:
1. Suppose one works on the Russian dict (or just checks it). It is still reasonable and probable she/he may like to read the help in English (or in the language she/he have set in LANGUAGE/LANG, if such an help available) and not in Russian (supposing a Russian file is available). But it can be solved by still looking in LANGUAGE/LANG.
2. **The main problem** that I mentioned and the way you suggest doesn't address:
Let's look at the regular case in which the user has an LG package installed, and she/he develops a language in her/his own directory, say in `~/user/ru`.
This is a usual and supported (and even the recommended) way to work.
Say LANGUAGE/LANG also contains an `ru` locale (also reasonable).
Suppose the said installed package includes an `ru` help file.
But the user will get the `en` file, since the search will not find it in the data directory (that is in the user directory) and will then refer to the system directory.
With the API you mentioned, there is no way to search it in the system `ru` directory before referring to the system data directory which contains the `en` help file.

But `linkgrammar_get_file_location(char *fn)` solves this problem and doesn't leave unhandled cases as far as I can see.

linas:
OK. I want to publish version 5.5.0 "real soon now" .. should I wait for changes discussed above?

ampli:
Currently the help file will not get found if the user uses her/own data directory which was not entirely got duplicated from the system one. So using `linkgrammar_get_file_location()` may be a desired fix.
I can implement that and send a patch very shortly.

linas:
> it can just open the file instead of returning its location.

...

> with no cached location

The point of finding the path was, of course, to re-use that path to obtain all the other dictionary files.   If you ever decide to have the help file be split into two or more parts, that issue arises again:  returning the `FILE*` is not enough if you also need to get a second file from the same directory.

ampli:
Here is how to consistently read a dict which have several file components without a cached path:

The function `dictionary_six()` (or a one that replaces it) will first locate the first component (`4.0.dict`).
It will get its path. It will pass this path to the dict reading code, and then use this path to read all the other components.

A similar thing can be done for reading an help facility which consist of several files.

linas:
OK

PR #745, opened 2018-04-21
Cleanup

ampli:
Some of the changes:
- Python bindings: Allow verbosity values up to 120.
- Remove unneeded include files.
- Fix typos and fix/remove comments.
- Move fclose() after errno just in case.
- Prevent some int  type mix size/sign possible warnings (more ~1000 to fix...).

### DIFF
--- a/link-grammar/dict-common/file-utils.c
+++ b/link-grammar/dict-common/file-utils.c
@@ -367,17 +367,18 @@ char *get_file_contents(const char * dict_name)
 		if (0 == read_size)
 		{
 			bool err = (0 != ferror(fp));
-			fclose(fp);
 
 			if (err)
 			{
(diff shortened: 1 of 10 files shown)


link-grammar/dict-common/file-utils.c
```diff
@@ -359,25 +359,26 @@ char *get_file_contents(const char * dict_name)
...
 			{
 				char errbuf[64];
 
 				strerror_r(errno, errbuf, sizeof(errbuf));
+				fclose(fp);
 				prt_error("Error: %s: Read error (%s)\n", dict_name, errbuf);
 				free(contents);
 				return NULL;
 			}
+			fclose(fp);
```
linas: Moving fclose like this serves no purpose...

PR #752, opened 2018-04-24
Unrepeatable random

ampli:
Use random seed when repeatable_random=false (!rand=0).
I checked it by setting linkage_limit=1 and inspecting several linkages.

Note:
I didn't understand the following code in `sentence_split()`, so I don't know if any change is needed regarding it:
``` c
	/* 0 == global_rand_state denotes "repeatable rand".
	 * If non-zero, set it here so that anysplit can use it.
	 */
	if (false == opts->repeatable_rand && 0 == sent->rand_state)
	{
		if (0 == global_rand_state) global_rand_state = 42;
		sent->rand_state = global_rand_state;
	}
```
(I also didn't understand the comment.)

### DIFF
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -450,7 +450,16 @@ Sentence sentence_create(const char *input_string, Dictionary dict)
 
 	sent->dict = dict;
 	sent->string_set = string_set_create();
-	sent->rand_state = global_rand_state;
+
+	if (0 == global_rand_state)
+	{
+		sent->rand_state = 0;
+	}
+	else
+	{
+		sent->rand_state = (unsigned int)current_usage_time()*1000000;
+	}
+
 	sent->disjuncts_connectors_memblock = NULL;
 
 	sent->postprocessor = post_process_new(dict->base_knowledge);
--- a/link-grammar/resources.c
+++ b/link-grammar/resources.c
@@ -45,7 +45,7 @@ int getrusage(int who, struct rusage *rusage);
 #define MAX_MEMORY_UNLIMITED ((size_t) -1)
 
 /** returns the current usage time clock in seconds */
-static double current_usage_time(void)
+double current_usage_time(void)
 {
 #if !defined(_WIN32)
 	struct rusage u;
--- a/link-grammar/resources.h
+++ b/link-grammar/resources.h
@@ -25,4 +25,5 @@ bool      resources_memory_exhausted(Resources r);
 bool      resources_exhausted(Resources r);
 Resources resources_create(void);
 void      resources_delete(Resources ti);
+double    current_usage_time(void);
 #endif /* _RESOURCES_H */


link-grammar/api.c
```diff
@@ -450,7 +450,16 @@ Sentence sentence_create(const char *input_string, Dictionary dict)
...
 	sent->string_set = string_set_create();
-	sent->rand_state = global_rand_state;
+
+	if (0 == global_rand_state)
+	{
+		sent->rand_state = 0;
+	}
+	else
+	{
+		sent->rand_state = (unsigned int)current_usage_time()*1000000;
```
linas: this won't work, because current_usage_time() will almost always return zero or some small, fixed number, because it just counts the cpu time used since the program was started.

linas:
`global_rand_state` saves the rand state from one sentence to the next. 

The correct fix is to replace the '42' by address of `global_rand_state` or address of `sent`.  When the executable is loaded into memory, the memory addresses will be randomized, and this is sufficient to provide a seed.  A one-line fix is enough:  replace `42` by  `(unsigned int) (&sent & 0x7fffffff)`   and that's all.

ampli:
It seems to work, maybe because the CPU time that is returned is in microsecond resolution, so it changes "enough" between different calls to sentence_parse(). On Windows, clock() is in millisecond resolution, but for some reason it returns elapsed time (it is not the POSIX or c99 clock()), so it should work for setting from `link-parser` but maybe not from a program that issues a series of same sentence reparses within 1 millisecond.

I can redo it using a system API for clock time in microseconds (for both POSIX and Windows).

ampli:
> The correct fix is to replace the '42' by address of global_rand_state or address of sent. When the executable is loaded into memory, the memory addresses will be randomized, and this is sufficient to provide a seed. A one-line fix is enough: replace 42 by (unsigned int) (&sent & 0x7fffffff) and that's all.

But this will again give a repeatable sequence.
I.e., if I invoke link-parser and parse the same sentence more than once (with random pick), I will always get the same parses.

With my fix, I tried the following using the sentence `I am having trouble with it.`:
```
!rand=0
!limit=1
linkparser> I am having trouble with it.
```
I repeated the parse (using editline history) several times, and for every try I got a different linkage.
With your method I will get the same linkage.

linas:
I'm confused. The rand state is saved and set precisely to make it not repeatable, and this used to work, back when I first created the code.   After every sentence, the rand state is (should be) diffferent; if not, your changes are simply a work-around of some other bug, somewhere else.

Changing the 42 to some random value simply makes it non-repeatable across multiple invocations of the command line client.  Once you are in the client, it should be fully non-repeatable, on every sentence (or there's a bug).  It used to work fine.

linas:
... and, indeed, it is not working, right now. WTF.   I processed shit-tons of data with this flag. It worked when I last tested it. Arg. It takes me a month to process the data.

ampli:
> I'm confused. The rand state is saved and set precisely to make it not repeatable, and this used to work, back when I first created the code. After every sentence, the rand state is (should be) diffferent; if not, your changes are simply a work-around of some other bug, somewhere else.

I cannot find anything in the current code that does it. The seed starts with either 0 or 42, and each of them produces a repeatable sequence. If you recall the version in which it worked, we can find what happens to it.

ampli:
I now see how it was supposed to work... The last pseudo-random number was supposed to get saved and serve as a new seed for the next sentence.

If this is the way it supposed to work, I can check how to fix it.

ampli:
@linas,
Are you looking into it, or should I?

linas:
> The last pseudo-random number was supposed to get saved and serve as a new seed for the next sentence.

Yes, exactly.  Could you look at it?

ampli:
> Yes, exactly. Could you look at it?

OK.

PR #758, opened 2018-04-26
Add regex for <word>

ampli:


ampli:
This turned out to be ineffective.

PR #760, opened 2018-04-26
Fix unrepeatable random selection

ampli:
Issue #752.

The problem was that `sent->rand_state` has never been set to the linkage random state.

On first glance it seems it is more economical to set it once, from the last linkage.
However,  there may be linkages that are not produced after it, and if so the rand_state should be of the last one of them.

So I just added setting `sent->rand_state` in `extract_links()`.

Comments:
1. I don't know if this problem affected anysplit or if this fix interferes with it., since `sent->rand_state` is already assigned there.  **Please check that.**
2. It seems as if `sent->rand_state` is redundant since `globa_rand_state` could be directly used instead.
3. I still don't understand the reason for setting `sent->rand_state` in `sentence_split()`, as it already set in `sentence_create()`. If I try to remove it, it seems non-repeatable rand still works fine.
The comment of this code is somehow misleading and doesn't help to understand its purpose.

### DIFF
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -716,7 +716,10 @@ void extract_links(extractor_t * pex, Linkage lkg)
 		if (0 == pex->rand_state) repeatable = true;
 		if (repeatable) pex->rand_state = index;
 		list_random_links(lkg, pex, pex->parse_set);
-		if (repeatable) pex->rand_state = 0;
+		if (repeatable)
+			pex->rand_state = 0;
+		else
+			lkg->sent->rand_state = pex->rand_state;
 	}
 	else {
 		list_links(lkg, pex->parse_set, index);


linas:
This does not fix the ANY parsing reporting. I'm confused. I have to examine this more closely.

ampli:
> This does not fix the ANY parsing reporting

I don't understand...

Here is how to check that it indeed fixes the problem and allows unrepeatable linkage selection on !rand=0:

```
linkparser> !limit=1
linkparser> I am having trouble with it.
Found 21 linkages (1 of 1 random linkages had no P.P. violations)
	Unique linkage, cost vector = (UNUSED=0 DIS= 6.50 LEN=7)

    +---------------------Xp--------------------+
    |           +-------Out------+              |
    +->Wd--+-SX-+      +----AN---+---Mp---+-J-+ |
    |      |    |      |         |        |   | |
LEFT-WALL I.p am.v having.g trouble.n-u with it .

linkparser>  I am having trouble with it.
...
```
Recall this sentence several times and inspect its single (since !limit=1) parse.
It is the same all the time, as expected from the repeatable rand.

Now do (still with `!limit=1`):
```
linkparser> !rand=0
linkparser>  I am having trouble with it.
```
With my patch, **now you will get a different one for each try**!!!

linas:
OK.  Hadn't eaten dinner. Got confused.  Looks like it works.

linas:
> It seems as if `sent->rand_state` is redundant since `global_rand_state` could be directly used instead.

Yes, it is, but I wanted to limit the visibility of the global.

> I still don't understand the reason for setting sent->rand_state in sentence_split(), as it already

The `opts->repeatable_rand` is not visible to sentence_create.

PR #768, opened 2018-05-02
configure: Set JAVA_SRCDIR to an absolute path

ampli:
Hopefully this will fix the problem once for all.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -609,7 +609,8 @@ if test "x$enable_java_bindings" = "xyes"; then
 		CFLAGS="${CFLAGS} -arch x86_64"
 	fi
 
-	JAVA_SRCDIR=$srcdir/../../bindings/java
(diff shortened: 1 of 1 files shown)


linas:
Does the current code not work for windows?

linas:
OK, I just saw your other comment.  For me, I was using "case 3", building in a "build" directory, subdirectory of the source directory.

ampli:
> Does the current code not work for windows?
 
I guess you mean windows under MinGW (since under "pure" windows there is no "configure").
The 5.5.0 released `configure` generated Java bindings just fine in all the 4 combinations that I tested.
You latest change broke it  on Linux when building under `/tmp`. I didn't check but I'm pretty sure this problem would also happen with this change under MinGW (because configure sets $srcdir to an absolute path if configured in external directory, and then `$srcdir/../../bindings/java` is a nonexistent directory).

Issue #774, opened 2018-05-09
SQL dict is currently broken


ampli:
The current "condesc"  design doesn't support it.  Furthermore, `link-parser demo-sql` just crashes.
Also, LENGTH-LIMIT-n is not supported.

I will add "condesc" support so it will work.
Adding LENGTH-LIMIT-n support is more tricky, as the sql dict is not read in advance.
But if there is a partial-key access (need to check), all such entries can be read on dict open.

Should dynamic update of the sql db be supported (i.e. while the program runs)?
(This can make LENGTH-LIMIT-n very tricky if it can be added/changed dynamically.)

ampli:
I checked the last working version (5.4.3) of the sql dict, and found the following problems:
1. The `!!` `link-parser` command doesn't work. It seems particularly useful if it could work.
2. There is a memory leak:
```
    #0 0x7f1ec858c850 in malloc (/lib64/libasan.so.4+0xde850)
    #1 0x7f1ec80fdf1f in xalloc /usr/local/src/link-grammar-devel/tests/link-grammar/utilities.c:367
    #2 0x7f1ec804e079 in make_expression /usr/local/src/link-grammar-devel/tests/link-grammar/dict-sql/read-sql.c:106
    #3 0x7f1ec804e5b9 in exp_cb /usr/local/src/link-grammar-devel/tests/link-grammar/dict-sql/read-sql.c:151
    #4 0x7f1ec6e314b2 in sqlite3_exec (/lib64/libsqlite3.so.0+0x954b2)
```
3. There are no SQL dict tests.

I will try to fix all of that.

ampli:
From the docs:
>1.2.5 Multi-connectors. A word can also make an indefinite number of links of the same type to other words. For this, we use the "multi-connector" symbol "@". 

How this is done using the SQL dict?

linas:
> `!!` command doesn't work.

I suppose. It is not likely to be useful, at least in the short term; currently, the dictionaries are huge, and are a complete mess. 

> mem leak

No surprise

> no tests

tests should be created.

> @ symbol

Not sue. I think its stored in there. Don't know how its treated. The auto-genned dicts do not use @ 

> read sql dict

The original goal was to keep it in sql, rather than putting it into RAM, i.e. to let the sql engine do the "hard work" of managing memory.  They are designed to do that...

> dynamic update of the sql db be supported

[...]

PR #777, opened 2018-05-18
SQL dict fixes

ampli:
SQL dict fixes (see issue #774):
- Adapt the connector descriptor table for the SQL dict.
- Fix a memory leak.
- Avoid a crash if trying to show constituents.
- Actually read the affix file.
- Add a test in tests.py.

Minor changes:
- Check h/d in the same bitwise operation of checking the connector subscript.
- Don't compute uppercase part hash.
- Some code rearrangement.

Other changes:
- Change/rearrange comments.
- Remove unneeded include.
- Fix whitespace.

Note: Force-pushed because it didn't compile with no sqlite3.


linas:
thanks!

PR #778, opened 2018-05-20
link-parser usability fixes

ampli:
Currently `link-parser` uses stdout/stderr inconsistently. Especially, it uses both at once.
See the problem by:
`link-parser --help > /tmp/help`
`link-parser -xyz > /tmp/err`
(On Windows the problem is even more severe, as the stderr output is printed before the stdout one.)

stdout/stderr Fixes:
1. Send `link-parser --help` to stdout only (for the reason see the commit).
2. Send `link-parser -BOGUSARG` messages to stderr only.
3. Send user variable/command errors only to stderr.

Fix also 2 problems:
1. Prevent bogus usage message on `-!word` argument.
2. Require full command name for undocumented (hidden) commands. This allows to specify `!cost=x`.
(Maybe these hidden commands, which are not useful for now, should just ifdef'ed out.)

### DIFF
--- a/link-parser/command-line.c
+++ b/link-parser/command-line.c
@@ -669,7 +669,7 @@ static int x_issue_special_command(char * line, Command_Options *copts, Dictiona
 
 	if (count > 1)
 	{
-		printf("Ambiguous command.  Type \"!help\" or \"!variables\"\n");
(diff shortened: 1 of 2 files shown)


linas:
thanks, I think this makes sense.

PR #781, opened 2018-05-26
print_expression() fixes

ampli:
Main fixes:
1. Fix printing ([()] or X+) (currently the cost is neglected).
2. Print also negative costs (currently totally neglected).
3.  Fix printing costly optional (e.g. `[{X+}]`).
4. Use dyn_strcat() for all fixed strings (consistency and efficiency).
5. Don't print too many [] levels - use numbers instead (something like `[[[[[[[[X+]]]]]]]]` is not useful).
4. Avoid printing negative-zero costs such as `[X+]-0.00` (may be result of cost arithmetics).

### DIFF
--- a/link-grammar/dict-common/print-dict.c
+++ b/link-grammar/dict-common/print-dict.c
@@ -132,7 +132,6 @@ static dyn_str *print_expression_parens(dyn_str *e,
 		return e;
 	}
 
-	/* Look for optional, and print only that */
 	el = n->u.l;
 	if (el == NULL)
 	{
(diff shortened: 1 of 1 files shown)


ampli:
Please don't apply yet - the "Exp print: Avoid [X+]-0.00" commit committed unintended stuff.
I will force-push a fix.

ampli:
Forced-push to fix bad commits.
Now it is ready.

linas:
thanks!

PR #782, opened 2018-06-03
Linkparser

ampli:
Fix showing info for "string.sub" that can split.
(Plus several minor fixes.)

Proposals: 
1. `dict_display_word_info()` and `dict_display_word_expr()` are documented (in `link-includes.h`) as "not intended for general public use". I suggest to still define them under `Clinkgrammar`, but with an `_` prepended to them (to mark them as "internal use") so they can get tested.
2. To generally test `link-parser`, I propose to create a `link-parser`  invocation test suite.


linas:
Thanks!

PR #786, opened 2018-06-17
SAT-parser fixes

ampli:
1. Check for alternative-compatibility when doing connector matches.
This speeds up sentences that have many solutions with "insane" morphism.
2. Fix several comments (comment rot).
3. Allow displaying bad linkages (postprocess violations) when there are no good linkages.

BTW, postprocess violations currently extremely slow down the SAT-parser. I will try to fix that by a feedback from the postprocessing into the SAT encoding.
(This fix can be used to also speed up the postprocessing for the classic parser.)


ampli:
4. Minor speed improvements.
5. Additional push: Don't perform pp-prune if `perform_pp_prune==false`.

PR #788, opened 2018-06-21
Misc. cleanups

ampli:
Fix code and comment typos.
Update comments.

Add USE_VITERBI over use_viterbi - can the relevant parse_option calls be removed from link-grammar.def?:w

### DIFF
--- a/link-grammar/Makefile.am
+++ b/link-grammar/Makefile.am
@@ -205,7 +205,7 @@ EXTRA_DIST=                         \
 
 # Clean up generated file that Windows MSVC compilation leaves behind.
 clean-local:
-	rm -f $(top_builddir)/link-grammar/link-features.h
+	rm -f $(top_srcdir)/link-grammar/link-features.h
 
 # -----------------------------------------------------------
 install-data-local: install-libtool-import-lib
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -70,15 +70,15 @@ struct Parse_Options_s
 {
 	/* General options */
 	short verbosity;       /* Level of detail to give about the computation 0 */
-	char * debug;          /* comma-separated function names to debug "" */
-	char * test;           /* comma-separated features to test "" */
+	char * debug;          /* Comma-separated function names to debug "" */
+	char * test;           /* Comma-separated features to test "" */
 	Resources resources;   /* For deciding when to abort the parsing */
 
 	/* Options governing the tokenizer (sentence-splitter) */
-	short use_spell_guess;  /* Perform spell-guessing of unknown words. */
+	short use_spell_guess; /* Up to this many spell-guesses per unknown word 7 */
 
 	/* Choice of the parser to use */
-	bool use_sat_solver;   /* Use the Boolean SAT based parser */
+	bool use_sat_solver;   /* Use the Boolean SAT based parser FALSE */
 #if USE_VITERBI
 	bool use_viterbi;      /* Use the Viterbi decoder-based parser */
 #endif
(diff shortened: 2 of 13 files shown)


link-grammar/Makefile.am
```diff
@@ -205,7 +205,7 @@ EXTRA_DIST=                         \
 
 # Clean up generated file that Windows MSVC compilation leaves behind.
 clean-local:
-	rm -f $(top_builddir)/link-grammar/link-features.h
+	rm -f $(top_srcdir)/link-grammar/link-features.h
```
linas: Oof. Nothing should ever get written into the srcdir during build. This suggests a different bug.

link-grammar/api-structures.h
```diff
@@ -70,16 +70,18 @@ struct Parse_Options_s
...
 	/* Options governing the tokenizer (sentence-splitter) */
-	short use_spell_guess;  /* Perform spell-guessing of unknown words. */
+	short use_spell_guess; /* Up to this many spell-guesses per unknown word 7 */
 
 	/* Choice of the parser to use */
-	bool use_sat_solver;   /* Use the Boolean SAT based parser */
+	bool use_sat_solver;   /* Use the Boolean SAT based parser FALSE */
+#if USE_VITERBI
 	bool use_viterbi;      /* Use the Viterbi decoder-based parser */
+#endif
```
linas: We are approaching a day (if not already past it) when the viterbi code should be removed. It is perhaps a bad idea.

link-grammar/connectors.c
```diff
@@ -70,7 +70,7 @@ Connector * connector_new(const condesc_t *desc, Parse_Options opts)
 }
 
 /* ======================================================== */
-/* UNLIMITED-CONNECTORS handling. */
+/* Connector length limit handling. */
```
linas: Well, its in the dict as `UNLIMITED-CONNECTORS` so its nice to have it be searchable by that string name.   UI dunno. Maybe there is some other easy way to find this code.

link-grammar/Makefile.am
ampli: This is Windows, not Un*x...
The LG README documented stuff used to tell to copy `link-grammar/link-features.h.in` into link-`grammar/link-features.h`. Now the Windows build stuff does the same (but better - w/changing the @variables@).

This  cleanup command is if you later want to compile on a Un*x machine (my src dir is remote-mounted on Windows, with each branch at its own directory).

*Proposal*:
I can instead change the build procedure to create it in the `msvc14` directory, and add this directory to the Windows LG-build include path. The only problem is that this name `msvc14` is a moving target... so when renamed, a change will need to be done in several places too (but maybe I will find a way to not use an explicit name in the definitions).

link-grammar/api-structures.h
ampli: I can just remove it from link-grammar-def and ifdef-out the remaining stuff with USE_VITERBY.

link-grammar/Makefile.am
ampli: Or we can just have this cleanup rule to support compiling from Windows in the same directory....

linas: I'm not sure I care, at this point. In principle source dirs should be read-only during builds, but there are so many other packages that break this - maybe the majority of all packages ... there's only so many decisions one can make in a day, I don't know if I want to make this one.

link-grammar/api-structures.h
linas: Yes. I looked at it again, and I just don't see any future for it.

linas:
yes.

link-grammar/Makefile.am
ampli: Possibilities:
1. Remove the current rule (it is harmful, as it removes a file created by `configure`).
2. Change it to remove from srcdir, which is a no-op for proper Un*x usage (but can be a bad surprise for those who anyway use the srcdir for build).
3. Also (1), plus putting it under `msvc14` (or whatever it will be renamed to).

Just tell me what to do...

ampli: I will remove the rule altogether, and will later (on conversion to MSVC15) will try to see if I can generate `link-grammar/link-features.h` under `msvcX` w/o mentioning `msvcX` anywhere.

PR #790, opened 2018-06-23
Fixes per discussion at PR #788

ampli:
Per the discussion at PR #788:
- Delete the bad `link-feature.h` clean-local entry.
- Complete the viterbi ifdef-out (+ change the existing ones from `#if` to `#ifdef` just for consistency).
- Update the length-limit function section comment.


linas:
re viterbi, I meant `git rm -r viterbi`

PR #795, opened 2018-07-06
 Fix crash on linkage_get_word_*() with combined morphemes ++

ampli:
The fixes here (see the details in the commit messages):
- Fix crash on linkage_get_word_*() with combined morphemes
- Fix wrong unsplit_word of combined morphemes in the display wordgraph path
- Remove w/ from LPUNC
- Don't hide morphology for languages with no INFIX_MARK
- tests.py: Add possibility for setting parse options in parse*.txt
- tests.py: Add getting word position tests
- tests.py: Fix broken library locale setup testing


ampli:
Maybe a fix is needed in `linkage_get_word_char_start` and `linkage_get_word_char_end()`:
They return `0` in case of a too big word index:
 `if (linkage->num_words <= w) return 0; /* bounds-check */`
But `0` is a valid return value for both start (of LEFT_WALL and the first word) and end (LEFT-WALL).
I propose to make it `-1` (like in some clib functions that return `size_t`).

linas:
Yes, -1 would work.

linas:
The python-examples directory is getting crowded.  Perhaps there should be a python-tests directory, either in `bindings` or in the main `tests` directory?  When the tests were simple, they made for good "examples"; now that they're complicated,  I fear that they mostly confuse the reader, and serve to hide the simplest example(s).

Issue #800, opened 2018-07-20
python unit tests are failing for spell-check


linas:
running `tests.py` generates an error:
```
AssertionError: 
'LEFT-WALL(0, 0) the(0, 3) seasonal[~].a(4, 11) lakes.n(12, 17) are.v(18, 21) hot.a(22, 25) RIGHT-WALL(25, 25)\nLEFT-WALL(0, 0) the(0, 3) seasonal[~].a(4, 11) lakes.n(12, 17) are.v(18, 21) hot.a(22, 25) RIGHT-WALL(25, 25)\n'
!= 
'LEFT-WALL(0, 0) the(0, 3) seas[&].n(4, 8) and[&].j-n(8, 11) lakes.n(12, 17) are.v(18, 21) hot.a(22, 25) RIGHT-WALL(25, 25)\nLEFT-WALL(0, 0) the(0, 3) seas[&].n(4, 8) and[&].j-n(8, 11) lakes.n(12, 17) are.v(18, 21) hot.a(22, 25) RIGHT-WALL(25, 25)\n'
: at ./parses-pos-spell-en.txt:12
```
which is confirmed at the command line:
```
linkparser> the seasand lakes are hot
Found 50 linkages (50 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS= 0.00 LEN=11)

    +--------------->WV-------------->+
    +----------->Wd------------+      |
    |      +--------Dmc--------+      |
    |      |        +-----A----+--Spx-+--Pa-+
    |      |        |          |      |     |
LEFT-WALL the seasonal[~].a lakes.n are.v hot.a

Press RETURN for the next linkage.
linkparser> 
	Linkage 2, cost vector = (UNUSED=0 DIS= 0.00 LEN=13)

    +------------------>WV------------------>+
    +---------->Wd-----------+------Spx------+
    |      +--Dmc-+---SJlp---+--SJrp--+      +--Pa-+
    |      |      |          |        |      |     |
LEFT-WALL the seas[&].n and[&].j-n lakes.n are.v hot.a

Press RETURN for the next linkage.
linkparser> 

	Linkage 3, cost vector = (UNUSED=0 DIS= 0.10 LEN=11)

    +-------------->WV------------->+
    +---------->Wd-----------+      |
    |      +-------Dmc-------+      |
    |      |       +----AN---+--Spx-+--Pa-+
    |      |       |         |      |     |
[...]

PR #801, opened 2018-07-20
Fix for issue #800

ampli:
Fix issue #800.

Reason of the problem:
The check passes with "hunspell" but not "aspell", because "aspell" generates a spell guess which changes the order of the linkages.

Fixed (for now, at least) by reducing the number of allowed spell corrections from 10 to 1.

EDIT: In the same occasion fix typos and improve the opcode check.

---
This is not a general fix, as the problem may theoretical reappear here and in other tests.

I initially made a more general fix, but omitted it due to its added complexity and because the current fix is simpler. (I will keep it around in case it will ever be needed for something.)
In this fix, a regex optional argument to the "N" (generate next linkage) opcode is used in order to obtain the next matching linkage (tested again a concatenation of the linkage words).
In the case of this check, which is intended to check the position of word spell splits (marked with `[&]`),  the check definition would be:
```
Ithe seasand lakes are hot
N /&/
P ...
P ...
P
N /&/
```
(Part of this omitted fix was to make the first "N" a no-op and only use its argument.)

### DIFF
--- a/bindings/python-examples/lg_testutils.py
+++ b/bindings/python-examples/lg_testutils.py
@@ -7,7 +7,7 @@ def add_eqcost_linkage_order(original_class):
     be in a deterministic order.
     Usage: lg_testutils.add_eqcost_linkage_order(Sentence)
     """
-    class eqcost_soretd_parse(original_class.sentence_parse):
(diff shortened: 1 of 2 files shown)


linas:
Thanks!

PR #806, opened 2018-09-13
Windows fixes

ampli:
I'm trying to improve the Windows ports (MinGW, Cygwin and MSVC) so they will be in the same level as on Linux. This is a work in progress, but it has mostly be done.

The number of commits for that is now very big, and I also did many more general fixes in the way (that help with Windows too).
This big number of changes can make it hard to review, due to the volume and diversity.

So I will gather the commits mostly according to type (Windows, configure, tests, and more) and will start to send PRs.
(Beside of that, I have a huge number of backlog changes in many branches, some are incomplete, but some only need final touch, or merging due to code changes meanwhile.)

This is the first PR from the backlog.
Main changes:
- WIN32: Add #define for strtod_l()
  Fix compilation problem MSVC in the last PR (#805).
- MSVC: Call _set_printf_count_output(1) to enable using %n format
- Windows: Convert argv from the startup locale to UTF-8 (solve possible crahshs)
- Use fputs() instead of printf() - printf() on Windows sometimes has redirection problems maybe because problems related to POSIX functions
- Windows: jni-client.c: Disable langinfo.h and its usage
- Documentation fixes and updates
- Support DICTIONARY_DIR on MinGW and MSVC
- Fix configuring hunspell on MinGW
- MinGW: Solve hunspell crash on multi-threading
- MinGW: Avoid a (bad) macro definition of rand_r
- Fix compilation warnings
- Cygwin: Python bindings
- MinGW: Fix the problematic Python bindings

Still WIP (for future PRs):
- Windows wordgraph-display improvements.
- Optional use of pcre2 (motivated by a need for Cygwin, but it runs much faster on Linux too).

### DIFF
--- a/msvc/README.md
+++ b/msvc/README.md
@@ -292,8 +292,8 @@ Samba), and still be able to use the custom build steps in the Project files,
 there is a need to "convince" Windows it is a local filesystem.  Else you will
 get "UNC path are not supported." on the batch runs, with bad results. This
 method will also allow the `link-parser.bat` file to run.  (For other solutions
-see tackoverflow.com/questions/9013941). You will need to find out by yourself
-if this makes a security or another problem in your case.
+see https:/stackoverflow.com/questions/9013941). You will need to find out by
+yourself if this makes a security or another problem in your case.
 
 Here is what worked for me:<br>
 Suppose you use `host:/usr/local/src` remotely as share `src`:
--- a/msvc/README.md
+++ b/msvc/README.md
@@ -7,7 +7,7 @@ This directory contains project files for building Link Grammar with the
 Microsoft Visual Studio 2017 IDE (MSVC15). They were created and tested with
 the Community Edition of that product.
 
-**!!!WARNING!!!**
+**!!!WARNING!!!**<br>
 In the examples below, "console-prompt>" at start of line means the console
 prompt.  Don't type it because the result could be destructive!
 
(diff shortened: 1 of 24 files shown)


msvc/README.md
```diff
@@ -263,11 +263,11 @@ For using the library independently of the build directory:
 
 1) If Python bindings were generated, copy the following modules to a
    directory `linkgrammar` in a fixed location: `linkgrammar.py`,
-   `clinkgrammar.py`, `__init__.py`, `_clinkgrammar.pyd`.
+   `clinkgrammar.py`, `__init__.py`, `_clinkgrammar.pyd`.<br>
```
linas: instead of `<br>`, how about just a newline?

msvc/README.md
```diff
@@ -291,15 +291,31 @@ In order to use a link-grammar source repository from a network share (e.g. via
 Samba), and still be able to use the custom build steps in the Project files,
 there is a need to "convince" Windows it is a local filesystem.  Else you will
 get "UNC path are not supported." on the batch runs, with bad results. This
-method will also allow the "link-parser.bat" file to run.  (For other solutions
-see tackoverflow.com/questions/9013941). You will need to find out by yourself
-if this makes a security or another problem in your case.
+method will also allow the `link-parser.bat` file to run.  (For other solutions
+see https:/stackoverflow.com/questions/9013941). You will need to find out by
```
linas: two slashes after `https://`

ampli: Will be fixed...

PR #809, opened 2018-09-16
Fix  read dict cost when ther is no locale_t + add/change tests

ampli:
1. Locale-related tests:
-- Fix converting dict cost to float in case no locale_t:
On MinGW there is still no locale_t (this is the only system that I have access to with no locale_t, on MSVC/Linux/BSD/OSX there is locale_t). In that case my recent fix 01f3044bcba9631311de407598872eb19475f321 is buggy.
-- Add dict cost reading with a comma-separator LC_NUMERIC.
-- Move the per-dict locale test to its own class.
So if it fails, the test teardown will restore the previous locale.
Also skip it if no tr locale (may be better than a test failure).

2. Test creating dictionary using various dictionary directory paths:
-- Test opening dictionary by relative/absolute paths.
It validates success and also that the dictionary language is set properly.
This also validate the recent directory-separator handling change.

The tests run successfully on Linux, MinGW64, MSVC 15, and newest macOS (using Homebrew libtr and python). On Cygwin the regex `\w` test fails - I will send a patch soon.

### DIFF
--- a/link-grammar/dict-file/dictionary.c
+++ b/link-grammar/dict-file/dictionary.c
@@ -154,6 +154,11 @@ dictionary_six_str(const char * lang,
 	}
 
 	/* Read dictionary from the input string. */
+
+	/* Make sure "." is used in string conversion to floating point.
+	 * FIXME: Use a locale-agnostic conversion function. */
+	setlocale(LC_NUMERIC, "C");
+
 	dict->input = input;
 	dict->pin = dict->input;
 	if (!read_dictionary(dict))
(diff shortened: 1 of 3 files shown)


link-grammar/dict-file/dictionary.c
```diff
@@ -154,6 +154,11 @@ dictionary_six_str(const char * lang,
...
 
 	/* Read dictionary from the input string. */
+
+	/* Make sure "." is used in string conversion to floating point.
+	 * FIXME: Use a locale-agnostic conversion function. */
+	setlocale(LC_NUMERIC, "C");
+
 	dict->input = input;
 	dict->pin = dict->input;
 	if (!read_dictionary(dict))
```
linas: It would be best to put the `setlocale` into `dictionary_setup_locale` which already does that kind of work.

The call to `dictionary_setup_locale` is immediately below, after the dictionary read (!) ... that does not seem right...

ampli: `dictionary_setup_locale` is after the dictionary read because it needs to know the dictionary locale in order to set it.

PR #813, opened 2018-09-25
Configuration changes

ampli:
This PR includes `configure` changes and corresponding changes in source files and documentations when needed. It also contains build documentation addition and changes.

Main changes:
- Fix compilation issues on FreeBSD and add documentation for building on it.
- Fix Java detection failure - make it a soft failure.
- Take care for C++ linking if sat-solver is configure. The current way is incompatible with **clang**.
- Show the compiler vendor (to make it clear when `gcc` is not really GNU `gcc`, like on macOS).
-  Rewrite the Cygwin documentation. Currently it is like the Linux build, including Python and Java bindings, and pass all the tests. Other language bindings have not been tested (but they are untested on Linux too for much time now.) However, its multi-threading is very slow.
- General build documentation update.
- Remove obsolete build doc on old MinGW and rename the current MinGW doc.
- Add default hunspell dict for macOS. Also use the name **macOS** in other places instead of OSX and other names because it is now the official name.

More `configure` changes:
- Clean up the `TLS` disabling option (is it actually needed?).
- Show in the summary if the hunspell dict directory is not found.
- Show in the summary the path names of the python binaries that will have bindings.
- Just with no real good reason: Use `AS_UNSET` instead of `unset`.
- Use `AC_MSG_WARN` instead of `echo WARN`.
- Unify `CFLAGS` and `CXXFLAGS` for various systems.


linas:
wow. Impressive.

PR #815, opened 2018-09-27
MSVC: Fix get_dictionary_dir() to use LG DLL address

ampli:
Recently a possibility of relative DICTIONARY_DIR has been re-added to the MSVC code. However, this code makes it relative to the location of the link-parser executable, but the user may use programs in other locations, that use the LG library.

This change makes DICTIONARY_DIR (if it is a relative path) relative to the location of the LG DLL.

---
BTW, maybe it will be better to also add environment variable for the dictionary data directory that will take precedence over  DICTIONARY_DIR. (I think that there once was such a variable but it was been removed long ago.)

Now it is possible to produce a zip-file distribution of Windows -10 MSVC-compiled code, that the user will be able to extract and use in any filesystem location.
(If we would like to also provide Python bindings, we will also need to include a binary distribution of python3 in this zip-file.)

**Question:**
Will it be a good idea to do that?

(Of course we should also recommend users to use LG on Linux if possible.
I guess that a Linux-distribution LG package will run just fine under  [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10).)

BTW, I think it may also be possible to make a zip-file distribution of Cygwin compilation result by including in a zip-file a selective portion of the Cygwin file system after install LG. But we will need the said environment variable, or maybe better, to conditionally compiled the code of this PR also on Cygwin.

### DIFF
--- a/link-grammar/dict-common/file-utils.c
+++ b/link-grammar/dict-common/file-utils.c
@@ -131,120 +131,149 @@ static void free_dictionary_dir(void)
 
 /**
  * Return the path of the system dictionary directory.
- * If DICTIONARY_DIR is an absolute path (must be so on POSIX systems)
- * then it is directly used.
- * On Windows, a relative DICTIONARY_DIR is combined with the location
- * of the invocation directory of our program.
- * Note: The directory separators in the path are always supposed to be
- * forward slashes.
- * @param find If true, find the system dictionary directory; else free it.
+ * If DICTIONARY_DIR is an absolute path (must be so on systems other than
+ * Windows) then it is directly used.
+ * On Windows (currently implemented on MSVC only), a relative
+ * DICTIONARY_DIR is combined with the location of this library DLL, and
+ * if it includes a drive letter or is a UNC path then it must be
+ * absolute. It shouldn't end with a directory separator.
+ * @param find If true, find the system dictionary directory; else free it
+ * if needed.
+ * @return The actual dictionary directory.
  */
 static const char *get_dictionary_dir(bool find)
 {
 #ifndef _MSC_VER
 		return DICTIONARY_DIR;
 #else
+	/* We are on Windows. DICTIONARY_DIR, if defined, can be relative or
+	 * absolute. An absolute path must start with a drive letter or be a
+	 * UNC path, and a relative path should not start with a drive letter
+	 * or be a UNC path.
+	 * In case it is not so, the result is undefined.  If it is a relative
+	 * path, it is concatenated to the program location. */
+	static const char *dictionary_dir;
 
-	/* DICTIONARY_DIR, if defined, can be relative or absolute.  An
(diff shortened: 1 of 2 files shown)


linas:
I know little about what people expect in the windows world.   Ideally, there would be a distribution maintainer, who would decide before-hand where to install files, and configure LG at compile time to find files in the expected locations.  Yet, from what I can tell, there is no such thing as a "distribution" for windows.

I somewhat dislike environment variables, as these increase the attack surface for malicious actors.

ampli:
> Yet, from what I can tell, there is no such thing as a "distribution" for windows.

There are [installation packages](https://docs.microsoft.com/en-us/windows/desktop/msi/installation-package). They can be managed (installed/reinstalled/repaired/removed/upgraded) by a Windows package manager. There is even an official Unix-like (command-line) one.

My first though was to use something like [Chocolatey](https://chocolatey.org/) that is free for open source. However, there is no need to start like that (and indeed many projects don't care to make an official install package for their distribution).

**What we can do instead:**
An LG release can just be compiled by MSVC, and a Zip file of it (including the source code, but maybe after cleanup of unneeded intermediate results like object files) can be provided for download. The user can unzip it everywhere on the computer, and just put the executable directory in the PATH (or alternatively, put customized copies of provided `link-parser.bat` and `lgpython.bat` files in some directory). This allows easy installation, and also easy use of  more than one LG release in parallel (easy to upgrade and develop). In addition, the file tree can be prepared in such a way (externally, not in the LG repository) that one can just start Visual Studio, point it to the provided Solution file (File->Open) and press F7 to recompile everything.

So in such a scheme there is no need to decide where to install the LG system dictionary and the python components - they just remain in their place in the source code tree.

(Or maybe distribute LG+Ubuntu Docker image to be used with Docker for Windows? But I don't know  how to do that...). 

Environment variables are not really needed for the data directory even for Cygwin, as the Cygwin port can use the same scheme like the MSVC port (i.e. a data directory relative to the library DLL). This may be better than having an additional thing (an environment variable) to document and debug... But see below...

> I somewhat dislike environment variables, as these increase the attack surface for malicious actors.

BTW, even without an environment variable for the data directory, malicious actors that can change environment variables may change (among really many other environment variables) LD_PRELOAD and LD_LIBRARY_PATH and possibly have an easier control. Many other libc components use environment variables. E.g. `malloc()` - see the manual of `mallopt`,  or the locale, timezone, NLS and ld.so stuff - LD_DEBUG_OUTPUT can even create files. (Web servers and the like are said to sanitize the environment and you need to explicitly allow things to pass.)

linas:
Well, we do not create packages for ubuntu, debian, redhat, suse, etc. so I'm not really interested in creating packages for windows, either.  On the other hand, the standard convention for unix is that the source containers never create packages.  I don't know what the expected, standard behavior for windows is.

PR #817, opened 2018-09-28
Remove the viterbi code

ampli:
Per discussion at PR #790.

(Maybe there is a need to mention something more about that.)
In any case I starting to work on interactive parsing, but before I continue with that I decided to first publish some old fixes.


link-grammar/dict-file/dictionary.c
```diff
@@ -295,15 +295,15 @@ Dictionary dictionary_create_from_file(const char * lang)
...
  * This routine is intended for unit-testing ONLY.
  *
- * FIXME? (Used only by code that is in use for now.)
+ * FIXME? (Not used for now.)
  * 1. get_default_locale() returns locale, not language.
- * 2. "lang" memory leak on success.
+ * 2. "lang" memory-leak on success.
  */
 Dictionary dictionary_create_from_utf8(const char * input)
 {
```
linas: We should keep this, as it is something that Anton & friends keep seeming to want to use, but never actually get around to using.  Another is that it is potentially useful for writing unit tests.

ampli: I can fix it.
But I think it should be generalized to read strings for all the file components - not only the dict file:
`dictionary_create_from_utf8(const char *lang, const chart *dictdir, const char *dict, const char *regex, const char *affix, ...)`
The `lang` argument is just for setting dict->lang.
if `dictdir`is not NULL, for NULL components it will use the corresponding files from there.
BTW I propose to define "lang" in the dict file and not use the directory name for that by default.

linas: Since there are no users at this time, over-engineering this does not seem to make sense.

ampli: My only use for specifying regex/affix files by strings is for adding tests in case new features are added to them.  So I guess this may indeed be postponed until it is needed.

PR #818, opened 2018-10-01
SAT cleanup and speedup

ampli:
Main changes:

Speed:
- Add/fix debug printouts related to to parse speed. Including:
 -- Print  step times on -verbosity=2 similarly to what is done in the classic parser.
 -- Add -test<values> (and document in debug/README.md).
- Remove a redundant comparison in expression handling.
- Fix and re-enable the main code in the main power_prune() code. It got broken (and hence disabled)
when I added the optional-word fix (to replace the use of empty-words). Note that the purpose of the power pruning code is only to add restrictions in order to speed up the SAT-solving.
- Add a partial CONTAINS_NONE_RULES  pruning to reduce the number of postprocessing failures. This is a proof-of-concept that demonstrates that encoding postprocessing rules may have a significant speed-up effect.

Cleanup:
- Remove dead code.
- Ensure word variables are equal to word numbers (to allow code changes, as in this PR).
- Fix typo/whitespace.
- Port unused code to minisat 2.2 (in case it will be needed).

TODO:
- Add much more hashing - for various expression aspects. Among other things, currently the pruning code  has much encoding overhead due to the need of rescanning a large amount of data.
- Add more postprocessing pruning.
- **New idea**: Prevent duplicate linkages due to duplicate disjuncts.
Motivation: The classic-parser works with disjuncts, and thus duplicate ones can be eliminated. This saves parsing overhead and also prevents duplicate results. However, the SAT-parser works "directly" on expressions and thus is not able to prevent duplicate disjuncts. Because duplicate results is exponential to the number of different disjuncts that have duplicates and to the number of duplicates, a very large amount of duplicates may happen (a low percentage of distinct results). This has overhead implications in case several results are needed (they may be needed because the SAT-parser is not effectively able now to provide the best results first).
The idea is to prohibit further solutions with the exact same links after each solution (instead of the current prohibition of only the same exact solution).

### DIFF
--- a/debug/README.md
+++ b/debug/README.md
@@ -171,6 +171,11 @@ For more examples of how to use the wordgraph display, see
 [link-grammar/README.md](/link-grammar/README.md)
 and [msvc/README.md](/msvc/README.md).
 
+5) -test=<values> for SAT parser debugging:
+linkage-disconnected - Display also solutions which don't have a full linkage.
+sat-stats - Display the number of PP-violations and disconected linkages.
+no-pp_pruning_1 - Disable a partial CONTAINS_NONE_RULES pruning
+
 Debugging and STDIO streams
 ---------------------------
 Messages at severity Info and higher (i.e. also Warning, Error and
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -319,7 +319,7 @@ static void clean_table(unsigned int size, C_list ** t)
  * Return true iff they may become adjacent (i.e. all the words
  * between them are optional).
  */
-static bool optional_gap_collapse(Sentence sent, int w1, int w2)
+bool optional_gap_collapse(Sentence sent, int w1, int w2)
 {
 	for (int w = w1+1; w < w2; w++)
 		if (!sent->word[w].optional) return false;
--- a/link-grammar/parse/prune.h
+++ b/link-grammar/parse/prune.h
@@ -17,5 +17,6 @@
 
 int        power_prune(Sentence, Parse_Options);
 void       pp_and_power_prune(Sentence, Parse_Options);
+bool       optional_gap_collapse(Sentence, int, int);
 
 #endif /* _PRUNE_H */
--- a/link-grammar/sat-solver/clock.hpp
+++ b/link-grammar/sat-solver/clock.hpp
@@ -2,8 +2,8 @@
 #define __CLOCK_H__
 
 /**
-*   Time measurment functions
-*/
+ *   Time measurement functions
+ */
 
 #include <iostream>
 #include <ctime>
@@ -15,19 +15,17 @@ public:
 	Clock() {
 		reset();
 	}
-	
+
 	void reset()
 	{
 		start = clock();
 	}
 
-public:	
+public:
 	double elapsed()
 	{
 		clock_t stop = clock();
 		return ((double)stop-(double)start)/CLOCKS_PER_SEC;
-	
 	}
 };
-
 #endif
--- a/link-grammar/sat-solver/clock.hpp
+++ b/link-grammar/sat-solver/clock.hpp
@@ -8,6 +8,8 @@
 #include <iostream>
 #include <ctime>
 
+#include "error.h"
+
 class Clock {
 private:
 	clock_t  start;
@@ -26,6 +28,14 @@ public:
 	{
 		clock_t stop = clock();
 		return ((double)stop-(double)start)/CLOCKS_PER_SEC;
+	}
+
+	void print_time(int verbosity_opt, const char *s)
+	{
+		if (verbosity_opt >= D_USER_TIMES)
+		{
+			printf("++++ %-36s %7.2f seconds\n", s, elapsed());
+		}
 	}
 };
 #endif
--- a/link-grammar/sat-solver/guiding.hpp
+++ b/link-grammar/sat-solver/guiding.hpp
@@ -91,11 +91,6 @@ public:
     setParameters(var, isDecision, 0.0, 0.0);
   }
 
-#if 0
-  /* thin_link variables */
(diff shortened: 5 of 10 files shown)


linas:
Cool!  Would it make sense to have the SAT parser work on disjuncts, instead of expressions?

linas:
Oh, ah, interesting. I think I now understand what it was about the cost system that was bugging me, in the other post on the other pull req. So you can disregard that notice. 

Actually, this is very interesting from the theoretical point of view: its a key place where a combinatorial explosion happens, and I will have to ponder what that means.  Conversely, when working with disjuncts, one could perform "compression" by working with expressions instead, and that this might be an advantageous place to perform compression (i.e. simplification).  Very interesting.

Would this be possible: During dictionary read, expand all expressions into disjuncts (thus eliminating duplicates) and then compress a set of disjuncts back down to a collection of expressions, which, when expanded, no longer would generate duplicates?

ampli:
> [...] and then compress a set of disjuncts back down to a collection of expressions, which, when expanded, no longer would generate duplicates?

Is it a conversion from CNF to DNF and back to CNF?
I guess this is doable, but I suspect the result may not be too compact.

BTW, as far as I recall, much of the duplication is a result of discarding optional connectors in expression pruning, that leaves subexpressions like `( [()] OR () OR X+ OR [[[()]]])`. It may be easy enough to retain only the lowest cost null in each such subexpression. Maybe this can be done as the last step of the current expression pruning.

linas:
Well, expressions are not currently in CNF, they are in whatever ad-hoc form written is in the dictionary, right?

The future machine-learned dictionaries will be almost entirely in DNF (at least, in the short term), so if SAT needs/wants/works better with CNF, then it will need to perform an explicit conversion.

Yes, converting to DNF allows duplicate high-cost expressions to be eliminated.

Yes, retaining the lowest-cost null in subexpressions might eliminate most duplicates (that's for you to find out); but clearly it won't eliminate all of them.

ampli:
> Would this be possible: During dictionary read, expand all expressions into disjuncts (thus eliminating duplicates) and then compress a set of disjuncts back down to a collection of expressions, which, when expanded, no longer would generate duplicates?

Maybe it can be done this way, with an additional benefit due to power pruning:
1. Perform the power pruning and duplicate disjunct elimination.
2. Mark in the expressions the connectors that appear in the resulted disjuncts.
3. Prune unused parts of the expressions.

ampli:
> Maybe it can be done this way, with an additional benefit due to power pruning:
> 
> 1. Perform the power pruning and duplicate disjunct elimination.
> 2. Mark in the expressions the connectors that appear in the resulted disjuncts.
> 3. Prune unused parts of the expressions.

I implemented it and it makes a very significant expression pruning (on English but not Russian - I still need to investigate why).
The SAT solver then runs faster on the fixes batch, and even several times faster on long sentences.
There are also much less duplicate parses.

(Should I make a new issue from this message?) 

Remaining problems:
1. Reduced speed on Russian (seems as if no expression pruning is done after the feedback, so the overhead of the extra processing is showing).
2. Two sentences from the fixes batch don't parse, a thing that seems to be due to a cost-cutoff problem, as if somehow connectors with lower cost got pruned in the expression due to duplicate disjuncts (I still investigate that).
3. I couldn't completely eliminate all the duplicate disjuncts due to a subtle sane-morphism problem - I didn't find a way to propagate discarded connectors that belong to duplicate disjuncts that originate from different tokens (most of the duplicate disjuncts don't have this problem).

ampli:
> Well, expressions are not currently in CNF, they are in whatever ad-hoc form written is in the dictionary, right?

Yes,it's just in some mixed notation.

> The future machine-learned dictionaries will be almost entirely in DNF (at least, in the short term), so if SAT needs/wants/works better with CNF, then it will need to perform an explicit conversion.

I think such a conversion may incur very much overhead by itself so I don't know it will help.
It will be interesting to compare the performance of the classic and SAT parsers on pure DNF expressions (it is possible to convert the current en dict for such a test).

Anyway, here is a quote from page 56 of the SAT parser doc:

--
**Comparison of DNF and CNF approach.** Carefull examination of classic
link-parser implementation shows that it uses sophisticated DNF based
algorithms. On the other hand, SAT based link-parser is uses CNF based
algorithms.
DNF based algorithms used in classic link-parser imlementation suffer from
the problem that the number of disjuncts can become unmanageably large for
sentences that are long and complicated. The time and memory required for
generating and storing the DNF formula becomes inacceptably large. CNF
based approaches do not suffer from these problems. Even for very long and
complicated sentences, formulas remain manegeable and are generated reasonably
fast. On the other hand, once the DNF formula is constructed classic
implementation can perfrom the search for satisfying solutions very fast. This
indicates that a breaktrough with the SAT approach can happen with long and
complicated sentences which are the bottleneck of the classic implementation.
However, the search time for the CNF formula grows with the formula and can
be very long for complicated formulae.

ampli:
BTW, maybe it is possible to simplify the current dict expressions by a such a feedback from duplicate disjunct elimination. But I don't know if this has any significant benefit (besides maybe save some expression memory) since duplicate disjunct elimination will still be needed at runtime (due to multiple tokens in each token slot).

linas:
> simplify the current dict expressions

Not for the files, themselves; they're just barely human-readable, and changes are always threatening to make them more unreadable.

linas:
> I implemented it and it makes a very significant expression pruning 

Awesome!  I don't quite understand what you did, but still, awesome!

For Russian, I assume its the length=1 stem-suffix links. An earlier perl-only hack for Russian, before the current system, created a mini-dictionary consisting only of the words that appeared in the sentence, and parsed with that dictionary (containing only the whole-words).  So a trick like that could work again, except it would be a large re-design. Roughly speaking, its like running sane-morphism before starting the parse ... 

> Should I make a new issue from this message?

Yeah, probably. Most/all of the discussion here should have happened in an issue.

Issue #785, opened 2018-06-17
Fixing panic mode


ampli:
These is the current setup of panic mode:
```
	parse_options_set_disjunct_cost(opts, 4.0f);
	parse_options_set_min_null_count(opts, 1);
	parse_options_set_max_null_count(opts, 100);
	parse_options_set_max_parse_time(opts, 60);
	parse_options_set_islands_ok(opts, false);
	parse_options_set_short_length(opts, 12);
	parse_options_set_all_short_connectors(opts, true);
	parse_options_set_linkage_limit(opts, 100);
	parse_options_set_spell_guess(opts, 0);
```

Proposed fixes:
1. If `!null=0`, I propose to set `min_null_count=0` and `max_null_count=0` since there is no reason to allow parsing with nulls in that case.
2. Copy `islands_ok` from the regular setup instead of hard-coding it to `false`.
3. Don't set `short_length` and `linkage_limit` to more than the current ones.
4. Document the user-variable prefix `panic_`.
5. Add `DEFAUT_COST` and PANIC_COST` optional dict entries (this will address issue #390).
6. Add panic timeout the the SAT parser.

linas:
This sounds very reasonable.  Could even set `linkage_limit` to be maybe 1/2 or 1/4th of the current limit (but at least one).

Likewise, `max_time` could be set to current max_time, or maybe even half of the current max_time. Or double max_time, but double seems not too wise.

ampli:
I'm now trying to produce a PR for fixing this issue and found something that looks like a bug.
This is the condition for doing a panic parse:
https://github.com/opencog/link-grammar/blob/972ea245bde7365ee6942610154da88a46d12c1f/link-parser/link-parser.c#L1005-L1008
Note the
https://github.com/opencog/link-grammar/blob/972ea245bde7365ee6942610154da88a46d12c1f/link-parser/link-parser.c#L1005
condition and consider. This check is old (from the start of version 4).

Consider this:  When the timer is exhausted, nested `do_count()` calls return count zero, originally on connector endpoints that are not in the table, and unconditionally after I added the connector-word cache. In both cases, the number of parses may be non-zero but the parses are then most probably incomplete. However, `link-parser` just ignores this condition and may print an incomplete parse without any indication that it is problematic.

ampli:
I'm in the last step of producing a PR to fix most of these problems and some others.
[...]

Issue #783, opened 2018-06-03
Cost cutoff


ampli:
# Background
I'm trying to fix the SAT-parser disjunct cost calculations to get accurate linkage disjunct and sentence costs, and mostly succeeded with that. (See issue #188 on the ignoring the costly-nulls cost there.)

(BTW, I started again to investigate the SAT inaccurate disjunct cost problem due to the WIP of dialects support (issue #402). For inspecting the modified expressions I needed to fix `print_expression()` (PR #781), and due to the needed investigation of expressions I realize I may have a fix to the said SAT-parser problem.)

When validating to the disjunct costs as reported by my said SAT-parser fix, I inspected these costs as reported by the classic parser. I found an apparent problem in the cost cutoff as applied by the classic parser or in my understanding of it - in any case one of them needs a fix...

# The apparent problem in the classic parser
I noted that sometimes disjuncts which have `disjunct->cost > cost_cutoff` are used in the parse.
This is because `build_disjunct()` uses the criteria `cl->maxcost <= cost_cutoff` but set the actual cost of the disjunct to cl->cost (which seems to be the correct cost) which is often greater than `cl->maxcost`.

## Example
Test sentence: What's up?
``` 
...
cost-max     Largest cost to be considered                  2.70
...
	Linkage 3, cost vector = (UNUSED=0 DIS= 3.00 LEN=4)

    +---------Xp---------+
    +---->WV---->+       |
    +-->Ws--+Ss*w+--K-+  |
    |       |    |    |  |
LEFT-WALL what 's.v up.r ?

            LEFT-WALL    0.000  hWs+ hWV+ Xp+ 
                 what    0.000  Ws- Ss*w+ 
                 's.v    3.000  Ss- dWV- K+ 
                 up.r    0.000  K- 
                    ?    0.000  Xp- RW+ 
           RIGHT-WALL    0.000  RW- 
```

There are examples (other sentences) of even larger disjunct costs that are used (with cost-max=2.7).

The disjunct in question (for `'s.v`) Is:
`[Ss- & dWV- & [K+ & [()]]]`
The `& [()]]` is due to expression pruning. The cost is indeed 3.
What is maxcost, and why it can apparently be set to be less then the cost?

(My fix to the SAT-parser also somehow neglects to make this same cutoff.)

linas:
> What is maxcost, and why it can apparently be set to be less then the cost?

Seems like a bug to me; `build_disjunct()` would appear to be mis-applying the cutoff if-test.

[...]

Issue #188, opened 2015-09-19
Fixing the SAT solver


ampli:
Hi Linas!

I mostly finished the full integration of the SAT solver.
In addition to the previous fixes, that are listed in the comment in #187, I have fixed several **more** aspects of the SAT solver:
1. Initialized the domain array so it doesn't crash on !links.
2. It now works fine on Hebrew too.
3. It frees properly all memory.
4. I rearranged some of the code to be more similar to that of the standard resolver.
5. It now works fine on batches!
* It gives **exactly** the same number of errors on all the languages.
* From first glance, differences in diagrams of errors are due to different linkages order.
* Initial batch runtime comparisons to the regular solver:
   - Slightly slower on en/4.0/batch.
   - Definitely faster on en/4.0/fixes.batch.
   - About twice faster on the Russian batch.

​6. The maximum number of shown linkages is now **!limit** instead of the arbitrary 222.

---
Remained to fix:
1. Different costs are shown in !disjuncts.
2. The DIS= cost in the status message is always 0.
3. It still checks only alternative[0] in in guiding.cpp. I need to check if fixing this will help finding linkages faster for the affected sentences.
4. More code can be made similar, and more code can be shared (i.e. the SAT solver code can call more functions from the regular LG lib).
However, this will need adding more functions to **link-grammar.def** (I already added many).
**Alternatively**, and maybe better, the shared functions can be moved to the corresponding **.h** files, and be removed from **link-grammar.def**. This will solve the problem of the many non-API  symbols that are exported by the LG lib.
5. It may be interesting to replace the current MiniSAT code (an old release) by the latest release (version 2.2). Github: [niklasso/minisat](https://github.com/niklasso/minisat).
6. Please add more issues if I forgot something...

linas:
The link-grammar.def file is not needed by Linux; its used only by Windows and MacOS, where it plays a vital role. Basically, it lists the functions that are publically visibile from a shared library. 

I guess you are suggesting that we declare these functions as public API functions in various header files, but I really really don't want to do that, since they real are not means to be used by the public.   A better solution would be to make the sat library be a STATIC libraries (not shared) and then link it directly into the LG library, if SAT compilation is enabled.  That way, I believe we could remove most of the crud from the link-grammar.def file.

ampli:
On Linux I just had to add these functions to linkgrammar.def - there was no need to define them as public PAI.

if so, isn't this linkgrammar.def addition enough also for Windows and MacOS?

[...]

PR #820, opened 2018-10-02
SAT solver cost

ampli:
Fix the SAT parser disjunct cost to take into account null expression cost. See (1) in this post in issue [#188](https://github.com/opencog/link-grammar/issues/188#issuecomment-143903475). (It seems I didn't open a new issue for that and instead continued the discussion in the same issue.)

Also fix the total cost per disjunct to be according to the calculation that is done in the classic parser, including the cost-cutoff calculation (see issue #783).

The result is that the reported costs (can be inspected with `!disjunct`) should now be identical to those reported by the classic parser. Note that because duplicate disjuncts are not discarded, and they may have different costs (in the classic parser only the one with the lowest cost is retained) the same parse can be shown multiple times, with different costs.

I had a difficulty to word the ChangeLog description.
I also used there "Modify" and not "fixed" because I am not convinced yet that the cost-cutoff calculation method that is currently used is "correct" (or "optimal").

I did comparison tests (SAT vs classic parsers) when I produce all the possible solutions for each input sentence.  The order of the parses is of course different, and also the SAT parser produces many duplicates. But by using md5 hashes of each parse I validated they they produce exactly the same parses. I didn't try to programmatically match the reported disjunct costs - I only inspected them manually and they seem identical (if you compare to the linkage with the lower total cost among the duplicates).

If you find different costs please tell me and I will try to investigate/fix that.

Some remaining problems (in addition to the open one from issue #188):
1. I suspect that the cost-guiding solution preference is currently broken (the intention is that the SAT solver will produce solutions with lower cost first.) I still don't know how to fix that.
2. The reported connector order (as displayed by `!disjuncts`) is not correct (so the original disjuncts cannot be correctly reconstructed and used). I know how to fix that.
3. Panic timeout is not implemented (will be fixed when fixing #785 is done).
4. "Robust parsing" (parsing with nulls) is not implemented. I started this work but still need to learn more some SAT related material in order to make an efficient implementation.
5. Improving the speed in various ways (there is very much room for that).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,7 @@ Version 5.5.2 (XXX 2018)
  * Major documentation update for building with Cygwin.
  * Revise the manpage.
  * Remove the experimental Viterbi code.
+ * Modify the SAT parser disjunct cost metric to be like in the classic parser.
 
 Version 5.5.1 (27 July 2018)
  * Fix broken Java bindings build.
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -582,7 +582,6 @@ int sentence_link_cost(Sentence sent, LinkageIdx i)
 {
 	if (!sent) return 0;
 
-	/* The sat solver (currently) fails to fill in link_info */
 	if (!sent->lnkages) return 0;
 	if (sent->num_linkages_alloced <= i) return 0; /* bounds check */
 	return sent->lnkages[i].lifo.link_cost;
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -573,7 +573,6 @@ double sentence_disjunct_cost(Sentence sent, LinkageIdx i)
 {
 	if (!sent) return 0.0;
 
-	/* The sat solver (currently) fails to fill in link_info */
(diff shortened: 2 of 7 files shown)


linas:
The changelog entry is fine.  A 'corporate-speak' variation of if it might  be

`Revise the SAT parser cost model to align it with the classic parser.`
 
I'm still nervous about the finer details of how the costs are being computed and handled. I'm working on the theoretical side (you've seen the `skippy.pdf`, I'm sure) and it's not obvious that what I'm thinking of there actually matches what the code is doing. However, I don't have time to think about this now. Just be aware that, in the long run, costs need to behave as if they were log-probabilities, or rather, as if they were mutual-information. More or less. Details TBD.

ampli:
> It anyway must be reassigned inside the loop for each iteration (line 1584).

However, I admit it is not clear enough. I'm not sure that adding an initial assignment will make it clearer,
since it is not actually needed.

So the question remains: How to make it clearer?

linas:
It's fine. I'm reading the diffs, not the full code, so things like loop structure are not particularly visible.

Issue #799, opened 2018-07-13
link-parser fails to load dictionary although warning that error word would be ignored


alexei-gl:
`link-parser` falls into fatal error although previously saying that a word with reserved suffix will be ignored. I realize that the suffix ".Ix" is reserved and should not be used in a way other than for idioms, but it seems to me that in that case if it is sad 'Warning' the word should either be simply excluded from parsing rules and the process continue or it should not be a warnig but a fatal error to avoid incertanity.
```
link-parser poc-turtle_90C_2018-06-01_0004/
link-grammar: Info: Dictionary found at ./poc-turtle_90C_2018-06-01_0004/4.0.dict
link-grammar: Warning: Word "C.I" found near line 234 of poc-turtle_90C_2018-06-01_0004/4.0.dict.
	Words ending ".Ix" (x a number) are reserved for idioms.
	This word will be ignored.Error opening word file /a/
link-grammar: Info: Freeing dictionary poc-turtle_90C_2018-06-01_0004/4.0.dict
link-grammar: Fatal error: Unable to open dictionary.
```
The dictionary is in attachment.
[dictionary.tar.gz](https://github.com/opencog/link-grammar/files/2190799/dictionary.tar.gz)

ampli:
Thanks for the bug report.
To be fixed.

ampli:
It turns out the idiom-suffix error is not the one that caused the dict read to fail, and is already properly ignored.

The fatal error was:
`Error opening word file /a/`

It is easy to miss it due to 3 error-message print formatting problems:
1. A missing newline after `This word will be ignored.`.
The fix adds it there and also in two additional places in which it is missing.
2. The word file name `/a/` is not quoted, and thus is not easy to spot.
The fix adds double-quotes over it (and also in additional similar places).
3. This is an "old style" error, apparently from the times before introducing error severities.
The fix was to prefix it with `Error:` and change it appropriately. I didn't find additional such messages.

Now these error combination is to be reported as:
```
link-grammar: Warning: Word "C.I" found near line 234 of poc-turtle_90C_2018-06-01_0004/4.0.dict.
	Words ending ".Ix" (x a number) are reserved for idioms.
	This word will be ignored.
Error: Cannot open word file "/a".`
link-grammar: Info: Freeing dictionary poc-turtle_90C_2018-06-01_0004/4.0.dict
link-grammar: Fatal error: Unable to open dictionary.
```

[...]

PR #822, opened 2018-10-02
Fix some dict error reporting

ampli:
Handle issue #799.

The actual problem is not the one that has been reported.
See more on that there.


linas:
thanks!

PR #829, opened 2018-10-11
Speed up build-disjuncts

ampli:
This patch is for a major speedup of the disjunct building out of expressions.

The ideas behind the changes are:
1. Cost cutoff pruning while building the clauses. This is a major saving.
2.  Improved Tconnector list traversing.
3.  Not trying to concatenate empty lists (it's much overhead to traverse a list only to append a NULL).
4. Avoid reversing lists altogether. The new list traversing/build algos build them in the needed order.
5. Extract both types of connectors (+ and -) at once.
6. Use memory pools.
7. Don't free intermediate memory until the end.

In addition this patch includes slight memory-pool changes:
1. Ifdef out the yet unused POOL_EXACT feature.
2. Actually check the POOL_FREE feature (but it is left ifdef'ed out).
3. Add ASAN checks for POOL_FREE and fake-pools (fake-pools is a debug feature).
TBD: Add ASAN checks to the whole memory-pool system.

I validated the correctness of the change by producing all the linkages for en/corpus-fixes.batch (except for the "The problem is" sentences, which overflows). They are completely identical.

The speedup of building the disjuncts is > 50% and may be way over 80% for long sentences.
For example, check the long sentence at the end of `ru` corpus:
```
# BEFORE
$ link-parser ru -v=2
...
linkparser> как и ожидалось, он заявил, что дсб не поддержит вотум недоверия правительству, предложенный левой  изза отказа от проекта
...
++++ Built disjuncts                         0.41 seconds
...

# AFTER
++++ Built disjuncts                         0.07 seconds
```

### DIFF
--- a/data/en/4.0.dict
+++ b/data/en/4.0.dict
@@ -14,6 +14,10 @@
 <dictionary-version-number>: V5v5v2+;
 <dictionary-locale>: EN4us+;
 
+t1: Wd- & A+ & ([()] or [[[()]]]) & C+;
+t2: A- or () or [()];
+t3: C-;
+
  % _ORGANIZATION OF THE DICTIONARY_
  %
  % I. NOUNS
--- a/link-grammar/memory-pool.h
+++ b/link-grammar/memory-pool.h
@@ -68,4 +68,27 @@ struct  Pool_desc_s
 	bool exact;                 // Abort if more than num_elements are needed.
 #endif /* POOL_EXACT */
 };
+
+// Macros for our memory-pool usage debugging.
+// https://github.com/google/sanitizers/wiki/AddressSanitizerManualPoisoning
+#if !defined(__has_feature)
+#define __has_feature(x) 0
+#endif
+#if __has_feature(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
+#include <sanitizer/asan_interface.h>
+#define ASAN_POISON_MEMORY_REGION(addr, size) \
+	__asan_poison_memory_region((addr), (size))
+#define ASAN_UNPOISON_MEMORY_REGION(addr, size) \
+	__asan_unpoison_memory_region((addr), (size))
+#define ASAN_ADDRESS_IS_POISONED(addr) \
+	__asan_address_is_poisoned(addr)
+#else
+#define ASAN_POISON_MEMORY_REGION(addr, size) \
+	((void)(addr), (void)(size))
+#define ASAN_UNPOISON_MEMORY_REGION(addr, size) \
+	((void)(addr), (void)(size))
+#define ASAN_ADDRESS_IS_POISONED(addr) \
+	((void)(addr), false)
+#endif // __SANITIZE_ADDRESS__
+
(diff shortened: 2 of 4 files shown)


ampli:
Pushed again for:
1. Whitespace fix of: `build-disjuncts.c: Use memory pools`
2. Add a commit: `Tconnector: Reduce memory footprint`

data/en/4.0.dict
```diff
@@ -14,6 +14,10 @@
 <dictionary-version-number>: V5v5v2+;
 <dictionary-locale>: EN4us+;
 
+t1: Wd- & A+ & ([()] or [[[()]]]) & C+;
+t2: A- or () or [()];
+t3: C-;
+
```
linas: left-over debugging crud ...

link-grammar/memory-pool.h
```diff
@@ -63,6 +69,31 @@ struct  Pool_desc_s
...
+#define ASAN_ADDRESS_IS_POISONED(addr) \
+	__asan_address_is_poisoned(addr)
+#else
+#define ASAN_POISON_MEMORY_REGION(addr, size) \
+	((void)(addr), (void)(size))
+#define ASAN_UNPOISON_MEMORY_REGION(addr, size) \
+	((void)(addr), (void)(size))
+#define ASAN_ADDRESS_IS_POISONED(addr) \
+	((void)(addr), false)
+#endif // __SANITIZE_ADDRESS__
```
linas: This probably should have been two pull reqs: one to fixup the mem pool, the other to change the connector stuff.   I'm going to merge as is, though.

link-grammar/prepare/build-disjuncts.c
```diff
@@ -13,22 +13,19 @@
...
 
 /* Temporary connectors used while converting expressions into disjunct lists */
 typedef struct Tconnector_struct Tconnector;
 struct Tconnector_struct
 {
-	char multi;   /* TRUE if this is a multi-connector */
-	char dir;     /* '-' for left and '+' for right */
 	Tconnector * next;
-	const condesc_t * condesc;
+	const Exp *e;
```
linas: Hmm. This is now a strange and confusing name. The old `Tconnectors` were obviously connectors: a multi, a dir, a connector-struct. The new one is .. a linked-list of expressions.

I guess that each expression in the list describes exactly one connector, right? There should be a note that explains this.  A name-change would be good, also: rename `Tconnector` to `SingleConnectorExp`  or something like that.

linas: Or 'ConnectorPerExp` or `ConnectorInExp` or ... or maybe we can leave it as Tconnector, as long as there's an explanation of the expected contents.

ampli: > I guess that each expression in the list describes exactly one connector, right? There should be a note that explains this.

Yes.
> .. or maybe we can leave it as Tconnector, as long as there's an explanation of the expected contents.

I will add an explanation.

Issue #833, opened 2018-10-25
The ru/4.0.*knowledge files  just add unneeded overhead


ampli:
These files have some content so `pp_knowledge_open()` will not give an error.
This has the following slight disadvantages for `ru` and languages that don't actually use them:
1. Overhead in dict opening file (~2.5% for `ru`). 
2. Overhead **for each sentence**, since `pp_prune()` is then tried for each sentence, with null results. This is of course a small overhead relative to the rest of processing, but it is unneeded.
3. When !constituents=x is set, dummy output is produced.

Solutions;
1. Allow for empty files, which will disable them with no message. This has a disadvantage of disabling them by mistake for English, with bad results.
2. Add a directive DISABLE, which will signify that the file is disabled. This has a disadvantage
of checking for this string also for English, but this overhead is negligible.

I implemented (2) above in  `pp_knowledge_open()`.
With it, trying to set e.g. `!cons=1` results in the message
```
link-grammar: Error: Can't generate constituents.
Constituent processing has been turned off.
```
but I don't think this needs a fix.

If this is fine, I can send a PR.

PR #836, opened 2018-10-27
Pruning minor speed ups

ampli:
- Address #833: Add DISABLE directive for 4.0/*knowledge
- Also skip unneeded operations when no pruning is done:
 -- expression_prune(): zero_connector_table()
 -- pp_prune(): delete_unmarked_disjuncts()

BTW, I tried the following, but they only seem to add overhead, so they are not included here:
1. In expression_prune(): Skip purge_EXP()/clean_up_expressions() in case  nothing got marked.
2. In power_prune(): Skip handling of BAD_WORD marks if nothing got deleted.
This was surprising since many times nothing is deleted (especially in the last iteration).


linas:
thanks!

ampli:
Regarding commit 695fb34:
Can we change  `;DISABLE` to just `DISABLE`, so using  (e.g. for debug and validations) the dict by recent previous versions that don't know this construct will not generate a lot of warnings?

linas:
> DISABLE

Sure. You added this code :-)

Issue #838, opened 2018-10-29
LG URLs at link-parser's help text


ampli:
I would like to add URL info in the help text of `link-parser` (both `--help` and `!help`) as provided in the manpage. (For `--help`, this is according to the  [GNU coding standards](https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dhelp).)
I shortened the URL text descriptors to enable using a table format that still fits in 79 character line length.

If this seems fine, I will send a PR right away.

---
``` text
Usage: /tmp/link-grammar/help/link-parser/.libs/lt-link-parser [language|dictionary location]
                   [-<special "!" command>]
                   [--version]

Special commands are:
 Variable     Controls                                          Value
 --------     --------                                          -----
 bad          Display of bad linkages                           0 (Off)
 batch        Batch mode                                        0 (Off)
 constituents Generate constituent output                       0
 cost-max     Largest cost to be considered                  2.70
 disjuncts    Display of disjuncts used                         0 (Off)
 echo         Echoing of input sentence                         0 (Off)
 graphics     Graphical display of linkage                      1 (On)
 islands-ok   Use of null-linked islands                        0 (Off)
 limit        The maximum linkages processed                 1000
 links        Display of complete link data                     0 (Off)
 morphology   Display word morphology                           0 (Off)
 null         Allow null links                                  1 (On)
 panic        Use of "panic mode"                               1 (On)
 postscript   Generate postscript output                        0 (Off)
 ps-header    Generate postscript header                        0 (Off)
 rand         Use repeatable random numbers                     1 (On)
 short        Max length of short links                        16
 spell        Up to this many spell-guesses per unknown word    7
 timeout      Abort parsing after this many seconds            30
 use-sat      Use Boolean SAT-based parser                      0 (Off)
 verbosity    Level of detail in output                         1
 debug        Comma-separated function names to debug            
 test         Comma-separated test features                      
 walls        Display wall words                                0 (Off)
 width        The width of the display                      16381
 wordgraph    Display sentence word-graph                       0

Toggle a Boolean variable as in "!batch"; Set a variable as in "!width=100".
Get a more detailed help on a variable as in "!help var".

Home page:                     https://www.abisource.com/projects/link-grammar/
Discussion group mailing list: link-grammar@googlegroups.com
Peer-reviewed papers:          https://www.link.cs.cmu.edu/link/papers/
Source code repository:        https://github.com/opencog/link-grammar
[...]

PR #839, opened 2018-10-31
Add URLs at link-parser's help text

ampli:
Per issue #838.

Add the following printout to -`-help` and `!help`:
``` text
Overview:            https://en.wikipedia.org/wiki/Link_grammar
Home page:           https://www.abisource.com/projects/link-grammar/
Documentation:       https://www.abisource.com/projects/link-grammar/dict/
Discussion group:    https://groups.google.com/d/forum/link-grammar
Report bugs to:      https://github.com/opencog/link-grammar/issues
```

The URLs are defined at the start of `configure.ac`. They are extracted from there also by the Windows build. The usual URL for the discussion group contains `!`, which is not preserved when extracted by Windows (a fix is welcome - it seems complex). Hence I used an alternative URL.

Also, as mentioned in issue #838:
In the usage message use the basename of the program instead of its full path.
BTW, in the same occasion of touching this code, I simplified some of it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,16 @@
-AC_INIT([link-grammar],[5.5.1],[link-grammar@googlegroups.com])
+# This file is processed by the Windows build system using confvar.bat. It
+# assumes AC_INIT is the first command, with no blank lines before it and with
+# a blank line after it. If this things are changed, change confvar.bat
+# accordingly.
+# Don't end the URLs with a slash even if it appears in the official address,
+# because we may append a component. Also, they cannot include "!".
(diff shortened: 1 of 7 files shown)


ampli:
One minute - I will push it again to remove an extra slash...

linas:
I don't see an extra slash, but I also don't see a commit in the last half hour. Will wait till tomorrow.

ampli:
I removed the extra slash and used "push -f" to fix the bad commit.
Hence you don't see an extra one.
It's now OK. You may apply it just now.

linas:
thanks

Issue #402, opened 2016-09-16
Dialect support!


linas:
Below is a sketch of how to add dialect support, and why its a good idea.

Currently, {} is used to indicate optional connectors: for example:  A+ & {B- & C+} indicates that (B- & C+) is optional.

Lets give options names!  These names will be names of dialects!  So, for example: A+ & {B- & C+}{irish}  means that  (B- & C+) is optional, but only if the "irish" dialect is enabled; otherwise, it is never allowed.

In my imagination, this solve zillions of problems. These include:

A) the bad-spelling problem: create a kant-spel dialect, that merges together the disjuncts for they're there and their (and throws in thier, for good measure)

B) enhanced support for ... irish-english, black-american-english, australian-english, hillbilly-basilect, archaic 19th-century English, twitterese, newspaper-headlines

C) Automatic detection of dialects!  So, for example, if a sentence does not parse normally, but does parse after enabling some dialect, we can guess that it must be that dialect.

D) post-parse parse-ranking.  That is, parse a sentence with all dialect enabled, but then fiddle with the costs associated with each particular dialect.  Thus, to turn off the kant-spel dialect, one simply gives those connectors a very high cost, and they would be raked last.

linas:
BTW, this is very similar to the "named disjuncts" idea in https://groups.google.com/forum/#!msg/link-grammar/fFUhgSO0oL4/6RPgcRfpBAAJ  just using a different syntax.

To really get named disjuncts, one should allow the syntax (A+ & B+){some-name}  which means that (A+ & B+) is mandatory, and is given "some-name".   We can still pretend that "some-name" is a dialect, and we can also give it a cost outside of the dicationary. By default, all names have a cost of zero.

ampli:
You propose
```
{DISJUNCT}{DIALECT}
```
to select a disjunct by a dialect.
But what if I want to deselect disjuncts? It can be very cumbersome if only disjunct including by dialect is possible.
Maybe:
```
{DISJUNCT}{-DIALECT}
```
Similarly, a syntax for different costs by dialect seems to me desired.
Maybe:
```
{DISJUNCT}[default_cost][cost_for_DIALECT1]{DIALECT1}
```
I.e., {DIALECT} /  {-DIALECT} just select/deselect the item before them.

There  may be several other shortcuts needed, like letting one symbol represent several similar dialects that only slightly differ.

linas:
[...]

Issue #250, opened 2015-12-14
parsing performance is getting worse... 


linas:
Amir: consider this:

```
time echo "In vivo studies of the activity of four of the kinases, KinA, KinC, KinD (ykvD) and KinE (ykrQ), using abrB transcription as an indicator of Spo0A~P level, revealed that KinC and KinD were responsible for Spo0A~P production during the exponential phase of growth in the absence of KinA and KinB." | link-parser ./data/en -timeout=300 
```

how long does this take?  With todays 5.3.3 parser and 5.3.3 dicts: 48 seconds.   Some other times:

```
5.3.3 parser 5.3.3 dict: 48 sec
             5.2.5 dict: 48 secs
             5.1.3 dict: 44 sec
             4.8.6 dict: 22 sec
```

i.e. all of the above with todays parser, but other dicts.

But the 4.8.6 parser with the 4.8.6 dicts parsed this sentence in 8 seconds.   So, despite performance improvements, things have gotten a lot worse.

Some of this I understand: changes to the dicts do add more possibilities.  Some I don't: why is the the 5.3.3 parser 4x slower than the 4.8.6 parser, when both use the same dict?   According to my notes, I once parsed the above in under 4 seconds, using 5.1.0 ...  so .. wtf

ampli:
Hi Linas,

Consider that the default constraints setup is in favor of the old version.
In 4.8.6 we have:
cost-max 2.00
short 10
While in 5.3.3 we have:
cost-max 2.70
short 16

Besides of this, we must disable spelling on such tests, since different versions handle it differently (different alternatives may be generated).

Also, the latest improvements concentrated on batch processing.

So I suggest to make the benchmark with similar parameters:
`link-parser DICT -spell=0 -short=10 -max-cost=2 -echo -batch -timeout=300`
(Or: `link-parser DICT -spell=0 -short=16 -max-cost=2.7 -echo -batch -timeout=300`)

But even with this (-short=10 -max-cost=2) we get a large performance difference in favor of the old versions (using the 4.8.6 dict):
4.8.6: 1.31s
5.3.3: 6.11s

Using -verbosity=5 reveals a possible problem in 5.3.3:
```
power prune cost: 9427707
++++power pruned                            4.76 seconds
```
While in 4.8.6:
```
2858193 power prune cost:
++++power pruned (ruthless)                 0.63 seconds
```
We need to do bisections to find out when this apparent pruning problem happened.

[...]

PR #845, opened 2018-11-05
Prune speedup

ampli:
Corpus batch benchmarks (10x10 runs):
en basic: 2+%
en fixes: 3+%
ru: ~0.5%

Individual runs  of a small sample of very long sentences shows ~5% speed up.
Runs with `-verbosity=2` show a big power_prune() speedup. However, `power_prune()` currently takes only a small fraction of the total CPU, so the total saving is not as big.

The largest speedup is due to the shallow/deep connector ordering that allows an early termination of the loop in `*_table_search()` , that is implemented in the following commit:
prune.c: Matching shallow connectors: Stop earlier

Some other changes didn't show a noticeable save in their benchmarks, so they are not included here:
1. Using memory pools for `pp_prune()`.
2. Consolidating all the memory allocations and initializations in `power_table_new()`.
But now I may know better and can improve their implementation, so I may include them in a future PR.

Some other changes in this PR:
- Reformat touched code and localize some of its variables.
- Add some code for ALT_*_CONSISTENCY while pruning (still commented out, more changes are needed in order to make more efficient).
- Make `power_prune()` static.
- Define `lc_easy_match()` as inline (I guess it got inlined anyway).

(I also have major exprune.c changes with a greater speedup impact, to be sent soon.)

BTW, here is a `power_prune()` speedup idea for a future check:
Eliminate `bool shallow` in `struct c_list_s`, in order that it will be exactly half CPU cache line size.
- If we had Connector_struct in the expressions, it could be added there. It is not clear if using Connector_struct in expressions will worth their memory handling overhead, but they have other advantages that may offset this overhead.
-  Alternatively, the hashing can be done separately for shallow/deep connectors (with the expense of using some more memory), and this will save the "shallow" checks in `*_table_search()` loop.

### DIFF
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -295,20 +295,19 @@ static power_table * power_table_new(Sentence sent)
  * who are obsolete.  The word fields of an obsolete one has been set to
  * BAD_WORD.
  */
-static void clean_table(unsigned int size, C_list ** t)
+static void clean_table(unsigned int size, C_list **t)
 {
-	unsigned int i;
-	C_list * m, * xm, * head;
-	for (i = 0; i < size; i++) {
-		head = NULL;
-		for (m = t[i]; m != NULL; m = xm) {
-			xm = m->next;
-			if (m->c->nearest_word != BAD_WORD) {
-				m->next = head;
-				head = m;
-			}
-		}
-		t[i] = head;
+	for (unsigned int i = 0; i < size; i++)
+	{
+		C_list **m = &t[i];
+
+		while (NULL != *m)
+		{
+			if ((*m)->c->nearest_word == BAD_WORD)
+				*m = (*m)->next;
+			else
+				m = &(*m)->next;
+		};
 	}
 }
 
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -627,7 +627,7 @@ right_connector_list_update(prune_context *pc, Connector *c,
 }
 
 /** The return value is the number of disjuncts deleted */
-int power_prune(Sentence sent, Parse_Options opts)
+static int power_prune(Sentence sent, Parse_Options opts)
 {
 	power_table *pt;
 	prune_context pc;
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -659,9 +659,6 @@ int power_prune(Sentence sent, Parse_Options opts)
 					for (c=d->left;  c != NULL; c = c->next) c->nearest_word = BAD_WORD;
 					for (c=d->right; c != NULL; c = c->next) c->nearest_word = BAD_WORD;
 					N_deleted++;
-					total_deleted++;
-				}
-			}
 
 			clean_table(pt->r_table_size[w], pt->r_table[w]);
 			nd = NULL;
@@ -677,6 +674,8 @@ int power_prune(Sentence sent, Parse_Options opts)
 			}
 			sent->word[w].d = nd;
 		}
+
+		total_deleted += N_deleted;
 		lgdebug(D_PRUNE, "Debug: l->r pass changed %d and deleted %zu\n",
 		        pc.N_changed, N_deleted);
 
@@ -695,7 +694,6 @@ int power_prune(Sentence sent, Parse_Options opts)
 					for (c=d->right; c != NULL; c = c->next) c->nearest_word = BAD_WORD;
 					for (c=d->left;  c != NULL; c = c->next) c->nearest_word = BAD_WORD;
 					N_deleted++;
-					total_deleted++;
 				}
 			}
 			clean_table(pt->l_table_size[w], pt->l_table[w]);
@@ -713,6 +711,7 @@ int power_prune(Sentence sent, Parse_Options opts)
 			sent->word[w].d = nd;
 		}
 
+		total_deleted += N_deleted;
 		lgdebug(D_PRUNE, "Debug: r->l pass changed %d and deleted %zu\n",
 		        pc.N_changed, N_deleted);
 
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -593,21 +593,20 @@ right_connector_list_update(prune_context *pc, Connector *c,
 int power_prune(Sentence sent, Parse_Options opts)
 {
 	power_table *pt;
-	prune_context *pc;
+	prune_context pc;
 	Disjunct *d, *free_later, *dx, *nd;
 	Connector *c;
 	size_t N_deleted, total_deleted;
 	size_t w;
 
-	pc = (prune_context *) xalloc (sizeof(prune_context));
-	pc->power_cost = 0;
-	pc->null_links = (opts->min_null_count > 0);
-	pc->N_changed = 1;  /* forces it always to make at least two passes */
+	pc.power_cost = 0;
+	pc.null_links = (opts->min_null_count > 0);
+	pc.N_changed = 1;  /* forces it always to make at least two passes */
 
-	pc->sent = sent;
+	pc.sent = sent;
 
 	pt = power_table_new(sent);
-	pc->pt = pt;
+	pc.pt = pt;
 
 	free_later = NULL;
 	N_deleted = 0;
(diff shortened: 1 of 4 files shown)


link-grammar/parse/prune.c
```diff
@@ -213,16 +216,19 @@ static void put_into_power_table(C_list * m, unsigned int size, C_list ** t,
...
 /**
- * Allocates and builds the initial power hash tables
+ * Allocates and builds the initial power hash tables.
+ * Each word has 2 tables - for its left and right connectors.
+ * In these tables, the connectors are hashed according to their
+ * uppercase part.
+ * In each hash slot, the shallow connectors appear first, so when
+ * matching deep connectors to the connectors in a slot, the
+ * match loop can stop when there are no more shallow connectors in that
+ * slot (since if both are deep there cannot be matched).
```
linas: type: `there cannot`

linas:
looks good.

link-grammar/parse/prune.c
linas: I just hacked this with the web editor.

linas:
So. Stupid computer tricks. Just to see how things are going, I thought to measure the speed of the ten-year-old version 4.2.5 ... had to do some hacking to get past a broken file-search-path design. I was shocked to find that it is 3x faster. 

Let that sink in. Basically, `4.0.dict` is much much more complex today, than it was back then, so that its at least three times slower.  Of course, back then, corpus-fixes.batch had 2654 errors vs 385 errors today.  So the complex dictionary is a huge impact on performance.

Lets try a different experiment: run today's parser with the old dictionary. Easy to say. Hard to do. For multiple reasons. Conclude: today's parser is almost 2x faster than the old parser.

What to conclude?  The "almost 2x faster" result is ... great! But that also means that the more complex dict is 6x slower  (since 6 = 2x3).  So what does this mean? Well, part of me wants to say that this is like the fractal "length of the coastline of England" -- the more complex the dictionary, the more fractal language coastline can it cover, more accurately.  But that's just hand-wavey meaningless blather. 

Its still interesting, that a more accurate dictionary might even result in exponentially slower parsing? Is that possible?

ampli:
Yes. It seems it is more subtle than what it seems at first glance.
Among other things, the 4.2.5 dict is so "not supportive" relative to the current dict, that after the prune stage many sentences have at least one  word with **zero** disjuncts, a thing that cause a very fast parse failure.

I made the following tests:

Version 4.2.5:

`sed 's/verbosity=1/verbosity=9/' ../master/data/en/corpus-fixes.batch | env LD_LIBRARY_PATH=link-grammar/.libs link-grammar/.libs/grammar-parse | grep -A1 'After pp_pruning' | grep -cF '(0)'`
```
2654 errors.
Freeing dictionary en/4.0.dict
Freeing dictionary en/4.0.affix
2772
4.522u 0.536s 0:04.98 101.4%	0+0k 0+8io 0pf+0w
```

So 2772 sentences fail immediately due to at least one word with 0 disjuncts.

For the current version (5.2.1):ed '/verbosity=/d' data/en/corpus-fixes.batch | link-parser -v=9 -debug=prune.c | grep -F -A1 'After pp_prune' | grep -cF '(0)'`
`
```385 errors.
link-grammar: Info: Freeing dictionary en/4.0.dict
link-grammar: Info: Freeing dictionary en/4.0.affix
+ rc=0
master: 13.12u 0.22s 0:13.38e 189300Kb
190
```
But as you can see for yourself (if you change `grep -cF` to just `grep -F`) is absolutely most of the times due to unit alternatives (e.g. `did` that gets broken to `d.u(8) id.u(0)`), and the sentence still has to have a costly parse because there are actually no words with 0 disjuncts.

Another thing to consider is parse with FAT links.  All the complex (and parse-costly) conjunction links didn't exist on 4.2.5. I don't know much on FAT links, but it may be that the effective sentence length was smaller. Better to compare to 4.8.6 (it is slower than 5.5.1 on the batch benchmarks when even each version uses its own dict).

As I noted in #250, the current aggressive unit split, that didn't exist on 4.2.5 and even not in 4.8.6, causes much slowness. As I also pointed out there, this aggressive split is actually mostly useless because it mostly not supported by the dict. The said aggressive split (that you once added) adds unit alternatives for many words,  e.g. `mall -> ma l l`,  `did -> d id`, `TNT -> T N T`, `As -> A s`, `All -> A l l`. But all of that add **much** parsing overhead for nothing (as multi unit, beside 1 or 2 exceptions are not supported by the dict).

Proposal: Turn off multi unit split (or add support for them, e.g. that they will be handled as a single unit).

> more accurate dictionary might even result in exponentially slower parsing? Is that possible?
 
Several times I noticed  a large slowness due to what seemed as small dict fixes.

ALSO: No regexes in 4.2.5!

Last thing: 5.5.1 using the 4.2.5 dict (after adding an empty `4.0.regex` and `<UNKNOWN>` is slightly faster on both the old and current batches. However, something with the handling of capital words seems totally different so I don't know if this comparison is valid.

ampli:
I fixed some typos in my previous post so you may want to read it on the web (and not email).

linas:
It doesn't work with an empty regex; to get approximately backwards compat behavior, you need at least this:
```
NUMBERS: /^[0-9][0-9]?(:[0-9][0-9]?(:[0-9][0-9]?)?)?$/
NUMBERS: /^[0-9.,-]*[0-9](\+\/-[0-9.,-]*[0-9])?$/
NUMBERS: /^[0-9,.]*[0-9]$/

PL-CAPITALIZED-WORDS:  /^[[:upper:]].*[^iuoys'’]s$/

CAPITALIZED-WORDS:     /^[[:upper:]][^'’]*[^[:punct:]]$/

ING-WORDS:        /^\w.+ing$/
S-WORDS:          /^\w.+[^iuoys'’]s$/
ED-WORDS:         /^\w.+ed$/
LY-WORDS:         /^\w.+ly$/

UNKNOWN-WORD: /^[.,-]{4}[.,-]*$/
```

linas:
Yeah, maybe turning off aggressive unit stripping is a good idea. Or maybe making it into a flag. At one point, I was trying to parse biochem papers, and at that time, it seemed like a good idea.

linas:
More generally, if a sentence does not parse quickly and easily, what is the fallback strategy? Currently its ad-hoc:
* Try skipping one or more words
* Use panic-options on timeout.

Should we have unit stripping turned off by default, and then try aggressive unit stripping before trying to ignore some words? Maybe not a bad idea...

 Then an oldie-but-a-goodie: what about newspaper headlines, which are short of determiners?

ampli:
I checked again the overhead of multi-unit stripping, and at least in the fixes batch it is now only a few percents (maybe due to the automatic subscript addition that was recently added).  I guess that for very long sentences it can be more.

But I'm for leaving the multi-split as is, and instead fixing the dict to use it.
For example, `Nm` gets split to `N m`, but the dict doesn't recognize it (currently no `Nm.u`, so it can be used as a test case).
(BTW, `Kgm` is getting split wrongly to `K g m` instead of `Kg m` - seems like a bug that can be fixed).

I thought to use it as a test case of "general idiom definitions".
Something like:
```
<length-unit>: m.u OR ft.u OR ...;
<force-unit>: N.u OR Kg.u. OR ...;
<torque-unit>: <force-unit>_<length-unit>;
<UNITS>: <units-suffix> OR <torque-unit>;
LENGTH-LIMIT-1: ID*+;
```
And maybe also something like that, if the split will be done even more aggressively:
```
<torque-unit>: <force-unit>_<length-unit> OR <force-unit>_*_<length-unit>;
```
(The `*` above can be also a macro of possible unit separating tokens.)


> Then an oldie-but-a-goodie: what about newspaper headlines, which are short of determiners?

#402 is in a WIP stage. I based it on an implementation of sub-expression tags, when the tags in that case are used for per-dialect costs. (They can also be used for other expression manipulations like the dict-capitalization.)

We will also need to find out how to set the dialect if a sentence is not parsed. For example, how can we know if the reason is that it is short of determiners? An easy case may be if there are no determiners at all.

### **But**:
We can also always parse with a high disjunct_cost (say disjunct_cost=10) and implement a cost cutoff as a postprocessing. The cost_max for this cutoff will still be usually the normal one (2.7), and the higher-scored sentences will be considered only if there are no ones <= cost_max. Such a strategy may also **automatically identify the dialects** (by looking back from which sub-expressions the chosen_disjuncts got generated).
I tested running the fixes batch with `-cost-max=10` and the overhead is "only" 25-30%, with no more combinatorial explosions than with the default cost-max - only one in both cases. (Some more optimizations may reduce this performance gap.)

linas:
I like all these ideas. However, my intent is to never hand-write dictionaries again, and just automate everything.  The automated data will have cost attached to everything, including morphology splits.  Sorry I don't have dicts for you to play with, I keep getting distracted by other things.

Issue #537, opened 2017-05-25
Low memory handling - unimplemented


s-e-r-g:
1) There are a lot of unchecked malloc/realloc/strdup...etc. 
    They cause crash on 32bit platforms very often.

2) xalloc/exalloc  does abort / exit  on low memory situation. But API users expect an error to be returned, not exiting the process.

linas:
Ah interesting. It certainly is very slow, on that sentence, and then shoots up to 6.6GB on my machine. So, yes, this is an example where the current dictionaries are lacking.

I have no plans on fixing this; I have other worries burning up my time. Are you a programmer? do you know anyone who programs? I'll accept any reasonably clean fix for this -- as long as it doesn't make a big mess or introduce obvious bugs, that would be fine. (well, also follow the indentation style; don't send sloppy, careless code ...)

ampli:
**1)**
I found the problem of the vast memory/slowness.
The linkage of this sentence has 4 null words.
It also includes many words with the  same LL links.
On the pruning pass for null_count=0, the LL+ links can only reach the next word, and much pruning results. However, the pruning that is done for null_count>0 is not effective for LL links because they can link far away.

Setting  the length of the LL links to 1 solves the problem.
Due to the overhead in that setup (a loop over a large number of disjuncts/connectors) it may be a good idea to set them only for the pruning step of null_count>0, and even then only for long enough sentences (or set connector length in another way, e.g when the connectors are created).

In the same occasion I can maybe add an affix-class for morphology links.
Is it reasonable to assume their connector length, if limited, is always 1?
Also, it may be a good idea to allow setting connector length also for individual links, not only link prefix.
So a flexible (and simple) definition scheme is needed.
I can just do something "reasonable", unless you have specific ideas regarding the definition format (or what we have to define at all).

**2)**
Regarding clean recovery from a memory allocation failure, it looks the easiest way to handle that without much cluttering the source code, is mainly by using longjump() (but malloc() in certain functions will still need special handling on failure). Is such a solution fine by you?

ampli:
BTW, long ago I did tests of setting the connector length of LL links and didn't find an improvement.
But I checked that on batch file benchmarks, in which it only introduce some slowness, because only null_count==0 is tried in these runs.

linas:
> a loop over a large number of disjuncts/connectors
Where is that loop?  It would be better if that loop was a part of the dictionary setup...

 > it may be a good idea to allow setting connector length also for individual links
[...]

PR #846, opened 2018-11-11
Long sentences performance breakthrough

ampli:
At last a major performance breakthrough. Very long sentences are not very slow any more!
See the comments in `preparation.c` for how this fix works.

This time it's not a few percents of speedup, it's even x10 to x100  for some sentences in the fix-long batch.

For example:
- The "When Sir William Petre ..." sentence that now takes 4970 seconds to fail parsing with null_count=0 (interactive link-parser), with this PR the null_count=0 step takes 45 seconds.
- The "In vivo studies..." sentence now takes 32 seconds to parse with null_count=2, but with this PR it takes only 2.3 seconds.
- The "The problem is" is currently 58 seconds (interactive, null_count=4). With this PR it's only 14 seconds for the same result.
- With this PR, the whole fix-long batch runs (without timeout) in 135 seconds.

For some reason this PR is less effective for Russian. The batch even runs ~1% slower.
However, the Russian sentence from issue #537, that took about one minute after the fix that is mentioned there and currently takes 48 seconds, with this PR it takes about 20 seconds. 

I had to add an `int` field to `Connector_struct`, and now it is > 32 bytes. Because the fix in this PR effects mainly long sentences, this addition slows down the basic batch run by ~1%. But it speeds up the fixes batch run by ~9% (due to some long sentences there).  Another potential problem that may need a fix is the connector pair hash distribution, which I suspect that it is far from ideal.

After applying this PR, the bottlenecks are different. Things that were not considered a bottleneck then show up. Also, I have some more ideas how to further speed up the classic parser. One of them is almost ready (I need to rebase it after you apply this PR).


ampli:
In the comments I mentioned "connector sequence suffix". I think a clearer term may be just "trailing connectors".

linas:
wow

link-grammar/string-id.h
```diff
@@ -0,0 +1,39 @@
+/*************************************************************************/
+/* Copyright (c) 2004                                                    */
+/* Daniel Sleator, David Temperley, and John Lafferty                    */
+/* All rights reserved                                                   */
```
linas: You should remove these three lines, and put in your name, instead. I'm not sure, but the I think the "all rights reserved" line is not necessary.

linas: Oh, I see, I guess you are thinking that this is a "minor" change to the older existing string-set code.  In that case, you can leave their copyright in place; but you should add your own name, as it is a fairly major revision.

linas:
This deserves an entry in the changelog; also, you should probably add `Copyright 2018 Amir` to `preparation.c`

linas:
Ah, OK, thanks. Slowdown was what, maybe half a percent? I presume most of the slowdown was a cache-miss penalty for `rw-re->length_limit`, dozens  if not over a hundred cycles.

ampli:
I don't know if a cache miss penalty is involved, since if `re->length_limit` is not examined here, it will be anyway examined shortly at the start of `form_match_list()`, and if it examined here and didn't have an effect, its check in `form_match_list()` will most probably not be a miss. Instead I think the main penalty is the extra addition/subtraction and comparison, that are always done for nothing in short sentences.

>Slowdown was what, maybe half a percent? 

Something like that. It was actually a slight speedup on the fixes batch until the trailing connectors patch, that reduced the work in `do_count()`.

A 6x6 benchmark (6 "best of 6") of fix-long found a speedup of 0.3%.

PR #848, opened 2018-11-13
Fixes to PR #846 (the initial trailing connectors hashing implementation)

ampli:
I fixed 3 bugs in the initial implementation:
1. The word number encoding should have been prepended to each trailing sequence and not only the to full sequence.
2. Identical trailing connectors sequences of disjuncts that are derived from different alternatives, should have different suffix_id (the number that is used for hashing). This is because each may have different number of linkages.
3. "multi" connectors check was not done for "right" connectors. It didn't affect anything. (However, not making this check for the "left" connectors", which is already done, causes  a crash on many sentences).

This PR includes prepending the Gword number of each disjunct to the connector strings so different alternatives will get different suffix_is's. This fixes both (2) and (1). It fixes (1) because the Gword number is prepended to each connector string, and different words cannot share the same Gwords.

Fixing (2) doesn't add much overhead, since alternative tokens with identical trailing connector are apparently not common. But fixing (1) adds about 0-15% overhead for long sentences (they are still very much faster than originally...).

I'm not sure why these two bugs didn't reveal themselves as wrong linkages or even crashes.
With and without this fix, the "trailing connectors hashing " code produces exactly the same linkages as the  original code (tested by producing up to 250K linkages per sentence for each batch, sorting them to get a "canonical order") and comparing the first 25K ones). I think this may be because the hash table is still sized as originally but now it has much less entries, and also the word numbers are still used in the hash function (in order a to add more entropy) so the chances are low that two bad pairs will get hashed to the same bucket. I have some more related speedup ideas to try so I hope that the answer to this question will become clearer when I get to it more deeply.

A related addition is `-test=len_trailing_hash:SENTLEN` when `SENTLEN` is the minimum sentence length to use the "trailing connectors hash" method. The default is, as before, 36.
This makes it easier to validate that shorter sentences work fine too and give the same result, and that longer sentences give the same result with and without this method.

More fixes in this PR:
- Fix/update formatting / comments / debug messages.
- Consolidate the left/right loops.

### DIFF
--- a/link-grammar/parse/preparation.c
+++ b/link-grammar/parse/preparation.c
@@ -176,7 +176,12 @@ static void set_connector_hash(Sentence sent)
 	/* FIXME: For short sentences, setting the optimized connector hashing
 	 * has a slight overhead. If this overhead is improved, maybe this
 	 * limit can be set lower. */
-	if (sent->length < 36)
+	size_t min_sent_len_trailing_hash = 36;
+	const char *len_trailing_hash = test_enabled("len_trailing_hash");
+	if (NULL != len_trailing_hash)
+		min_sent_len_trailing_hash = atoi(len_trailing_hash+1);
+
+	if (sent->length < min_sent_len_trailing_hash)
 	{
 		int id = WORD_OFFSET;
 		for (size_t w = 0; w < sent->length; w++)
@@ -196,6 +201,8 @@ static void set_connector_hash(Sentence sent)
 	}
 	else
 	{
+		lgdebug(D_PREP, "Debug: Using trailing hash (Sentence length %zu)\n",
+		    sent->length);
 #define CONSEP '&'      /* Connector string separator in the suffix sequence .*/
 #define WORDENC_ADD 'A'
 #define MAX_LINK_NAME_LENGTH 10 // XXX Use a global definition.
--- a/link-grammar/parse/preparation.c
+++ b/link-grammar/parse/preparation.c
@@ -279,7 +279,7 @@ void prepare_to_parse(Sentence sent, Parse_Options opts)
 	size_t i;
 
 	build_sentence_disjuncts(sent, opts->disjunct_cost, opts);
-	if (verbosity_level(5))
+	if (verbosity_level(D_PREP))
 	{
 		prt_error("Debug: After expanding expressions into disjuncts:\n");
 		print_disjunct_counts(sent);
@@ -296,7 +296,7 @@ void prepare_to_parse(Sentence sent, Parse_Options opts)
 	}
 	print_time(opts, "Eliminated duplicate disjuncts");
 
-	if (verbosity_level(5))
+	if (verbosity_level(D_PREP))
 	{
 		prt_error("Debug: After expression pruning and duplicate elimination:\n");
 		print_disjunct_counts(sent);
--- a/link-grammar/parse/preparation.c
+++ b/link-grammar/parse/preparation.c
@@ -14,7 +14,7 @@
 #include "api-structures.h"
 #include "prepare/build-disjuncts.h"
 #include "connectors.h"
-#include "dict-common/dict-common.h" // For Dictionary_s
+#include "dict-common/dict-common.h"    // Dictionary_s
 #include "disjunct-utils.h"
 #include "externs.h"
 #include "preparation.h"
@@ -24,9 +24,9 @@
 #include "string-set.h"
 #include "string-id.h"
 #include "utilities.h"
-#include "tokenize/word-structures.h" // for Word_struct
+#include "tokenize/word-structures.h"   // Word_struct
 
-#define D_PREP 5 // Debug level for this module
+#define D_PREP 5 // Debug level for this module.
 
 /**
  * Set c->nearest_word to the nearest word that this connector could
@@ -183,11 +183,11 @@ static void set_connector_hash(Sentence sent)
 			{
 				for (Connector *c = d->left; NULL != c; c = c->next)
 				{
-						c->suffix_id = id++;
+					c->suffix_id = id++;
 				}
 				for (Connector *c = d->right; NULL != c; c = c->next)
 				{
-						c->suffix_id = id++;
+					c->suffix_id = id++;
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


link-grammar/parse/preparation.c
```diff
@@ -160,21 +174,37 @@ static void print_connector_list(const char *s, const char *t, Connector * e)
...
  * In order to save the need to cache the endpoint word numbers the
- * connector identifiers are not shared between words. To that end the
- * word number is prepended to the said strings.
+ * connector identifiers should not be shared between words. The initial
+ * implementation tried to do that by prepending the word number to the
+ * connector list. However, this is buggy, because trailing sequences
+ * should be unique per word. But this is fixed anyway by the alternatives
+ * check described below. (This bug left undetected because the hash
+ * tables are now relatively empty and also we still hash with word
+ * numbers - for added entropy).
```
linas: So did you test, by making the hash tables small, so that they are dense?  If your theory is correct, the bug should then become visible.

linas:
I'm thinking that I should wait on merging this, until you answer the question I just posed.  I can merge now, if you wish, I just thought it would be wise to ask.

link-grammar/parse/preparation.c
ampli: I said this only by analyzing how the "trailing connectors hash" works.
Since the word numbers are not checked on cache reads, their result depends solely on the endpoint hash values. On my initial implementation they were shared (unintentionally) by the same trailing sequences of deep connectors on different words, but each should have its own independent number of linkages.
The problem for different words can be solved by checking the words numbers too - at runtime -as was done originally. However, it is more efficient to one-time encode a different suffix_id number instead. The problem of tokens of different alternatives on the same word slot, that also clearly cannot share the exact same number of linkages, can only be solved by a different suffix_id
So a different suffix_id per Gword (which means a different word or a different alternative) solves both problem.

Just now I tried the current repository version with one-slot cache size (shift=0). Surprisingly, `ru` then runs only a few percents slower. But it gives the same 47 errors. However, when producing all the linkages, one sentence (the last one) gives a different result, which is strange:
`Found 2147483647 linkages (0 of 0 random linkages had no P.P. violations) at null count 5
`
Note only the `0 of 0` is strange, but this number and the null-count is different on every run!
`Found 565524126 linkages (0 of 0 random linkages had no P.P. violations) at null count 2
`
I tried it with no address randomization (a mode I use for debug) and it is still non-deterministic (only for this sentence). This needs a deeper investigation. With the default cache size it is deterministic:
`Found 2376 linkages (1000 of 1000 random linkages had no P.P. violations) at null count 7
`

I also tried the fixes batch with 1 hash slot. It runs about 50x slower...
One sentence , that get parsed with the default cache size, doesn't get parsed then (however it get parsed fine when this PR is applied):
```
linkparser> People ages 55 to 64 on average are out of work for more than 54 weeks, while those ages 25 to 34 are unemployed for about 36 weeks, according to figures from the Bureau of Labor Statistics.
...
Found 1745111394 linkages (0 of 0 random linkages had no P.P. violations) at null count 2
```
This is a similar problem to the one is `ru`, including the random number of linkages at a random null-count on each try... BTW, this sentence has no alternatives. So this may be due to the word number encoding problem (bug (1) in the initial message).

 Like the `ru` problem, this PR also fixes this problem.

The `he` and `lt` batches give the exact same detailed linkage results when 1-slot is used.

Beside of these 2 problems there are no apparent wrong results in the current repository version. But I already know from a large amount of Connector_table hacking that subtle caching bugs may manifest themselves very rarely, sometimes in a single sentence.

Because these 2 problems are fixed in this PR, and the theoretical considerations that see to me right, **I propose to accept this PR** even so it causes a slight slowness on long sentences. Of course I will think how to improve that, as this whole idea (of trailing connectors hashing) is only several days old (and seems to have more theoretical implications).

I will also investigate this strange `0 of 0 random linkages` problem (even though it will not exist if this PR is applied) because it may hint on something fundamentally wrong.

ampli:
One more thing - I said that I compared the detailed linkages of the batches.
This doesn't include the `People ages 55 to 64 ...` sentence, which has an exponential explosion of linkages and the random sampling result is very version-dependent so it usually doesn't match.
Regarding the Russian problem, I checked that the problematic sentence got parsed just fine in the same repository version (current one) in which it is problematic now.

This needs more checks. But I still think it is a good idea to apply this PR.

ampli:
> Regarding the Russian problem, I checked that the problematic sentence got parsed just fine in the same repository version (current one) in which it is problematic now.

Sorry for the multiple messages. 
The above is not relevant since the problems appear only with a small cache.

linas:
OK, thanks. Should the size of the hash table be decreased?

ampli:
> Should the size of the hash table be decreased?

Yes.
Prior to the "trailing connectors" implementation, I made a patch that makes a table size estimation according to the connectors (counting desperately shallow and deep ones) and uses the lesser of that and the estimation according to the current way. It led to a noticeable speedup.
---> I will try it again and send a PR if it helps. But see below for more longterm solutions.

Even prior to that I made the following working patches:
- Reduce the table hashed item size. Currently it is 44. It can be reduced to 8, with a simpler hash function and a less costly retrieval. This involves a hash table per connector(!).
- A dynamic hash table that can grow on demand. The initial sizing calculation just estimates its minimum size, to prevent unneeded grow steps that would occur if the initial size was 0.
- A combination of these (dynamic hash table per connector).

These patches can be ported to the current `count.c` and tried out. I hope that the relatively small savings of some past patches may turn to a larger saving, now that the actual hard work of do_count() has been reduced.

Some other interesting things can be done with the hash table. The original implementation didn't share the initial **null_count=0** with the " **null_count>0** step. The reason was that power-pruning was done again (because the **null_count=0** power-pruning is tighter), and then the connector addresses that were used for hashing got changed. But with the hash done using the `suffix_id` of connectors which is assigned prior to the power-pruning, it seems the tables can be shared. If this is indeed possible, parsing with **null_count>0** may most probably be noticeable faster.

I have even some more ideas and past patches to try out now.

ampli:
> I made the following working patches

I checked them again and they don't yield a significant speedup (but add a significant complexity).
I also tried to find a better formula for the table size, and using complex formulas could get a speedup of up to 2%.
But it very much depends on the sentences in the tested batch. So I think that for now there is no point to change the current simple formula.

> I have even some more ideas

I am trying to find out if there are "0 count" table entries that there is no need to save because they will not be reachable, or existing ones that became "unreachable" and can be dropped from the table. No results yet...
So for now I will continue with the non-related (to count.c) speedup patches.

PR #852, opened 2018-11-21
 mk_parse_set() speedup

ampli:
The main change here is:
- mk_parse_set(): Optimize Rmatch processing

In the profiling for fix-long. it speeds up `mk_parse_set()` by about 40% (from 13.9% to 7.9% of the total CPU).
Before this patch (legend at the end [**]):
`     242   1.8%  85.4%     1916  13.9% mk_parse_set`
After it:
`     146   1.2%  89.2%     1002   7.9%  mk_parse_set`

However, since it consists only a small part of total CPU time, the batch benchmarks show only small speedup (for some sentences the speed increase is big [*]):
``` text
 +0.2%    en/corpus-basic.batch
 +1.0%    en/corpus-fixes.batch
 +1.4%    ru/corpus-basic.batch
+26.8%    ../debug/prune/ex/en/bms.batch (In vivo studies...) [*]
 +4.3%    data/en/corpus-fix-long.batch
```

fix-long profiling after this PR (per file, only >0.1% are shown):
```
    6158  49.2%  49.2%     6162  49.2% link-grammar/parse/fast-match.c
    3137  25.0%  74.2%     9948  79.4% link-grammar/parse/count.c
    1223   9.8%  84.0%     1223   9.8% link-grammar/connectors.h
     694   5.5%  89.5%      994   7.9% link-grammar/parse/extract-links.c
     447   3.6%  93.1%     1093   8.7% link-grammar/parse/prune.c
     220   1.8%  94.8%      240   1.9% link-grammar/post-process/post-process.c
     115   0.9%  96.8%      115   0.9% link-grammar/parse/fast-match.h
     109   0.9%  97.7%      266   2.1% link-grammar/parse/preparation.c
      61   0.5%  98.2%       95   0.8% link-grammar/disjunct-utils.c
      43   0.3%  98.9%       58   0.5% link-grammar/string-id.c
      14   0.1%  99.2%       59   0.5% link-grammar/prepare/build-disjuncts.c
      14   0.1%  99.3%       45   0.4% link-grammar/utilities.h
       9   0.1%  99.4%       49   0.4% link-grammar/memory-pool.c
       7   0.1%  99.6%       55   0.4% link-grammar/utilities.c
```

So there is no much room for more speedup for long sentences (only maybe 20%) unless processing shortcuts will be found. The profiling for short sentences is totally different and there is more room there for speedup (>0.5% are shown):
``` text
     227  20.5%  44.5%      264  23.8% link-grammar/post-process/post-process.c
     131  11.8%  56.4%      294  26.6% link-grammar/parse/prune.c
      65   5.9%  62.2%      118  10.7% link-grammar/prepare/exprune.c
      51   4.6%  66.8%       51   4.6% link-grammar/connectors.h
      44   4.0%  70.8%       90   8.1% link-grammar/parse/count.c
      35   3.2%  74.0%      132  11.9% link-grammar/dict-common/dict-utils.c
      29   2.6%  76.6%       29   2.6% link-grammar/parse/fast-match.c
      27   2.4%  79.0%       53   4.8% link-grammar/disjunct-utils.c
      24   2.2%  81.2%      116  10.5% link-grammar/prepare/build-disjuncts.c
      21   1.9%  83.1%      160  14.5% link-grammar/parse/preparation.c
      20   1.8%  84.9%       28   2.5% link-grammar/parse/extract-links.c
      15   1.4%  86.3%       21   1.9% link-grammar/dict-file/read-dict.c
      11   1.0%  89.5%       15   1.4% link-grammar/post-process/pp_linkset.c
      10   0.9%  90.4%       45   4.1% link-grammar/connectors.c
      10   0.9%  91.3%       64   5.8% link-grammar/memory-pool.c
       8   0.7%  93.0%       40   3.6% link-grammar/utilities.h
       5   0.5%  94.5%      158  14.3% link-grammar/tokenize/tokenize.c
```
[**] Fields in the above:

1. Number of profiling samples in this function
2. Percentage of profiling samples in this function
3. Percentage of profiling samples in the functions printed so far
4. Number of profiling samples in this function and its callees
5. Percentage of profiling samples in this function and its callees
6. Name

### DIFF
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -434,10 +434,9 @@ Parse_set * mk_parse_set(Word* words, fast_matcher_t *mchxt,
 	RECOUNT({xt->set.recount = 0;})
 	for (w = start_word; w < end_word; w++)
 	{
-		size_t mlb, mle;
-		mle = mlb = form_match_list(mchxt, w, le, lw, re, rw);
+		size_t mlb = form_match_list(mchxt, w, le, lw, re, rw);
 		// if (mlist) mlist = sort_matchlist(mlist);
-		for (; get_match_list_element(mchxt, mle) != NULL; mle++)
+		for (size_t mle = mlb; get_match_list_element(mchxt, mle) != NULL; mle++)
 		{
 			unsigned int lnull_count, rnull_count;
 			Disjunct *d = get_match_list_element(mchxt, mle);
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -443,6 +443,7 @@ Parse_set * mk_parse_set(Word* words, fast_matcher_t *mchxt,
 			Disjunct *d = get_match_list_element(mchxt, mle);
 			bool Lmatch = d->match_left;
 			bool Rmatch = d->match_right;
+			bool ls_exists = false;
 
 			for (lnull_count = 0; lnull_count <= null_count; lnull_count++)
 			{
@@ -477,9 +478,13 @@ Parse_set * mk_parse_set(Word* words, fast_matcher_t *mchxt,
 						ls[3] = mk_parse_set(words, mchxt, ctxt,
 						              ld, d, lw, w, le, d->left,
 						              lnull_count, pex, islands_ok);
+
+					ls_exists =
+						ls[0] != NULL || ls[1] != NULL || ls[2] != NULL || ls[3] != NULL;
 				}
 
-				if (Rmatch)
+
+				if (Rmatch && (ls_exists || le == NULL))
 				{
 					rs[0] = mk_parse_set(words, mchxt, ctxt,
 					                 d, rd, w, rw, d->right->next, re->next,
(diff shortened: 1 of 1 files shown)


linas:
Huh. That was interesting.

When I read through your code changes, its very counter-intuitive. Naively, I figured that passing args on stack would be faster, and since there is a lot of memory-locality on stack, there would be no cache misses. The only overhead is to actually place arguments on stack, and this overhead can be minimized by placing the non-changing once first.  Allowing e.g. tail-recursion optimizations by the compiler.  So your changes suggest that my intuition is dead-wrong.  Interesting...

There are a few other places in the code, where lots of args are passed in the call, instead of being held in a struct. Perhaps those can be sped up, as well. 

Perhaps a really really smart compiler engineer could find some way of automatically doing this kind of optimization...

ampli:
Maybe you are not wrong. I compared the assembly output (using https://godbolt.org) and when I removed the `island_ok` argument I could not find a difference in the `mk_parse_set()` call sequence. However, the number of total instructions in `mk_parse_set()` got reduced by 7 (I could not get a clear picture about the nature of this reduction with a fast glance at the `diff`, since there are some additions and some removals at different places).
This removal and the removal of "words" gave me a small (but consistent) speedup in 7x7 benchmaks of fix-long. However, I observed a slowdown when In tried to remove the `fast_matcher_t` argument.

Absolutely most of the speedup was from the first commit.

linas:
Ah. Compilers always have a cutoff: "if fewer than N arguments do this, else do that", together with variants for N occupying different numbers of registers, different number of bytes on stack, etc. If N is too large, everything gets copied to stack; if the argument order is "wrong" then it has to copy just to re-arrange the order into the correct, desired order.  There's a lot of mov, push, pop silliness that comes from that, time-wasting reshuffling of arguments.

PR #853, opened 2018-11-25
Trailing connector hash - bug fix: It should be unique per disjunct cost

ampli:
Bogus linkages (having wrong costs) have been observed due to this problem after another fix (that will be sent soon). 

An example sentence that has trailing connector sequences (on the same word) with different costs is:
`The fact that he smiled at me gives me hope`

In this PR also:
- Update some comments.
- Change`-test=len_trailing_hash` to use dashes like other test strings.

FIXME: A constant 20 is used in `set_connector_hash()` for max. connectors per disjuncts. In `lisjuncts.c ` a constant 20 is used for max. number of links per word. Maybe use a common symbolic definition (need to find a name for it).

### DIFF
--- a/debug/README.md
+++ b/debug/README.md
@@ -171,9 +171,9 @@ and [msvc/README.md](/msvc/README.md).
 
 5) Test the "trailing connector" hashing for short sentences too (e.g. for
 all sentences with more than 10 tokens):
-`link-parser test=len_trailing_hash:10`
+`link-parser test=len-trailing-hash:10`
 Or optionally (in order to see relevant debug messages from `preparation.c`):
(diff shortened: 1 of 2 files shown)


ampli:
@linas,
Please don't apply this PR for now, since I still observe wrong costs.
I hope to have a fix by tomorrow.

linas:
ok

ampli:
I could not reproduce the bad cost output, and I'm not sure this fix of different suffix_id for different costs is really needed. I am continuing to investigate this case and most probably I will change this PR.

ampli:
It turns out the different suffix_id for connectors on disjuncts with different codes,, is not actually needed.
I am closing this PR and will send another one.

PR #855, opened 2018-11-27
Revert "set_connector_hash(): Bug fix: Consider disjunct cost too" from PR #854

ampli:
Revert "set_connector_hash(): Bug fix: Consider disjunct cost too" ￼… from PR #854 .
This reverts commit 9ba850c.

It was in the PR that I cancelled, and somehow it got entered again into PR #854 when I cherry-picked commits from various branches.

It turned out this change is unneeded in the current code.
This is because the linkage count between endpoint connectors doesn't depend on the cost of their disjuncts. It was added in a try to support using suffix_id hushing in extract-links.c, a thing that has not been implemented.

Anyway, I compared the en&ru basic, and en fixes batches detailed linkages (including parsing with nulls) before and after making this revert, and the results are exactly the same. This ensures the reverted code is indeed unneeded. (BTW, the detailed linkages file for the fixes batch is ~500MB.)


linas:
:-)

Issue #860, opened 2018-11-29
An additional speedup improvement for long sentences - with a catch


ampli:
I tested a proof-of concept optimization that yields ~60% speedup for the fix-long batch.
Using it, the most CPU consuming sentence in it ("And yet [...]" - total of 174 tokens) parses for me at about 31 seconds (but more optimizations in the framework of this idea are still possible).
The linkages of the (ru and en) basic, and fixes batches found to be the same (I cannot test all the linkages of the long sentences in fix-long since they have combinatorial explosion).

The code has these main problems:
1. It is not so "nice" and it clutters do_count() (mainly at 3 points).
2. It may be hard to verify it by reading.

It can be optimized very much more, but it may become even less readable.

I guess you may want to see it and comment on it. This may help me to finish it if desired.
It is still in the proof-of-concept coding style, but if desired I can send it as a PR that is not to be applied.

ampli:
I sent my WIP as PR #861.
Here is the idea:

The fast-matcher brings us a list of disjuncts from a middle-word, that match both of the given left and right connectors. Most of these lists contribute nothing to the intermediate linkage counts, because at least one of the sides is often "not grammatical" and yields a count of 0.

Due to the nature of the disjuncts, many of them contain the same connectors, and even the same connector sequences.
We can observe that (this description uses `le` but for `re` it is similar, with one addition [*]):
1. In the fast matcher, the same endpoint connector will always select the same candidate disjuncts.
2. Between 2 certain words, the same trailing connector sequences (all end, of course, with the same connector), will have the the same linkage count on the next level of do_count(). To see that, let's see one them:
`do_count(ctxt, lw, w, le->next, d->left->next, lnull_cnt)`
`lw` and `w` are the 2 certain words, `le` is the endpoint connector, `d` is a selected disjunct.
Since the same disjuncts will be selected by the same endpoint trailing connector sequences and only their next connectors of each are used here, the returned count will be always the same. This is true for the rest of the do_count() calls that use the endpoint connectors.
3. If all the counts due to a particular endpoint trailing connector sequence are 0, there will never be any total count when such an endpoint is tried. This fact can be cached. Note that no selected disjuncts at all, which is very common, is cached here too as 0. Also note that the other optimizations (at the first level only) must be disabled in order to ensure we really have 0 count.
4. The triple (lw, w, le) has much less combinations than the currently cached counts by (lw,rw,le,re), so it is much more scalable to use it as a key for this caching.
[...]

PR #861, opened 2018-11-30
Add fast-match 0-count caching

ampli:
@linas,
Here is the speedup modifications (issue #860). **This code is ONLY for review.**

For clarity, I committed most of the  debug printouts (a few was left by mistake in the main commit).
Simple additions of some more optimizations are possible, but I haven't added them yet because this code is enough to demonstrate the idea.

I will post the principles of this speedup on issue #860.

### DIFF
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -515,6 +515,10 @@ static Count_bin do_count(
 			if (le != NULL)
 			{
 				le_w_count = is_lcount(ctxt, lw, w, le);
+#if 00
+				fprintf(stderr, "CHECK IS_LCOUNT: suffix_id %d w %d: %d (SKIP %d)\n",
+					 le->suffix_id, w, le_w_count, 0 == le_w_count && NULL == re);
+#endif
 				if (le_w_count == 0)
 				{
 					if (re == NULL) continue;
@@ -528,6 +532,10 @@ static Count_bin do_count(
 			if (re != NULL)
 			{
 				re_w_count = is_rcount(ctxt, w, rw, re);
+#if 00
+				fprintf(stderr, "CHECK IS_RCOUNT: suffix_id %d w %d: %d (SKIP %d)\n",
+					 re->suffix_id, w, re_w_count, 0 == re_w_count && NULL == le);
+#endif
 				if (re_w_count == 0)
 				{
 					if (le == NULL) continue;
@@ -549,14 +557,30 @@ static Count_bin do_count(
 			size_t mle;
 			for (mle = mlb; get_match_list_element(mchxt, mle) != NULL; mle++)
 				if (get_match_list_element(mchxt, mle)->match_left) break;
+#if 00
+			fprintf(stderr, "ID %d: le->suffix_id %d w %d: %d\n",
+					  id, le->suffix_id, w, le_w_count);
+#endif
 			if (get_match_list_element(mchxt, mle) == NULL)
 			{
 				store_le_w_count(ctxt, lw, w, le, 0);
+#if 00
+				fprintf(stderr, "ID %d: le->suffix_id %d w %d: STORE1 %d\n",
+			        id, le->suffix_id, w, lcount_found);
+#endif
 				pop_match_list(mchxt, mlb);
 				continue;
 			}
 		}
 #endif /* LCOUNT_CACHE */
+#if 00
+		else
+		{
+			if (le != NULL)
+			fprintf(stderr, "ID %d: suffix_id %d w %d: %d\n",
+						  id, le->suffix_id, w, le_w_count);
+		}
+#endif
 
 		for (size_t mle = mlb; get_match_list_element(mchxt, mle) != NULL; mle++)
 		{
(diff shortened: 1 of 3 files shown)


linas:
I looked at it, but its late at night and I'm not sure what I should be looking for. It looks plausible. You're writing pretty clean code. Is there something in particular I should look at or think about?

linas:
> Currently the length_limit is assigned only per connector string.

I think its reasonable to assume that this will hold in the future. If one wanted more subtle results, the cost could be (should be) made some function of the length. Also, if the learning code needs to have different effective lengths, its probably better just to declare a new link type, rather than forcing one link type to behave two different ways (depending on the disjunct).  So I don't think we loose anything because the length-limit is per-connector.

link-grammar/parse/count.c
```diff
@@ -630,6 +775,15 @@ static Count_bin do_count(
 						}
 					}
 				}
+				if (need_re_w_count && !rcount_found && (0 != hist_total(&l_bnr)))
```
linas: It seems to me that you do not actually need to test `&& (0 != hist_total(&l_bnr))` here, unless you are trying to somehow prevent `rcount_found` being set to true.

linas: Oh. Maybe this is the point-2. that you are making above? Hmmm.

ampli:
Regarding `length_limit`, here is  a current potential problem in both optimizations:
In` do_count()`, say the endpoint connectors are le=**A+** (word lw) and re=**B-** (word rw) and they have **L_suffix_id** and **R_suffix_id**. The suffix_id trick assumes that all pairs with the same suffix_id's have exactly the same count.

Let's look on two left connectors (1 and 2) **A+**, each having the same **L_suffix_id**.
The LHS of their disjuncts is:
1: X+ A+ B+
2: Y+ A+ B+

Now say X+ has some `length_limit` and Y+ has an unlimited length, and the `length_limit` of X+ is less than the word distance rw-lw.   Then the count for connector 1 will be 0, while for connector 2 it may be non-zero.

I didn't find different linkages in my detailed linkages comparison to old versions (with the new dict and corpuses), so I assume that the dict don't contain such expression, or something in my analysis above is flawed.

ampli:
> The LHS of their disjuncts is:

RHS

linas:
> this code is very subtle

Yes, maybe too subtle for me, at the moment.  I recall trying to fiddle with it, once, long ago, and came up empty-handed.  I don't particularly want to spend an afternoon pondering it, if you already are.

ampli:
> 1: X+ A+ B+
> 2: Y+ A+ B+

My using of B+ here may be misleading, since it is not related to the example's endpoint **re=B-**.
So for the purpose of my argument it could be as well:

1: X+ A+ C+
2: Y+ A+ C+

linas:
> Now say X+ has some length_limit and Y+ has an unlimited length, and the length_limit of X+ is less than the word distance rw-lw. Then the count for connector 1 will be 0, while for connector 2 it may be non-zero.

OK, that I understand. On the one hand, I can see how that is a valid concern. On the other hand, as I stare at the  `UNLIMITED-CONNECTORS` list, its hard to imagine how any of those would ever be `or`'ed with a limited connector.   That is, I'm not surprised that the dict does never contains `(X+ or Y+) & A+ & C+` .. of course, I cannot prove this either.  Nor can I guarantee that such a construction might appear someday; otherwise, you could check  if this is the case, when the dictionary is loaded.

link-grammar/parse/count.c
ampli: We need to find out if the RHS that matches `re` may **ever** have any count that can be multiplied by an LHS count (if exists). Here we know that there is no count due to Rmatch (since ` !rcount_found`.  But maybe there is a count due to  `l_bnr`.  `If `(0 == hist_total(&l_bnr)` then we know there is no such count, so this calculation is skipped. If `(0 != hist_total(&l_bnr)` it may be that the count exists (and the cached `l_bnr` will be used by the `CACHE_COUNT` macro) or it is `NO_COUNT` (and the macro will call `do_count()` to find it in the hard way).

Note that only `rcount_found` is set in case there is an `l_bnr` count, and the multiplication/accumulation is not done. This is because currently this check is performed only when `le == NULL`, and in that case there cannot be a total count ("subtle" we already said).
``` c
    544                                 else if (re_w_count == -1 && le == NULL) need_re_w_count = true;
```
Also, since we only need to know in this case and others if `rcount_found`, a further optimization could be done if a special argument of `do_count()` to tell it to return immediately on positive count (without continuing to calculate the count). But such optimizations start to be unmanageable...).

Applying this PR to the current master (with an added argument to `form_match_list()` to disable the just added "lc optimization" when calculating the cached `le_w_count`) yeilds **37% speedup of the fix-long** run. The total runtime for me is 75 seconds, about half of it is due to the `And yet [...]` sentence (with still has a panic on the standard timeout because it takes 32 seconds).
I also validated that the linkages are exactly the same as without this PR (checked with `-test=len-trailing-hash:0` in order to generate suffix_id's for all the sentences).

But despite of the big speedup, I still propose not to use this idea yet. I have more ideas to test, and I hope to get a deeper understanding when testing them.

linas: OK.

ampli:
> Regarding `length_limit`, here is a current potential problem in both optimizations:

>1: X+ A+ B+
>2: Y+ A+ B+

On a further analysis it seems this is not a problem, since `X` and `Y`are "outer" to `A+ & B+`, and only the "inner" connector sequence determines the count.

In any case, as said in issue #860, I am closing this PR.

PR #863, opened 2018-12-01
form_match_list() shortcuts

ampli:


ampli:
(I sent it w/o description by mistake...)

This simple fix speeds up the long-batch run by 30%.

The main fix here is:
- form_match_list(): Optimize for no match of lc!=NULL

It avoids extra processing in cases that would not yield a count in `do_count()`.

In `do_count()`, if there is a right-match count, a special left side counting is done if `(le == NULL)`,
to be multiplied by the right-match count in order to yield the total count.
It means that if (le  != NULL)` this special counting will not be done, so the total counting will be 0 if the left-count is 0. And we know the left-count will be 0 if there is no left-match.

The corresponding shortcuts are implemented in 3 places in `form_match_list()` (in this order),
all **after checking** that `(lc != NULL)`:
1. If `lc` doesn't select any disjuncts, return an empty list.
2. if `lc` selects disjuncts, but none of them matches, return an empty list.
In principle this covers case (1), but no speed tests have been done without it.
3. if `rc` selects a disjunct that `lc` didn't match, skip this disjunct without trying to match `rc`.

The second shortcut is:
- form_match_list(): Optimize for no matches for both lc and rc

This yields only a small speedup.

The 3rd commit (Add get_match_list_position()) is a cosmetic one.

The last commit (Defer generating the unique debug id) saves small processing in the common case in which there are no selected disjuncts at all. It is still enabled to make sure the list is not corrupted due to the optimizations that are being added.

The down side of this PR is a slight slowness (~0.3%) in the `en` basic batch run.

**Notes:**
1. Additional similar optimizations (most probably with a noticeable speedup) can be done for `rc` too, and for `(lc==NULL`). However, they involve a Table_connector lookup, which may look like "modularity violation". But it it seems fine for you I can implement that too.
2. These is an overlap in this optimization and that of PR #860, so their combination doesn't yield the multiplication of speeds up (1.3x*1.6x =  ~ 2x) but only ~70%. However, in the method of PR #860 there seems to be a big room for improvements.

ampli:
> 1. Additional similar optimizations (most probably with a noticeable speedup) can be done for `rc` too, and for `(lc==NULL`). However, they involve a Table_connector lookup, which may look like "modularity violation". But it it seems fine for you I can implement that too.

I tested 3 possible uses of using `Table_connector` lookup in` form_match_list()`. Only one of them (for `(mx->d->right == NULL) || (rc == NULL)`) yields a speedup - of about 2% for long sentences. It probably doesn't worth the added complexity (6 added lines in `form_match_list()` that call `table_lookup()` + an added argument).

PR #864, opened 2018-12-01
en HYPHENATED-WORDS: Don't match words ending by punctuation

ampli:
See [my post](https://github.com/opencog/link-grammar/issues/860#issuecomment-443070264) in PR #860.

### DIFF
--- a/data/en/4.0.regex
+++ b/data/en/4.0.regex
@@ -246,6 +246,7 @@
 % Explicitly call out (5'|3') so that we don't all a generic match to 'll
 %  /^[[:alnum:]][][:alnum:],:.\[-]*-[][:alnum:],:.\[-]*[[:alnum:]]$/
 <HYPHENATED-WORDS>: !/--/
+<HYPHENATED-WORDS>: !/[[:punct:]]$/
 <HYPHENATED-WORDS>:
   /^([[:alnum:]]|5'|3'|2'|N')([][:alnum:],:.()[-]|5'|3'|2'|N')*-[][:alnum:],:.()[-]*[[:alnum:]]*$/
 


ampli:
This correction (which I think is right) has the following effect:
Test sentence (from the fixes batch): `You want part no. 1234-567.`

Current parse:
`LEFT-WALL you want.v part.n no.x . 1234-567.[!].n`
(I think it is incorrect to include the ending dot in a part number.)

But after this PR, this sentence is not parsed:
`LEFT-WALL you want.v part.n [no.] 1234-567[!<NUMBERS>] .`

The problem that `1234-567` is getting caught by `<NUMBERS>` and currently only one regex can match.
Adding `<NUMBERS>` to the `<HYPHENATED-WORDS>.n` entry doesn't seem the right solution, since `<NUMBERS>` match things that don't look like part numbers, and things that validly get ` `<HYPHENATED-WORDS>.n` may be meaningless with numbers (I guess).

So I don't have a proposal for a dictionary fix to support part numbers that are subset of `<NUMBERS>`.
The only thing that I can think of is a regex flag to allow adding alternatives by another regex, something like (`c` for continue):
`<NUMBERS>: /^[0-9.]+-[0-9]+$/c`

linas:
Thanks. Part numbers can be things like `a42-b63-asdf-1999`.

linas:
Hmm. HYPHENATED WORDS should probably match `1234-567` maybe.

ampli:
> Hmm. HYPHENATED WORDS should probably match `1234-567` maybe.

We can get something like that by adding:
`<NUMBERS>: !/^[0-9]+-/`

But then we get (from the fixes corpus) (the last one is a commented out sentence):
```
LEFT-WALL it was.v-d the 1989-90[!<HYPHENATED-WORDS>].a crop.n that.j-r failed.v-d .
LEFT-WALL the plane.n is.v a 747-400[!<HYPHENATED-WORDS>].n
LEFT-WALL people.p aged.v-d 20-40[!<HYPHENATED-WORDS>].a don't know.v how to.r use.v the word.n " aged.v-d " .
```

If this is nevertheless fine I can send a PR.

linas:
Ugh. `1989-90` is a date range while `20-40` is a numeric range and `747-400` is a part number. I don't want to mess with this stuff, right now.

Compare 
`There are 12 - 15 ml of acid in the bottle`  to
`There are 12-15 ml in the bottle`

ampli:
But currently it is not much better:
```
LEFT-WALL it was.v-d the 1989-90[!<NUMBERS>] crop.n that.j-r failed.v-d .
LEFT-WALL the plane.n is.v [a] 747-400[!<NUMBERS>]
```
Note that `1989-90` is not identified as date range, and `747-400` is not identified as a part number...
(It was "always" so and didn't get affected by this PR (#864).)

I think that the only thing that can fix that is allowing multiple regex matches, which I think may be the right solution if we implement an additional cost cutoff processing I once mentioned. Maybe we can return to that after I finish to send PRs for my existing pending changes and WIP (which may take a while).

PR #868, opened 2018-12-22
Use suffix_id instead of connector address in x_table

ampli:
This change solves a very high CPU and memory usage in `mk_parse_set()` for long sentences,
especially those with several nulls.

Main changes:
- extract-links.c: Use suffix_id instead of connector address
- extractor_new(): Tune the number of x_table entries

This reduces the number of entries in the x_table to a small fraction
of the currently needed ones.

However, since disjuncts of connectors with the same suffix_id may have
different costs and may belong to different words, the disjunct cost
and word_string have also to be used as keys.

The hash and lookup functions now seems complex, but they take a negligible time of the total.
This change speeds up the fix-log batch by ~3%.

I ran detailed linkage run (including nulls) of the basic corpus of 'en`, `ru`, `lt`, `he`, and 'de',  on the code before the suffix_id change, after it, and after this PR, with these parameters,:
`-disjunct -test=auto-next-linkage:25000 -limit=81111 -test=len_trailing_hash:0`

`-disjunct`: Validate chosen disjuncts, including their cost.
`-test=auto-next-linkage:25000`: Print up to this number of linkages.
`-limit=81111`: Use a high limit to consider all the linkages (2 sentences with a very high number of linkages have been omitted from the fixes batch).
`-test=len_trailing_hash:0`: Use trailing has for all the sentences, to validate it. I also tested with `-test=len_trailing_hash:254` (no trailing hash).

The only difference was for the sentence:
`Although society has excluded the patients in the ward for their unique qualities, they feel 'safer' trying to fit in because they receive approval from nurses and the representatives of society.`
``` text
7516022c7516022
<                feel.v    1.000  CO- Wi- O+ @MV+ @MV+ 
---
>                feel.v    1.000  CO- Wi- O+ @MV+ 
7517311c7517311
<                feel.v    1.000  CO- Wi- O+ @MV+ @MV+ 
---
>                feel.v    1.000  CO- Wi- O+ @MV+ 
7517359c7517359
...
```

It looks like there are two @MV connectors on some disjuncts of `feel`, and now the !disjunct report is more accurate. I didn't actually try to inspect the real disjuncts so I'm not sure. I compared to 5.5.1 (before my `lisjuncts.c` change), which reports only one @MV.

### DIFF
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -190,7 +190,10 @@ extractor_t * extractor_new(int nwords, unsigned int ranstat)
 	pex->log2_x_table_size = log2_table_size;
 	pex->x_table_size = (1 << log2_table_size);
 
-	//printf("Allocating x_table of size %d (nwords %d)\n", pex->x_table_size, nwords);
+	DEBUG_X_TABLE(
+		printf("Allocating x_table of size %d (nwords %d)\n",
+		       pex->x_table_size, nwords);
+	)
 	pex->x_table = (Pset_bucket**) xalloc(pex->x_table_size * sizeof(Pset_bucket*));
 	memset(pex->x_table, 0, pex->x_table_size * sizeof(Pset_bucket*));
 
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -177,16 +177,16 @@ extractor_t * extractor_new(int nwords, unsigned int ranstat)
 	pex->rand_state = ranstat;
 
 	/* Alloc the x_table */
-	if (nwords >= 24) {
-		log2_table_size = 14 + nwords / 24;
+	if (nwords > 96) {
+		log2_table_size = 15 + nwords / 48;
 	} else if (nwords >= 10) {
-		log2_table_size = 14;
+		log2_table_size = 14 + nwords / 24;
 	} else if (nwords >= 4) {
-		log2_table_size = nwords + 1;
+		log2_table_size = 1 + 1.5 * nwords;
 	} else {
-		log2_table_size = 4;
+		log2_table_size = 5;
 	}
-	if (log2_table_size > 21) log2_table_size = 21;
+	/* if (log2_table_size > 21) log2_table_size = 21; */
 	pex->log2_x_table_size = log2_table_size;
 	pex->x_table_size = (1 << log2_table_size);
 
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -221,7 +221,7 @@ void free_extractor(extractor_t * pex)
 		}
 		DEBUG_X_TABLE(
 			if (c > 0)
-				printf("I %d: chain %d\n", i, c);
+				;//printf("I %d: chain %d\n", i, c);
 			else
 				N++;
 		)
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -47,7 +47,9 @@ struct Parse_set_struct
 {
 	short          lw, rw; /* left and right word index */
 	unsigned int   null_count; /* number of island words */
-	Connector      *le, *re; /* pending, unconnected connectors */
+	int            l_id, r_id; /* pending, unconnected connectors */
+	double         l_cost, r_cost;                 /* disjunct costs */
+	const char     *l_word_string, *r_word_string; /* disjunct word strings */
 
 	s64 count;      /* The number of ways to parse. */
 #ifdef RECOUNT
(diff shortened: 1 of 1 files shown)


linas:
Thank you!

ampli:
>```
><                feel.v    1.000  CO- Wi- O+ @MV+ @MV+ 
>---
>>                feel.v    1.000  CO- Wi- O+ @MV+ 
>```
This duplicate multi connector problem repeats in some other sentences too, and it is due to a bug (will be fixed).

But I found a more severe potential problem in this PR: The connectors in `chosen_disjuncts` are sometimes rather incorrect (in <<99.9% of the cases, but this is still a lot of cases). This is hidden because from some reason that I don't understand `!disjuncts` and also ` linkage_get_disjunct_str()` rebuild the disjuncts from `link_array` instead of just using chosen_disjuncts(done by `lg_compute_disjunct_strings()` in `lisjuncts.c`). Do you have an idea why?

Also, `issue_link()` (`extract-links.c`) builds both `link_array` and `chosen_disjuncts`, while it could build only one of them (and only extract `word_string` from the disjuntcs). Now (after the said duplicate multi connector fix) `link_array` is always correct but `chosen_disjuncts` sometimes not.

**Question**:
I suspect that the chosen disjuncts connectors are not uses for anything just now. Is this correct?

**Possible solution**:
1. Supposing that the chosen_disjuncts connectors are unused, document this problem and ignore it for now.
2. Since I couldn't find a way (using the trailing connector encoding shortcut) that guarantees to preserve the original disjunct connectors, I can just recreate them from `link_array`.
3. Dig deeper in the code to find why the chosen_disjuncts connectors are correct >>99.9% of the times.
Maybe there is just easy way to fix that. This is the preferred solution, as then an `assert()` can be added to compare `link_array` to `chosen_disjuncts` in DEBUG mode and just have some way to validate the trailing connector encoding code.

**My proposal**:
Since for now I don't know how to implement (3), and supposing (1) is true, I would just document this problem.

BTW, while digging in `extract-links.c` I found some big inefficiencies that I also would like fix them in the next PR.

ampli:
> `!disjuncts` and also ` linkage_get_disjunct_str()` rebuild the disjuncts from `link_array` instead of just using chosen_disjuncts(done by `lg_compute_disjunct_strings()` in `lisjuncts.c`).

Edited my previous post to answer why it has been done... (I should have known...).

linas:
I don't recall what  `chosen_disjuncts` is used for.

Looking at  `linkage/score.c` it seems that its somehow used for computing the total score on a linkage. Maybe its possible to use the `link_array` instead ??

Looking at `parse/parse.c` its used only for debug prints.

Looking at `linkage/linkage.c` it seems to be in the code a lot, but "unused", except to get the cost (at the end of the file)

So -- if `link_array` is always valid, and it is filled-in and available by the time that we need `chosen_disjuncts` for something, then maybe we should eliminate `chosen_disjuncts` from the code.  We still need cost, and we still need to print, but if those are the only reasons `chosen_disjuncts` ...

ampli:
> The connectors in `chosen_disjuncts` are sometimes rather incorrect

I think I found the reason.
When the endpoints are suffix_id's, the same suffix_id may be from different disjuncts.
However, [issue_links()](https://github.com/opencog/link-grammar/blob/1950527d0705c1374c92391be34c5ae5ef0188c1/link-grammar/parse/extract-links.c#L749) issues the endpoint disjuncts too, while only the middle disjunct is consistent.

When I tried to use only the middle disjunct, I found that it is missing for word 0.
It seems this is because `record_choice()` in case lw==1 is sloppy:
```
					record_choice(dummy, NULL, NULL,
									  pset,  NULL, NULL,
									  NULL, NULL, NULL, &xt->set);
```
The old code didn't depend on that because `ld` got assigned to `chosen_disjuncts[0]`.
But now this `ld` may be incorrect (if my analysis is right).
I'm still struggling with fixing that.

ampli:
> Looking at `linkage/score.c` it seems that its somehow used for computing the total score on a linkage. Maybe its possible to use the `link_array` instead ??

Since `lg_corpus_disjunct_score()` uses `lg_compute_disjunct_strings()`, link_array is already used for that.

So it seems the only use of `chosen_disjuncts` for now is for `chosen_words`. I don't know another way to find the linkage words, so we may need to retain `chosen_disjuncts`.

I also didn't find any problem in `chosen_words` (on a detailed run of the batches with a very large number of linkages and with nulls allowed) so it seems the word_strings in `chosen_disjuncts` are always correct. But I don't have a "theoretical proof" for that yet.

ampli:
I said in a previous post above:
> > The connectors in `chosen_disjuncts` are sometimes rather incorrect
> 
> I think I found the reason.
> When the endpoints are suffix_id's, the same suffix_id may be from different disjuncts.
> However, [issue_links()](https://github.com/opencog/link-grammar/blob/1950527d0705c1374c92391be34c5ae5ef0188c1/link-grammar/parse/extract-links.c#L749) issues the endpoint disjuncts too, while only the middle disjunct is consistent.

This indeed was the reason of the incorrect disjuncts.
I don't know why the endpoint disjuncts were used at all.
They just overwrote the assignment of the middle disjunct.
As I already said, when I removed them I found that the disjunct for word 0 was missing.
It reappeared after I fixed the "sloppy" `record_choice()`.
I then had to disable issuing links when rw==sent_length (a dummy word) since this generates a bogus link (a link with only one side, no link to the dummy word) that causes a label-related failures.

ampli:
> I then had to disable issuing links when rw==sent_length (a dummy word) since this generates a bogus link (a link with only one side, no link to the dummy word) that causes a label-related failures.

I changed this check to disabling issuing links when rc==NULL, which is more general.
Maybe there is still some sloppiness in the said `record_choice()` arguments, but it turned out that this bogus links are not actually needed and can be ignored with no harm (comparing to the pre-suffix_id version).

Issue #869, opened 2018-12-23
Parsing stops when num_valid_linkages==0 and there was a parse overflow


ampli:
Normally, when no valid linkages are found, the parsing continues with an incremented null_count.
However, at the end of the null_count incrementing loop there is this check:
``` c
    418                 /* If we are here, then no valid linkages were found.
    419                  * If there was a parse overflow, give up now. */
    420                 if (PARSE_NUM_OVERFLOW < total) break;
```
This happens when all the parses have PP violations.
In such a case there are no results.

I think this is not useful, and propose to remove this check. Maybe a warning can be added that suggest to increase the linkage_limit, which often solves the problem - some long sentences have a very low density of valid pp linkages and even linkage_limit>100K is needed.

linas:
Yes, sure. Seems OK to me.

ampli:
It will be included in the next PR (to be sent ASAP).
But I couldn't add the said warning due to a lot of false positives - the `fixes` batch contains a lot of such examples (of all the linkages containing P.P. violations and increasing the length_limit doesn't help).

BTW:
Still  this problem (of very low density of linkages with no P.P. violations) seems to be typical for long sentences. And if a very big linkage_limit is used in a try to solve that, the "Sorted all linkages" consumes MUCH CPU (~1 minute for linkage_limit=1M).
Because the violations tend to repeat in linkages of the same long sentence, I thought to solve that by "caching" the involving links of the P.P. violations, so a fast comparison can be used to see if  a linkage contains a "known" violation (before making an actual postprocessing). For the SAT parser (in which postprocessing is the main bottleneck for long sentences) there is not even a need for such caching, as a constraint can be added to forbid the seen violation in further linkages.

linas:
I once added some checks to post-processing, and found that maybe 1/3 or 1/2 of the PP rules never, ever get triggered.  This is after parsing not only the usual corpus files, but also after parsing some other large files (some books, I think)   I was going to remove the unused rules, but then ... I didn't.   

Given that this affects performance, maybe its time to remove them. 

[...]

PR #871, opened 2018-12-29
Classic parse

ampli:
**Classic-parser**:
- On one-step-parse, share the count connector table of the null_count==0 step with the null_count>0 step.
  This is done by defining the suffix_id's incrementally in 2 steps.
- Don't stop on linkage overflow (see issue #869).
- Redesign the suffix_id assignment algo to be more efficient (~+2% on the en/ru basic batches, ~+4% on the fixes batch).
- Improve debug messages.

**link-parser main changes**:
- Allow one-step parsing tests using `-test=one-step-parse`. It can be later changed to be the default.
- Print parse time on verbosity > 1 (verbosity=2 is the designed way to time messages only.)
- Improve help messages.
- Increase MAX_INPUT to 2048

BTW, I have an additional PR to send (cleanup + typo fixes) for version 3.6.0.
I also have some more speed fixes but if desired I can keep them for version 3.6.1.

### DIFF
--- a/bindings/python-examples/sentence-check.py
+++ b/bindings/python-examples/sentence-check.py
@@ -78,6 +78,7 @@ except LG_Error:
 po = ParseOptions(verbosity=arg.verbosity)
 
 po.max_null_count = 999  # > allowed maximum number of words
+po.linkage_limit = 10000 # maximum number of linkages to generate
 po.max_parse_time = 10   # actual parse timeout may be about twice bigger
 po.spell_guess = True if DISPLAY_GUESSES else False
 po.display_morphology = arg.morphology
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -45,6 +45,7 @@
 #include "corpus/corpus.h"
 #include "memory-pool.h"
 #include "string-set.h"
+#include "string-id.h"
 
 typedef struct Cost_Model_s Cost_Model;
 struct Cost_Model_s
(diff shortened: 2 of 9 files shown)


linas:
So... is it time for version 5.6.0 yet?

ampli:
> So... is it time for version 5.6.0 yet?

1. I still have a cleanup PR (will send it shortly).
2. **Two tests fail**, apparently due to recent dict changes (one due to the changes today). I can send a PR that fixes them. But please check if the new linkages (shown when running the tests) are fine.

ampli:
There is one additional new problem, but it is not fatal.
Currently C++ linkage is always done, even if the SAT parser is not configured.
This is due to a limitation in autoconf.
The previous way of doing the C++ linkage was rather buggy so it cannot be used.
As far as I know there is no official fix for that, but it can be solved by a small fix which uses undocumented autoconf feature. But I guess this is not too important.

ampli:
> So... is it time for version 5.6.0 yet?

I have sent the last 2 PRs.
The rest will be for the 5.6.1.

linas:
> The rest will be for the 5.6.1.

What's the rest? I guess its not something you feel like doing in the next few weeks...

ampli:
> What's the rest? I guess its not something you feel like doing in the next few weeks...

I have many  WIP branches with useful performance enhancement and feature additions. Some of the changes I committed, but some not committed yet. Some need porting, as the code got changed meanwhile, and some need completion.
No fixes for known bugs there (I don't count the dict-order fix (issue #671), which somewhat reduces performance and I couldn't demonstrate this apparently potential bug using the current dicts).

The next things in the "rest" are mainly expression and memory allocation speedups. I hope this will alleviate the slowness encountered in the lang-learning project (more than 50% overhead of malloc/free and a large part of the rest is expression pruning. The actual parsing overhead there is surprisingly negligible). I have some fixes that start to address that.

PR #874, opened 2018-12-29
Fix broken tests

ampli:


linas:
I cherry-picked the second commit and pushed that; the first commit revealed a bug in the dict.

PR #875, opened 2018-12-30
 corpus-failures.batch: Remove duplicate sentences

ampli:
The file remains in DOS format and extra blank after many sentences, since I though it may be a good idea to keep one such batch file. But if it is better I can remove them.


linas:
Thanks! leaving it in DOS format is OK.

PR #876, opened 2018-12-31
extract-links.c: Cleanup for "Issue only the middle disjunct"

ampli:
Cleanup fix: Remove the arguments that became redundant in commit dce7cb723.


linas:
thanks!

PR #877, opened 2019-01-04
Fix disabling 4.0.*knowledge to be backward compatible

ampli:
Fix it in a way that will not generate errors by older libraries.
(Original PR: #836.)


linas:
This just missed version 5.6.0 which I just now pushed.

PR #879, opened 2019-01-06
read_dict.c line numbers + cleanup

ampli:
- Use 1-based line numbers in dict messages.
- Improve subdictionary open error message.
- Cleanup.


linas:
thanks

PR #882, opened 2019-01-15
Expression manipulation speedup

ampli:
Total speedup -  20x(best of 5) benchmark:
~16%    en/corpus-basic.batch
~21%    data/en/corpus-fixes.batch
~30%    data/ru/corpus-basic.batch
 ~1%    data/ru/corpus-fix-long.batch

The main speedup is due to using memory pools for `E_list`, `Exp` and `X_node` (implementing "not to free anything at all until sentence_delete()" that was mentioned in PR #880).
2 other changes (in `copy_Exp()` and `purge_Exp()` yield a small but noticeable speedup.
Very long sentences are almost not affected by these changes because the parsing time dominants for them.

---
Some future speedup work (there is more, some of it already previously mentioned):
- There is still some room to improve the performance by adding more memory pools.
- About 1/3 of the CPU for short sentences is the post processing, and most of its time is taken by `post_process_match()`. Most probably improving this function will yield a big speedup.
- For long sentences `form_match_list()` takes about half of the time. It can be improved too.
Second to it is do_count(). I need to port old connector-table cache improvements and see if they help now.
-  For `ru/corpus-basic.batch`, about 1/3 of the time (of the total - which includes including dict read) is taken by suffix_split(), which can be vastly improved.  Reading its dict can be sped up by memory pools and other fixes.

### DIFF
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -121,6 +121,9 @@ struct Sentence_s
 	String_set *   string_set;  /* Used for assorted strings */
 	Pool_desc * fm_Match_node;  /* Fast-matcher Match_node memory pool */
 	Pool_desc * Table_connector_pool; /* Count memoizing memory pool */
+	Pool_desc * E_list_pool;
+	Pool_desc * Exp_pool;
+	Pool_desc * X_node_pool;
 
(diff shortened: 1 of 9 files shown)


linas:
thanks!

linas:
Yeah, that's a pretty big speedup!

PR #883, opened 2019-01-23
string-id.c: Bug fix: Return the correct ID after the table grows

ampli:
This is a potentially severe bug, that may cause a bad or missing linkage.
The bug happens when the string-id table grows. The the affected trailing connector sequence usually gets ID=256 again (but may get another incorrect number).

This problem doesn't cause any missing or different linkage with the basic (all languages) and en//fixes batches, but during a WIP of memory sharing of common trailing sequences I encountered an unexpected problem that was caused by this bug.

**I propose to issue a new release soon**, maybe after a cleanup PR and a small simple speed improvement that I can cherry-pick from the said WIP.

### DIFF
--- a/link-grammar/string-id.c
+++ b/link-grammar/string-id.c
@@ -131,12 +131,14 @@ int string_id_add(const char *source_string, String_id *ss)
 	ss->table[p].id = ss->count;
 	ss->count++;
 
+	int keep_id = ss->table[p].id;
+
 	/* We just added it to the table.  If the table got too big,
 	 * we grow it.  Too big is defined as being more than 3/8 full.
 	 * There's a huge boost from keeping this sparse. */
 	if ((8 * ss->count) > (3 * ss->size)) grow_table(ss);
 
-	return ss->table[p].id;
+	return keep_id;
 }
 
 int string_id_lookup(const char *source_string, String_id *ss)


linas:
I can do a new release tommorrow. Let me know

ampli:
> and a small simple speed improvement

I still haven't prepared yet the PR for the said small speed improvement.

>> I can do a new release tommorrow. Let me know

Please do a new release.
The rest of the fixes I have will go to 5.6.2, so we will have time to test them.

---
BTW,
> ... WIP of memory sharing of common trailing sequences

It saves about 90% of the connector memory for long sentences, and initial tests show ~15% speedup for the `fixes` corpus.
While implementing it I encountered a (rare) problem which I don't understand of different nearest-word values for the same trailing connector sequences (of the same word), a thing that I didn't expect. Even though I found how to overcome that for the purpose of connector sharing, I will soon open an issue to discuss it.

ampli:
After more thinking on this bug, it seems to me a new release is not urgent after all.

The string-id initial table size is  9973. There is a need for 3/8 of it to be full in order to generate a bogus suffix_id. It means one per about 3740 different suffixes.

And a necessary condition is to have at least 2 such bad suffix_id on the same word, and in the same direction. But I have not seen so many suffix_id's on the same word even on long sentences.

Moreover, the bad suffix_id's need to be on pairs with non-zero count. And only about 1% of the pairs have such a count. And a large part of the pairs with non-zero count doesn't participate in the linkage (I don't have a number).

To sum up: It seems this bug has a low chance to be seen at all.

linas:
I got distracted and did not create a release. I can still do one, "just to be safe".  If you have more work planned, I can also wait a bit.

PR #885, opened 2019-01-26
Share connectors

ampli:
This modification shares connector memory, so for long sentences it takes only a small fraction of the original one. The idea is to speed up parsing by reducing CPU cache trashing.

What is shared is identical trailing connector sequences. The first connector in a sequence defines the sequence. It has to have the same `suffix_id`, and same `nearest_word` value. The rest of fields are supposed to be the same, but the `originating_gword` field is verified.

The same `suffix_id` requirement identify identical sequences due to the way it is created.
The same `nearest_word` requirement is supposed to guarantee that the rest of the corresponding connector pairs have identical `nearest_word` too. I say "supposed" because I couldn't think of a way it cannot be so, but as usual I don't have a proof. I added a verification in the code for that, and it has never been triggered.

The result is a speedup of about 15% for the "fix-long" corpus.
Only long sentence are significantly sped up, because for small sentences the connector memory space is much smaller than the CPU cache.

The sharing amount can be impressive. Here are the numbers for the `And yet [...]` sentence.
`echo 'And yet [...]' | link-parser -v=5 -ti=100 -debug=pack_sentence,prune.c`
Quoting from the output of the command above (the first set of lines is for null_count==0, the second one for null_count==1):
```
Total: 67830 disjuncts, 381889 (193458+/188431-) connectors
Trace: pack_sentence: Info: 361549 connectors shared

Total: 129603 disjuncts, 773943 (409562+/364381-) connectors
Trace: pack_sentence: Info: 743666 connectors shared
```
Note: My current implementation doesn't share connectors with different directions.
It also doesn't share non-trailing connectors.
The intention was to allow "patching" of connectors fields in do_count(), for tuning of distance fields on the fly. A test of this idea indicated a significant speedup. But doing that in the parsing stage has an overhead, so I try now to see if it can be done while pruning.

### DIFF
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -117,13 +117,17 @@ struct Sentence_s
 	const char *orig_sentence;  /* Copy of original sentence */
 	size_t length;              /* Number of words */
 	Word  *word;                /* Array of words after tokenization */
-	String_id *connector_suffix_id; /* Used for connector trailing sequence ID */
 	String_set *   string_set;  /* Used for assorted strings */
 	Pool_desc * fm_Match_node;  /* Fast-matcher Match_node memory pool */
 	Pool_desc * Table_connector_pool; /* Count memoizing memory pool */
 	Pool_desc * E_list_pool;
 	Pool_desc * Exp_pool;
 	Pool_desc * X_node_pool;
+
+	/* Trailing connector encoding stuff (suffix_id), used for speeding up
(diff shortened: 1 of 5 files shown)


linas:
thanks!

ampli:
> It also doesn't share non-trailing connectors.

An array of connectors (instead of linked lists) is needed for that.
I has a benefit of not needling the "next" pointer and maybe a more efficient connector scanning by increasing an index.

PR #886, opened 2019-01-26
Share connectors - last commit (cleanup)

ampli:
This is a commit I added and pushed after the initial PR push.
For some reason it got appended only to my private repository and not also into the LG repository (I have never encountered such a thing). Moreover, GitHub didn't offer me to create a PR as it always do after I push a branch with changes ( I'm doing it now manually).
(Maybe you were in the process of applying this PR when I pushed this commit?)

Anyway, here it is.


linas:
Thanks.

> and not also into the ...  GitHub didn't offer me to create a PR 

You forgot to `git push origin share-connectors`  or possibly pushed some other branch by instead.  I do this every now and then, and then spend 5 minutes wondering why my code disappeared :-)

PR #889, opened 2019-02-02
Changes for -verbosity

ampli:
While implementing some major changes I decided to do some minor ones.
Changes here:
- Add verbosity for (some) Info messages at level 3 (but maybe it should be 2, for not cluttering the output with timing). Move the previous one (file search messages) to 4. Others Info messages may be added later. 
- Add a message when all the linkages are with P.P. violations. This is tipical to a too low linkage limit.
- The most drastic change, I hope you will like it (else just remove it...):
Don't notify "Freeing dictionary" unless the verbosity is at least 3. Maybe the only usefulness of it is if your dict gets out of scope unintentionally in Python.

### DIFF
--- a/README.md
+++ b/README.md
@@ -557,7 +557,7 @@ directory in that order, directly or under a directory names `data`:
 If link-parser cannot find the desired dictionary, use verbosity
 level 3 to debug the problem; for example:
 ```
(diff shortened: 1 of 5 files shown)


linas:
Thanks!

PR #890, opened 2019-02-04
tests.py: Fix HEnglishLinkageTestCase

ampli:
The sentence:
`Scientists sometimes may repeat experiments or use groups.`
now gives:
`Found 424 linkages (256 had no P.P. violations)`
instead of previously:
`Found 88 linkages (88 had no P.P. violations)`

As a result of the increased number of linkages and the P.P. violations, the lowest score linkage is not found now among 300 linkages, **and the test fails**.

Fix it by a more pessimistic value.


linas:
Thanks. That's an .. unexpected change.... I'm surprised that happened. :-(

PR #891, opened 2019-02-04
Cleanup ++

ampli:
Before sending a rewrite of `power_prune()` (with a drastic efficiency improving, but still same pruning) which is in the final stages, I am sending some commits that I separated from it.
(Another preparation PR for stuff needed for the new `power_prune()` will hopefully follow in the next hours).

In this PR:
- Removing of the (unused) ALT-consistency code.
- Type changes.
- Printout functions that I used to copy/modify etc. to WIP branches. But it will be less tedious for development if they are always included.
- A very small efficiency improve in `*_connector_list_update()`.
-  Prepare `put_into_power_table()` for a future change.
- Fix a comment rot.
- Initialize `suffix_id` (for convenience, e.g. preventing garbage on debug printouts of connectors).

### DIFF
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -16,7 +16,7 @@
 #include "api-structures.h"
 #include "corpus/corpus.h"
 #include "dict-common/dict-utils.h"
-#include "disjunct-utils.h"  // for free_disjuncts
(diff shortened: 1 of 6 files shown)


linas:
thanks!

linas:
that is, looks good!

PR #892, opened 2019-02-05
count.c: Fix a bug with initial words that have no left disjuncts

ampli:
Problem reported by Scott Guthery at
https://groups.google.com/d/msg/link-grammar/ynDNXmih5xk/4rm6sFRpHQAJ
See the commit message for a how to test this bug.

I tested this solution, but as always - I cannot prove its correctness.

There seem to be no sentences in the basic/fixes `en` corpus that has been affected by this problem.
However, 3 sentences in the `lt` corpus have been affected  (null `LEFT-WALL`):
Old:
`LEFT-WALL [ji] [eina] [į] [mokyklą] [.]`
With this PR:
```
                 +---LI--+
                 |  +-DG-+
                 |  |    |
LEFT-WALL [ji] eina į mokyklą [.]
```
Old:
`LEFT-WALL [Brolis] [eina] [į] [mokyklą] [.]`
With this PR:
```
                     +---LI--+
                     |  +-DG-+
                     |  |    |
LEFT-WALL [Brolis] eina į mokyklą [.]
```
Old:
`LEFT-WALL [medinis] [irklas] [buvo] [baltas] [.]
With this PR:
```
             +-------DVvv----->+
             |       +LLPIRMIas+
             |       |         |
LEFT-WALL medinis irkl.=    =as.pvv [buvo] [baltas] [.]
```

The `tr` corpus is also badly affected by this problem. Now I understand why its dict author used there `island-ok=1`

I also checked at the CMU site.
The problems seem not to exist in LG version 2.1, since it has this comment:
`    /* the following allows the possibility of a null link used on the leftmost word */`
However, LG version 4.1 contains this exact problem.

### DIFF
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -375,7 +375,8 @@ static Count_bin do_count(
 	{
 		int nopt_words = num_optional_words(ctxt, lw, rw);
 
-		if ((null_count == 0) || (!ctxt->islands_ok && (lw != -1)) )
+		if ((null_count == 0) ||
+		    (!ctxt->islands_ok && (lw != -1) && (ctxt->sent->word[lw].d != NULL)))
 		{
 			/* The null_count of skipping n words is just n.
 			 * In case the unparsable range contains optional words, we
@@ -409,6 +410,7 @@ static Count_bin do_count(
 		for (int opt = 0; opt <= !!ctxt->sent->word[w].optional; opt++)
 		{
 			null_count += opt;
+
 			for (Disjunct *d = ctxt->sent->word[w].d; d != NULL; d = d->next)
 			{
 				if (d->left == NULL)
@@ -417,6 +419,7 @@ static Count_bin do_count(
 						do_count(ctxt, w, rw, d->right, NULL, null_count-1));
 				}
 			}
+
 			hist_accumv(&t->count, 0.0,
 				do_count(ctxt, w, rw, NULL, NULL, null_count-1));
 		}
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -418,7 +418,8 @@ Parse_set * mk_parse_set(fast_matcher_t *mchxt,
 		Parse_set* dummy;
 		Disjunct* dis;
 
-		if (!pex->islands_ok && (lw != -1)) return &xt->set;
+		if (!pex->islands_ok && (lw != -1) && (pex->words[lw].d != NULL))
+			return &xt->set;
 		if (null_count == 0) return &xt->set;
 
 		RECOUNT({xt->set.recount = 0;})


linas:
thanks!

linas:
>  cannot prove its correctness.

:-)  I looked but did not have any particular insight. Scott also noticed that LG is kind of schizo w.r.t islands. I pointed out that "null words are just like islands of size one", but much of the code doesn't actually treat it that way.

It also raises interesting issues with cost: is it better to have two null words, or one island? Stuff like that ...

ampli:
> Scott also noticed that LG is kind of schizo w.r.t islands. I pointed out that "null words are just like islands of size one", but much of the code doesn't actually treat it that way.

It is actually **very consistent** with null words and islands, and handles them **exactly** the same!
The null_count is is  (num_islands-1). Null words are counted as islands. I think that the only "exception" is when all the words are null, in that case the null_count is the number of words. See the code of `sane_linkage_morphism()` and `num_islands()`.

All of this exact handling is a must in order to reject "insane morphism" and handle "optional words"
(because an optional word appears as a null word in a linkage but of course it should not be counted as such).

> It also raises interesting issues with cost: is it better to have two null words, or one island? Stuff like that ...

This is exactly the same linkage, **represented** in two different ways (I guess that by one island you actually mean a sentence with 2 groups of connected words, that I call "two islands which represent (num_islands-1) null words".

ampli:
As this PR may be considered a major bug fix, maybe it should be included in the ChangeLog.
But I am not sure what exactly to write. The meaning should be something like:
`Fix the ability to make a linkage when the first sentence words have no left disjuncts.`

linas:
> It is actually very consistent 

Oh, well, that's good, then. Scott's example was `linkage_unused_word_cost` which differs from `sentence_null_count` when islands are involved. What we really need are functions to return "number of islands", and "number of unlinked words" ... implementing these might be as easy as renaming some of the existing functions, but its not entirely clear that they are fully self-consistent, and the documentation for them is silent on the matter.  We might even want to have "get number of islands of size n" in the API.(?)

> This is exactly the same linkage, represented 

I mean "blah blah [blah] blah blah [blah] blah" which has two null words; there might be a different linkage with zero null words, but one island. Which is better?

> 2 groups of connected words

If it's connected to the left-wall, its not an island ....

linas:
> ChangeLog

Done. But this bug only shows up if there is no LEFT-WALL, right? It doesn't affect English or Russian, right?  I vaguely recall seeing crazy stuff that did not make sense, but I don't know if it was this bug, or simply user-error.

ampli:
> Scott's example was `linkage_unused_word_cost` which differs from `sentence_null_count` when islands are involved.

Indeed, because `linkage_unused_word_cost` just counts null words,  they may be validly different for some/all of the linkages (this is not related to the optional word problem described below). Because of this link-parser reports `sentence_null_count` for the whole sentence and `linkage_unused_word_cost` for each linkage.
So this documentation is correct only for `island_ok=0`:
"The linkage_unused_word_cost() routine should return the same value as sentence_null_count()."
Islands are a later addition so maybe this is a leftover from then.

It is indeed easy to provide the functions you state, even though I'm not sure how much ""get number of islands of size n" is useful.

In any case, I now see that `linkage_unused_word_cost` is not correct after I added optional words, because null optional words shouldn't be counted unless another optional word in their alternative is not null (this is the mechanism that replaces empty-word). This is something that `sane_linkage_morphism` knows about.
This is not related to Scott's example.
**I will add this to my list of needed fixes.

> I mean "blah blah [blah] blah blah [blah] blah" which has two null words; there might be a different linkage with zero null words, but one island. Which is better?

I don't know,  but I guess one or the other may be better for different purposes.
Maybe if `null_count !=0`, there is a need to produce linkages with and without islands (in the application, not the library of course), as the separated connected groups may be just interpreted as separate sentences.

> If it's connected to the left-wall, its not an island ....

For the library the left-wall is not special, so what you say is that the leftmost connected group is not to be counted as an island. I think this is just a matter of definition. For the purpose of the code there is a need to find out these groups, including the first one. Your definition maybe makes more sense because using it, the number of islands is equal to the null_count (no need to reduce one as in my definition).

linas:
> I think this is just a matter of definition. 

Yes, it is; but I think its a convenient one, for explaining things to others.  For the same reason, adding calls for "number of islands", and "number of unlinked words" while also deprecating "null count" is a good idea, not because the algo changes, but because it is easier to document, easier to explain, easier to understand.

>  `lt` corpus file.

The first half parses perfectly, the second half, not at all. The `lt` 4.0.dict explains about 1/20th of the grammar of Lithuanian. I coded up the first 1/3rd of chapter one of a book on grammar; there's 8 more chapters to go :-)

linas:
> a need to produce linkages with and without islands 

which is why, once upon a time, I had enabled `islands_ok`.  I think it would probably be a good idea to change the command line parser to explore both with and without islands-ok turned on, and report both  sets of results. When I say "would probably be a good idea", it means "I'm not really sure".  I would have to go through `4.0.corpus-failures` with and without islands, and see if islands make things better or make things worse.

ampli:
> both with and without islands-ok turned on

But how this can be done in practice,  if a sentence have more than a few linkages?
It doesn't seem so useful to show many linkages (tens or even hundreds) with `islands_ok=0` and then show the ones with `islands_ok=1`.

Maybe show them interleaved?

linas:
> But how this can be done in practice,

Don't know :-) If they were all given a cost, and dumped into one big pile, then they could be sorted according to cost. Easy to say, but maybe not practical to actually do it this way.

ampli:
Maybe one of the metrics can be "the length of the longest sequential connected words".

linas:
Yes.

PR #896, opened 2019-02-08
 Implement memory pools for disjuncts and connectors

ampli:
A significant speedup results.
The `en` basic/fixes corpus benchmarks show nearly 15% speedup and the `ru` corpus shows about 10% speedup.  For the fix-long corpus there is less speedup (~4%).

This implementation is a fast change for enabling me to finish the fast `power_prune()` implementation
(that also gives a significant speedup). While making last code reviews I decided that its memory management is over-complex because the new code uses memory pools for disjuncts/connectors, but the old code doesn't do that so it needs to consider that and the code has become complex and unmanageable (even though it works with no leaks). After this PR I will be able to simplify things considerably.

### DIFF
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -123,6 +123,8 @@ struct Sentence_s
 	Pool_desc * E_list_pool;
 	Pool_desc * Exp_pool;
 	Pool_desc * X_node_pool;
+	Pool_desc * Disjunct_pool;
+	Pool_desc * Connector_pool;
 
 	/* Trailing connector encoding stuff (suffix_id), used for speeding up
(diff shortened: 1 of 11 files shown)


ampli:
Please don't apply it yet - I was able to cause it to leak memory....
I will add a commit to fix it.

ampli:
Unfortunately, I was also able to cause heap-use-after-free...
I will fix it and make many more checks.

ampli:
Adding a commit may upset bisect, so I just forced push.
The problem was in the code that supports "one step parse".

Now it is ready for merging.

linas:
Wow. This looks real clean. Thanks!

PR #900, opened 2019-02-10
One-step-parse connector-pair table sharing bug bypass

ampli:
I hope I will be able to find a fix, but the speedup of this idea was not drastic.

### DIFF
--- a/link-grammar/parse/parse.c
+++ b/link-grammar/parse/parse.c
@@ -306,7 +306,15 @@ void classic_parse(Sentence sent, Parse_Options opts)
 			 * nulls only. So in case of one-step parse (min_null_count==0 &&
 			 * max_null_count>0) the table is shared and the null_count==0
 			 * parsing will be inferred from the table. */
+			// There are a few cases in the "fixes" corpus in which the number
+			// of parses is different. The missing linkages are all with P.P.
+			// violation, but nevertheless this is not expected and thus this
+			// idea is disabled until a fix is found. See "#if 0" below.
+			// Test sentence: He was too far gone to come back from such a loss
+			// (117 vs 111 linkages.)
+#if 0
 			if (!real_suffix_ids || (nl == 0) || !is_null_count_0)
+#endif
 			{
 				free_count_context(ctxt, sent);
 				ctxt = alloc_count_context(sent);


linas:
:-(

ampli:
I found a theoretical problem with this idea.

The idea was to make parsing with null_count>0 faster by reusing the count table of the null_count==0 step, thus avoiding a full recalculation of this step. This reusing is made possible if the "tracon IDs" that are used as keys of the count table are kept the same after the pruning step for null_count>0.

However, there is a very subtle problem due to the pp_pruning step (so this problem may currently happen only for English). Let's consider what happens without implementing this idea: In pruning for null_count>0 no word-distance optimization is done, and thus more disjuncts are retained. These disjuncts may contain criterion connectors, and thus pp_prune also may retain more disjuncts. These disjuncts may create more linkages. With the current  basic/fixes corpus sentences, all these linkages happened to be postprocess-discarded because the said more linkages somehow don't happen to contain the needed criterion links. But it seems that in theory they could, so some of the extra linkages may turn out to be valid. When implementing the idea, due to the reused null_count==0 table these potential good linkages will be missing.


Meanwhile I introduced a major improvement in `power_prune() `and `pp_prune()` that makes them yet several times faster (than the already recently-introduced fast implementation...). I also totally reimplemented the `tracon_id` (formerly `suffix_id`) generation by directly hashing the connectors instead of their string representation (which cause a major overhead), which is much more efficient. The result is a significant speedup in the basic/fixes batch benchmark. With the new extra-fast-pruning, another idea (first time mentioned here)  can be implemented efficiently: Extending the null_count==0 optimization to null_count>0 (so the pruning step is done again when the null_count is increased). A preliminary test for long sentences showed a several times speedup for sentences with null_count==1 (I haven't benchmarked it yet for null_count>1).

The said new implementation doesn't contain the ability to generate the same tracon IDs for the same tracons, so I will remove the implementation of this table sharing idea from the code (instead of just commenting out its final stage as done in this PR). If needed it can be re-implemented.

linas:
>  very subtle problem 

If you wish to test, you can also try `corpus-failures.batch` of which 1555 out of 1801 require null_count>0 I can also send you my version of pride-n-prejudice, which has  2028 out of 5839 null>0

> very subtle problem ... a major improvement

?? So you have a major improvement, but it has this subtle problem?  Or is it a different improvement, that does not have this problem?

I am somewhat flexible about "subtle problems". If a change enables significant speedups that don't affect corpus-basic and corpus-fixes, this seems fairly acceptable to me.  When link-grammar fails to parse a sentence in corpus-failures, its because the English dictionary is imperfect. Altering the null_count>0 behavior is not likely to lower parse quality .. when null_count>0 the parses are probably already bad, and I can't imagine that your changes make it worse.

ampli:
> I can also send you my version of pride-n-prejudice

Yes, please send me.
>> very subtle problem ... a major improvement

> ?? So you have a major improvement, but it has this subtle problem? Or is it a different improvement, that does not have this problem?

These are different things.

The subtle problem is for reusing the count table of `null_count==0` when parsing with `null_count>0`.
According to the analysis above, good parses may be missing. But this has not been observed in actual tests, so my guess is that it should be very rare.

The major improvement I referred too is making the pruning yet much faster (this is the git branch I'm now trying to convert to clean PRs). This may allow to make optimized pruning, with a very low overhead, before parsing with the next `null_count`.

Currently the pruning for `null_count==0` is specially optimized and if parsing with `null_count>0` is needed, the pruning is then done again without this optimization. However, this optimization can be adapted for any given null count. For example,an optimized `null_count==1` pruning speeds up the parsing of the `And yet [...` sentence (which currently has one null word `quite`), more than twice faster.
So the idea is to prune with the needed optimized `null_count` before each parse with an increased `null_count`.

> when null_count>0 the parses are probably already bad, and I can't imagine that your changes make it worse.

So I will try to reimplement the count table sharing. This will take some time because the new way of generating tracon IDs (to be submitted soon) is totally different than the current way, so generating them incrementally needs a radically different code. Implementing such an incremental tracon generation will also improve the efficiency for parsing with `null_count>0` (whether count table sharing will be used or not).

Issue #907, opened 2019-02-13
version after pull req #906 segfaults


linas:
I just observed a segfault when running the current LG version on my squonky copy of pride-n-prejudice.  I'm trying to figure out how to attach this file now.

linas:
file:///home/linas/pandp-union.txt.bz2

linas:
file:///home/linas/pandp-union.txt.gz

linas:
[pandp-union.txt.gz](https://github.com/opencog/link-grammar/files/2858923/pandp-union.txt.gz)

linas:
The crash was just under 1/2-way through with that file, for you it should be about 5-6 minutes in.

ampli:
It's something concerns panic mode.
I'm investigating it now.

linas:
I haven't tested #908 yet. Here's the stack: 
```
His behaviour to herself could now have had no tolerable motive; he had either been deceived with regard to her fortune, or had been gratifying his vanity by encouraging the preference which she believed she had most incautiously shown.
Every lingering struggle in his favour grew fainter and fainter; and in farther justification of Mr. Darcy, she could not but allow Mr. Bingley, when questioned by Jane, had long ago asserted his blamelessness in the affair; that proud and repulsive as were his manners, she had never, in the whole course of their acquaintance -- an acquaintance which had latterly brought them much together, and given her a sort of intimacy with his ways -- seen anything that betrayed him to be unprincipled or unjust -- anything that spoke him of irreligious or immoral habits; that among his own connections he was esteemed and valued -- that even Wickham had allowed him merit as a brother, and that she had often heard him speak so affectionately of his sister as to prove him capable of some amiable feeling; that had his actions been what Mr. Wickham represented them, so gross a violation of everything right could hardly have been concealed from the world; and that friendship between a person capable of it, and such an amiable man as Mr. Bingley, was incomprehensible.
Timer is expired!
Entering "panic" mode...

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7b6d476 in share_disjunct_jets (sent=0x5555564b9d50, 
    rebuild=<optimized out>) at ../../link-grammar/disjunct-utils.c:1150
1150					jet_table[dir][id]->suffix_id++;
(gdb)  bt
#0  0x00007ffff7b6d476 in share_disjunct_jets (sent=0x5555564b9d50, 
    rebuild=<optimized out>) at ../../link-grammar/disjunct-utils.c:1150
[...]

PR #908, opened 2019-02-13
 Fix crash in classic_parse(): Fix condition for freeing the jet table

ampli:
Most probably this is the problem of issue #907.
PR #906 has another bug, but I couldn't cause a crash even with UBSAN/ASAN enabled (I will send a fix too).


ampli:
> PR #906 has another bug, but I couldn't cause a crash even with UBSAN/ASAN enabled (I will send a fix too).

I thought there is a bug but I couldn't find one...
The algo is just somewhat tricky, for efficiency.
So I will add a comment on it.

PR #910, opened 2019-02-14
Russian speedup (35%) and more

ampli:
- Re-enable the code that prevents "recursive" morpheme splits.
I disabled it once when I realized that it is not compatible for the then new `tr` dict.
It didn't so badly affected Russian then because the parsing time dominated then.
Now the parsing-related time when running the basic corpus batch is unbelievable small (in the order of 10%) and most of the CPU is taken by the dict read, tokenizing and malloc().

- dictionary_six_str(): Replace the strange check for the affix file name.
I named my git branch directory `affix_suf` and wondered why I get horrible errors when running the tests... I then recalled the "temporary" hack I once did. I have never got to rewrite the dictionary file handling, but for some time it is higher in my list due to the need to read dictionaries from strings.

### DIFF
--- a/link-grammar/dict-common/dict-common.h
+++ b/link-grammar/dict-common/dict-common.h
@@ -93,6 +93,7 @@ struct Dictionary_s
 	/* Affixes are used during the tokenization stage. */
 	Dictionary      affix_table;
 	Afdict_class *  afdict_class;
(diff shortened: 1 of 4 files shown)


linas:
thanks. Sorry for delay.

PR #911, opened 2019-02-19
do_count(): Avoid linkages loss due to overflow

ampli:
While I was rewriting pp_prune() (with astonishing observations and results, BTW), I noted that I don't get any good linkage for this sentence from the fixes-long batch:
```
In 1608 he wrote a treatise of the Errors, False Gods, and Other Superstitions of the Indians of the Provinces of Huarochiri, Mama, and Chaclla, of which unfortunately only the first six chapters are known to exist and have been translated into English.
```
The number of linkages was the same - as -v=5 shows (the link-parser's result is the 24-bit truncated one).
However, all were with P.P. violations.

I first thought this may be a problem with the random sampling - if it is  not so random it may miss part of the solution space. However, inspecting the `list_random_links()` code convinced me it is indeed random (and I also found how to improve this code speed...). Printing the `Parse_set` structs showed that the expected linkages are just not there.

Neutralizing  the new `pp_prune()` code didn't help. But during the tests I found that the linkages (and especially no linkages) strongly depend on the order of the disjuncts on each word. For example, in the sentence quoted above when there are no good linkages, if one reverses the disjunct order on word 10 (a comma) all the missing linkages magically reappear! But if even one totally unrelated idiom is added to the dict (it adds a connector type that changes the order of pruning due to hashing effects - and hence the order of disjuncts ), they may be missed again... It seemed horribly and hopelessly strange.

But then I added a printout of the count due to processing of each disjunct on word 0.
It turns out the problem happens when the count overflows.
The following, added at 4.6.3, causes it:
``` c
    516                                         if (INT_MAX < total)
    517                                         {
    518                                                 total = INT_MAX;
    519                                                 t->count = total;
    520                                                 put_match_list(sent, m1);
    521                                                 return total;
```
When an overflow happens, it just aborts the disjunct scanning loop between the currently handled words, and hence the parse result of sentences with overflow strongly depend on the order of disjuncts . As a result, many/all of good linkages may miss from the solution space, especially if the overflow results in the start of the sentence processing (like what happens with this example sentence,
for which only few disjuncts on word 0 don't have a parsing path that leads to overflow).
Printout example from this sentence - **all** the good solutions are missing after the overflow:
```
w 0 0 count=0 hWc+ hWV+ RW+ 
w 0 1 count=0 hWd+ hWV+ RW+ 
w 0 2 count=0 hCPa+ 
w 0 3 count=0 hWo+ hWV+ RW+ 
w 0 4 count=0 hWp+ hWV+ RW+ 
w 0 5 count=0 hWc+ hWV+ Xx+ hWV+ 
w 0 6 count=0 hWd+ hWV+ Xx+ hWV+ 
w 0 7 count=0 hWp+ hWV+ Xx+ hWV+ 
w 0 8 count=0 hWa+ hCPi+ Xp+ 
w 0 9 count=0 hWd+ hCPi+ Xp+ 
w 0 10 count=0 hWp+ hCPCi+ Xp+ 
w 0 11 count=0 hWa+ hCPu+ Xp+ 
w 0 12 count=0 hWc+ hCPu+ Xp+ 
w 0 13 count=2147483647 hWd+ hCPu+ Xp+ 
w 0 14 count=2147483647 hWi+ hCPu+ Xp+ 
w 0 15 count=2147483647 hWp+ hCPu+ Xp+ 
```

Removing the loop termination code (lines 520-521 above) fixes everything.
Sentence batch "benchmarks":
1. No observed slowness on the `en` "basic"/"fixes" and `ru` "basic" batches.
2. About 3% slowness on the "fix-long" batch.
3. About 5% slowness on the "failures" batch (with **cost_max=2.7** instead of 2), but with slightly less errors (1567 instead of 1571). (The `cost_max=2` is maybe a remain from old times that needs an update.)

However, WIP speedups of several tens of percents may offset (2) and (3). Also, I think it is a **must** due to correctness, which is usually preferred over speed. Maybe part of the overflows can be mitigated by a more aggressive pp-pruning, because absolutely most of the linkages for long sentences contain P.P. violations. Also, may be an overflow optimization would help. (E.g. if the LHS overflows, maybe we only need to know if the RHS has any count without fully calculating it. But this of course will add overhead for sentences w/o overflow.)

BTW, even though this fix may prevent hiding the lower-metric solutions (or the only non-P.P. violation ones) it doesn't automatically give a better metric on the first result, because also more solutions with P.P. violations are produced, and the sampling is from the linkage_limit number of actually produced linkages. The new `pp_prune()` prunes more but I haven't checked it yet in this regard.

### DIFF
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -641,8 +641,6 @@ static Count_bin do_count(
 					total = INT_MAX;
 #endif /* PERFORM_COUNT_HISTOGRAMMING */
 					t->count = total;
-					pop_match_list(mchxt, mlb);
-					return total;
 				}
 			}
 		}


linas:
> the linkages (and especially no linkages) strongly depend on the order of the disjuncts on each word.

Well, that was alarming to read. I was relieved that it was only during overflow.

> The cost_max=2 is maybe a remain from old times that needs an update.

Yes.

> if (INT_MAX < total)

A long time ago, the count was stored in a 32-bit int, and overflows were very common. The reason for this was that `total_count=left_count * right_count` and so assuming the worst case, left and right had to be 16-bits or less. Yuck. Increasing the count to 64-bit helped. The worst-case assumption of left, right = sqrt(total) was not changed (viz sqrt(64-bits) = 32-bits). The 24-bit limit seemed very reasonable at the time, since 24=16 million which seemed like a crazy-large number. If I recall correctly, making this number larger did not improve any results, but did make things slower.

You are welcome to alter the counting-overflow logic.It's kind-of unfortunate that these overflows occur... but I don't see any easy short term fix, nor even a clear long-term fix.

ampli:
For complex/long sentences most of the linkages contain P.P. violations.
So while investigating `extract_links()`, and because I already rewrote `pp_prune()` and much improved the overhead of postprocessing (another WIP), I thought about making postprocessing filtering on the the `Parse_set` elements directly, remove the offending ones (the link names are already apparent there). Since the counts are then recalculated, maybe a significant number of overflows may be eliminated.

ampli:
In a try to get more accurate parses also in random sampling, in 5.6.0 I implemented the disjunct sorting idea that you once started to implement.
You can try it using:
`link-parser -test=sort-match-list`

For some reason it doesn't give better results in the first linkage.
For example, The `In 1608 [....]` sentence gives `DIS=12.81 LEN=171` without sorting and `DIS=13.90 LEN=277` with sorting. If the idea still seems sound, there is a need to check my implementation...

ampli:
Another possible approach is to use a graph minimal-distance approximation algorithm to traverse the `Parse_set` structures.

linas:
>  If the idea still seems sound, 

At this point, I don't know and am not sure I care that much. The scientific approach would be to take 500 or 1000 random parses, bin-count them (histogram them, say, with bind widths of 0.5 or 0.3) This will illustrate what the "typical" cost is, where most of them are, and based on the slope/falloff, maybe hint at what the best-possible might be.  

But that takes work. so maybe we can assume that DIS=14 is typical, and DIS=13 is best-case, which suggests one is unlikely to do better. Do you feel like being scientific?

I think it would be interesting to characterize the parser generically, in this way. Given some corpus ... how many parses? what costs, what distributions? Is distribution zipfian? something else? is the cost difference between the 1st, 2nd, 3rd, etc. N'th best parse small, or large? There's some kind of head/tail, what is it shaped like?  I can't say what one could do with these answers if I had them, but they could be .. interesting to ponder, for theoretical reasons.  ... these questions are posed for the hand-written English dict; what might they be like for the auto-gen'ed dictionaries (which are still low-quality; I don't have time to work on them any more, and the team that took over is caught in a sand-trap.)

PR #915, opened 2019-02-21
Fast pp_prune()

ampli:
Main speedup methods:
- Use the `cms` table for checking the the trigger connectors. 
Relatively few connectors are needed to be checked v.s. all the connectors in the current implementation.
- Less match tries for criterion links - `#` do not generate match tries.
- `rule_satisfiable()` is called on each rule at most once per sentence.
- `cms` table insertion is done using the jet table.
- Bad connector marking uses the jet table and the jet sharing.
- The disjunct deletion is done by `power_prune()`, along with additional disjuncts that it then able to delete.
- MRU order is kept on table insertion, to speed up checking for uniqueness.

Minor speedup methods:
- No use of reference counts (they are useless at the current implementation). This also results in `cms` table element size of power of 2.
- The `cms` table lookup is done using connector descriptor comparisons (saving indirections).
- `cms` table init using `memset()`.

Speedup methods that are not implemented yet:
- Use memory pools for the `cms` table.
- Use descriptors also for PP links. This will allow fast hashing and fast pp links matching.
- Find a way to avoid reference counts in the power_table (it seems possible). This will make the `cms` table unnecessary and save its creation.

It may be a good idea maybe that I also add copyright on this file for these and the `power_prune()` changes. However, there is a planned cleanup PR and for this file, some more speedup ideas to try, and also I have several different changes for more aggressive pruning (WIP). I also still think how to unify the jet and power tables.

### DIFF
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -951,13 +951,15 @@ static bool match_in_cms_table(multiset_table *cmt, const char *pp_link,
 	return false;
 }
 
-static Cms *lookup_in_cms_table(multiset_table *cmt, const char *pp_link)
+/* FIXME? There is some code duplication here and in insert_in_cms_table()
+ * but it seems cumbersome to fix it. */
+static Cms *lookup_in_cms_table(multiset_table *cmt, Connector *c)
 {
-	unsigned int h = cms_hash(pp_link);
+	unsigned int h = cms_hash(connector_string(c));
 
 	for (Cms *cms = cmt->cms_table[h]; cms != NULL; cms = cms->next)
 	{
-		if (string_set_cmp(pp_link, connector_string(cms->c))) return cms;
+		if (c->desc == cms->c->desc) return cms;
 	}
 
 	return NULL;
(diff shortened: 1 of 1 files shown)


ampli:
The PP criterion check is still not the strictest possible (but it is better than the original one).
For example, if a criterion PP link is `Xabc`, and there are 3 connectors `Xa`, `X*b` and `X**c`, then the current check will pass even though there is no way to construct the criterion link. The ability to construct the criterion link also depend on the direction of the connectors.

There is, of course, a way to write an exact test (with much more overhead due to more checks and the need to keep the connector direction in the `cms` table or use different tables for each direction) but I wonder if it would yield more pruning - at least by a quick glance I couldn't find such examples. Of course it is possible to actually test it (after implementing the strictest possible test). Maybe I will do that.

One algo for that can be:
1. Find all the connectors that may participate in the construction of the criterion link.
2. Separate the list to left and right connectors.
3. "easy_match" all the possible pairs (there should not be too many) until there is a match or the list exhausted. The criterion passes iff there is a match.

PR #919, opened 2019-02-24
Vector implementation for (Gword *) list

ampli:
The vector implementation is from https://github.com/augfab/vc_vector.
The original is https://github.com/skogorev/vc_vector but it is not maintained.
I had to fix many compilation warnings, and bad bugs in one function. I also improved performance of appending elements by converting a division to multiplication in the memory size computation.

There is still a small room for performance improvement of `vc_vector` by holding the current memory size in the vector header. Currently, this size is computed by multiplication on each append and vector-end operations. In addition, the implementation is small enough to maybe put in a header file for inlining.

It is also possible to improve the relevant LG code to make much less use of vector copy
(by moving the vector to the next word in case no alternatives are involved). Much of the Gword vector use can also be eliminated for sentences that don't have optional words. But for now I will put the efforts in the other WIP changes.

Benchmarks of `en` >"basic"/"fixes" doesn't indicate a performance change.
For Russian I saw about  ~1% speedup.
For long sentences there is about ~1% slowness, but didn't do too many tests ( 3x3 test as opposed to a 15x5 test for "basic"/"fixes").

### DIFF
--- a/link-grammar/linkage/sane.c
+++ b/link-grammar/linkage/sane.c
@@ -192,12 +192,12 @@ static size_t num_islands(const Linkage lkg, const Gwordlist *wg_path)
 	int inum = -1;
 	Disjunct **cdj = lkg->chosen_disjuncts;
 	Gword **wg_path_p = gwordlist_begin(wg_path);
+	Gword **wg_path_p_end = gwordlist_end(wg_path);
 
 	for (WordIdx w = 0; w < lkg->sent->length; w++)
 	{
 		/* Skip null words which are optional words. */
-		if ((gwordlist_end(wg_path) == wg_path_p) ||
-		    ((*wg_path_p)->sent_wordidx != w))
+		if ((wg_path_p_end == wg_path_p) || ((*wg_path_p)->sent_wordidx != w))
 		{
 			assert(word[w].prev == word[w].next);
 			assert((NULL == cdj[w]) && lkg->sent->word[w].optional);
@@ -207,7 +207,7 @@ static size_t num_islands(const Linkage lkg, const Gwordlist *wg_path)
 			continue;
 		}
 
-		if (gwordlist_end(wg_path) != wg_path_p)
+		if (wg_path_p_end != wg_path_p)
 			wg_path_p = gwordlist_next(wg_path, wg_path_p);
 		if (NO_WORD == word[w].prev) continue;
 
@@ -291,8 +291,9 @@ bool sane_linkage_morphism(Sentence sent, Linkage lkg, Parse_Options opts)
 
 	/* Populate the path word queue, initializing the path to NULL. */
 	Gwordlist *glsent = sent->wordgraph->next;
+	Gword **sent_end = gwordlist_end(glsent);
 	for (Gword **gw = gwordlist_begin(glsent);
-	             gw != gwordlist_end(glsent);
+	             gw != sent_end;
 	             gw = gwordlist_next(glsent, gw))
 	{
 		wordgraph_path_append(&wp_new, /*path*/NULL, /*add_word*/NULL, *gw);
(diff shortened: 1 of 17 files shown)


link-grammar/linkage/sane.c
```diff
@@ -400,9 +399,13 @@ bool sane_linkage_morphism(Sentence sent, Linkage lkg, Parse_Options opts)
 				if (gl->o_gword == wpp->word)
 				{
 					match_found = true;
-					for (next = wpp->word->next; NULL != *next; next++)
+					Gwordlist *wn = wpp->word->next;
+					Gword **gw_end = gwordlist_end(wn);
+					for (Gword **gw = gwordlist_begin(wn);
+		                      gw != gw_end;
+		                      gw = gwordlist_next(wn, gw))
 					{
```
linas: indentation

link-grammar/tokenize/tokenize.c
```diff
@@ -1561,23 +1581,25 @@ static bool mprefix_split(Sentence sent, Gword *unsplit_word, const char *word)
...
-		 strcmp("！", word->prev[0]->subword) == 0 )
+	if (strcmp(":", (*gwordlist_at(word->prev, 0))->subword) == 0 ||
+		 strcmp(".", (*gwordlist_at(word->prev, 0))->subword) == 0 ||
+		 strcmp("...", (*gwordlist_at(word->prev, 0))->subword) == 0 ||
+		 strcmp("…", (*gwordlist_at(word->prev, 0))->subword) == 0 ||
+		 strcmp("?", (*gwordlist_at(word->prev, 0))->subword) == 0 ||
+		 strcmp("!", (*gwordlist_at(word->prev, 0))->subword) == 0 ||
+		 strcmp("？", (*gwordlist_at(word->prev, 0))->subword) == 0 ||
+		 strcmp("！",(*gwordlist_at(word->prev, 0))->subword) == 0)
 		return true;
```
linas: !!!

linas:
So I guess you are just replacing a manually-maintained C array by the whizzier vector code? Why did this seem like an important thing to do?

linas:
I tried to read the code .. I did not see the commit that removes the old wordlist code .... for example: `gwordlist_append` -- it seems like there are now two versions in the code? Old and new? and several others, `wordlist_len()`, for example?

link-grammar/tokenize/tokenize.c
ampli: In any case (even if we don't apply this PR) this can be replaced by checking an affix class, something like:
`":.…?!？！" "...": AFEND+; % 2 tokens`
When the first component is checked by `in_afdict_class()` and anything afterwards (here we have only one token afterwards) by a loop. This also makes `AFDICT_BULLETS ` more general since it is then possible to to easily add bullets longer than one characters.

`

ampli:
> So I guess you are just replacing a manually-maintained C array by the whizzier vector code? Why did this seem like an important thing to do?

I thought it could speed up `sane_linkage_morphism()` by allowing pre-allocating of vector memory, as the current method needs an `malloc()` per component. I didn't realize the large overhead and clumsiness of using a "vector library".

It would better to just fix the current `gwordlist_append()` and `gwordlist_len()` to be able to work on pre-allocated vectors and still without a vector header, allocate the vectors (only when p;re-allocated ones are really needed - in 2 places) by a simple `gwordlist_new()` and leave all the rest of the code totally unchanged.

ampli:
> I tried to read the code .. I did not see the commit that removes the old wordlist code .... for example: `gwordlist_append` -- it seems like there are now two versions in the code? Old and new? and several others, `wordlist_len()`, for example?

Note that they are under `#if 0`.

ampli:
I said:
> I didn't realize the large overhead and clumsiness of using a "vector library".

 I forgot to add that I also thought this is a start of adding gword access methods, according to your comment is several files (see `tok-structures.h`). But as we see, this may add much clumsiness and overhead (unless done by small macro-like inline functions as is the case of `connector_*()`.

---
My conclusion:
This PR may be considered as a failed attempt to improve the memory allocation of variable-length arrays.
But this can be done much better by a much smaller change, in a way that may only increase performance.

So I suggest that we close this PR and eventually (after finishing some of the current WIP) I will send a new one to address that.

linas:
OK, that seems reasonable. It did seem clumsy.

> large overhead and clumsiness of using a "vector library".

Yeah. That happens. In C++ the loops would be `for (Gword w: glist)` instead of the verbose, complicated mess that C requires. Some of the overhead in these libraries is because they try to add features to please all possible users, making things slower and more bloated if you are not one of the users. I recall the day I found `boost::signals2` used eleven stack frames to deliver a signal. I promptly built a home-made variant that used zero, that was just good enough for what I needed, and no better.

PR #925, opened 2019-03-03
do_count(): Always check for count overflow

ampli:
In rare cases there is a possibility of overflow in 2 other places.
(I found this possible problem since in my branch with a 32-bit counter these overflows are not too rare.)
If such an overflow gets undetected, havoc may happen due to incorrect counters (I think even an assert() of "walked off the end in list_links").
Regretfully, I cannot provide an example sentence.

Due to the changes in the code here, a redundant assignment (`t->count = total;`) left in PR #911 is gone.


linas:
thanks!

PR #926, opened 2019-03-04
Small efficiency fixes

ampli:
- Several small efficiency improvements:
-- insert_in_cms_table(): Avoid temp in MRU move
-- exprune.c: Put N_deleted in the context struct
-- share_disjunct_jets(): Remove the direction indication at the start
-- build_idiom_word_name(): Remove the use of lg_strcpy()
-- lg_strlcpy(): Replace by a better version
-- Replace safe_strcpy() by lg_strcpy()
-- enumerate_connector_suffixes()/share_disjunct_jets(): save lg_strlcpy
-- read-dict.c: Add "restrict" where it matters
- One bug (currently not harmful) fix:
-- Included in: lg_strlcpy(): Replace by a better version
-  Fixes related to code rot/warning/debug:
-- pack_connectors(): Avoid integer truncation warnings
-- Add debug assertion dassert()
-- enumerate_connector_suffixes()/share_disjunct_jets(): Fix code rot

The total speedup is modest (based on 500 parallel runs as 100x5):
```
 2.085000 -   2.0520 =   -0.0330 +1.6%    data/en/corpus-basic.batch
 6.051000 -   5.9950 =   -0.0560 +0.9%    data/en/corpus-fixes.batch
 1.633000 -   1.6220 =   -0.0110 +0.7%    data/ru/corpus-basic.batch
```
(For long sentences the speedup is even smaller.)

### DIFF
--- a/link-grammar/dict-common/idiom.c
+++ b/link-grammar/dict-common/idiom.c
@@ -119,15 +119,13 @@ static int max_postfix_found(Dict_node * d)
 static const char * build_idiom_word_name(Dictionary dict, const char * s)
 {
 	char buff[2*MAX_WORD];
-	size_t bufsz = 2*MAX_WORD;
 	int count;
 
 	Dict_node *dn = dictionary_lookup_list(dict, s);
 	count = max_postfix_found(dn) + 1;
 	free_lookup_list(dict, dn);
 
-	size_t l = lg_strlcpy(buff, s, bufsz);
-	snprintf(buff+l, bufsz-l, "%cI%d", SUBSCRIPT_MARK, count);
+	snprintf(buff, sizeof(buff), "%s%cI%d", s, SUBSCRIPT_MARK, count);
 
 	return string_set_add(buff, dict->string_set);
 }
--- a/link-grammar/dict-file/read-dict.c
+++ b/link-grammar/dict-file/read-dict.c
@@ -591,7 +591,7 @@ static inline Dict_node * dict_node_new(void)
  * not match the dictionary string s. The matching is dictionary matching:
  * subscripted entries will match "bare" entries.
  */
-static Dict_node * prune_lookup_list(Dict_node *llist, const char * s)
+static Dict_node * prune_lookup_list(Dict_node * restrict llist, const char * restrict s)
 {
 	Dict_node *dn, *dnx, *list_new;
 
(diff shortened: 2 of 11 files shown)


linas:
It won't make a difference here, but the fastest copy routines copy 8 bytes at a time, as 64-bit-ints, and cleanup the rest. which is why in general you want to use the c-library versions of copy routines, not home-rolled ones. I probably should not even mention this to you because (1) you probably already know this, and (2) you will feel compelled to fix this.  

Whoops! Oh noo! too late!

ampli:
I used the exact latest c-library version of Linux (from libbsd 0.9.1)...
> copy 8 bytes at a time, as 64-bit-ints, and cleanup the rest.

The problem that you can do that only if both strings are aligned to 8 bytes.
(Theoretically you can do that if both are misaligned in exactly the same amount, but this is very rare.)

Because the LG lib currently makes an intensive usage of `strlcpy()`, I located **all** the `strlcpy()` sources that are publicly available. None of them directly copy 8 bytes (or even 4) at once. To my very surprise, some are even buggy (not always returning the exact destination length).

Some versions use `memcpy()` for the copy. This looks attractive, because `memcpy()` many times even uses SSE or the like.  I tried the best one of them (that does uses `memcpy()` to copy the source string including its ending NUL). It was much slower. I thought maybe this is due to the function call to `memcpy()`, so I included a copy of `memcp()` too (to be inlined(). It was much slower too.

My conclusion (best on x300 runs for the various tests) was that for the LG case use of `strlcpy()` the best (by much) is to copy byte by byte.
This is because most of the time, all of these are true:
- The destination string is **not aligned** to 8 buyes.
- The source string is short, not even 8 bytes.
- In case it is > 8 bytes, it is seldom a multiple of 8 bytes.

(In a typical use case of `memcpy()`, both buffers are aligned, long and are multiple of 8 bytes.)

BTW, I also found an optimized X86_64 assembly version of `strlcpy()`. I even didn't care to check its license since it seems it is too much to start maintaining our private library of such functions.

One more thing that can be done is to use the system `strlcpy()` if available (using `configure`).
It needs some more efforts so I didn't do it. Especially I didn't have a motivation to do that after I detected that most systems use a slower version.

The efforts here should be most probably concentrated on replacing the string-id mechanism by a special tailored-for-the-task function, that, among other things will use the connector-number. But we need to wait until the dust is over, see the next paragraph...

Today I found a better connector-sequence encoding for `power_prune()` that it seems would make it much faster. This may look unimportant, because currently it is so fast that even if you make it 0-time it would be barely noticeable. But when I added to it additional prunings it became too slow, so a faster version is indeed needed. This means again almost a total rewrite...

linas:
Huh. OK. I stand corrected.

PR #929, opened 2019-04-05
Install linkgrammar jar into local Maven repository

stellarspot:
This is the request from the discussion https://github.com/opencog/relex/pull/282#discussion_r272276153
The proposed fix checks that Maven is present and if yes installs it into local maven repository.

There are several issues with it:
- make install is called using sudo. It leads that the 'mvn install:install-file'  installs linkgrammar jar into the root user local repository. I believe that it is not that was originally required.
It would possible to call it way 'sudo -u $USER mvn install:install-file' but it is not clear for me is it a good idea to use 'sudo' in make script and how it works in Docker.
- There is the way to clean the jar from the local maven repository using 'purge-local-repository' but it requires to have pom.xml file in the directory there Maven is executed.
There is the fix [MDEP-537](https://issues.apache.org/jira/browse/MDEP-537) that claims that it is not necessary to have local maven project to clean the local repository when 'manualInclude' is used but it does not work on my side even it has been fixed in Maven 3.0.0 and I have installed Maven 3.5.2.
So I just use 'rm -rf ~/.m2/repository/org/opencog/linkgrammar/@VERSION@' to remove the linkgrammar jar but really it removes the file from the local user, not the root where this jar is installed.

### DIFF
--- a/bindings/java/Makefile.am
+++ b/bindings/java/Makefile.am
@@ -41,9 +41,21 @@ endif
 
 install-data-hook:
 	${LN_SF} ${javadir}/${java_DATA} ${DESTDIR}${javadir}/linkgrammar.jar
+if HAVE_MVN
+	# Install linkgrammar into the local maven repository
+	mvn install:install-file \
+	  -Dfile=${java_DATA} \
+	  -DgroupId=org.opencog \
+	  -DartifactId=linkgrammar \
+	  -Dversion=@VERSION@ \
+	  -Dpackaging=jar
+endif
 
 uninstall-hook:
 	-rm ${DESTDIR}${javadir}/linkgrammar.jar
+if HAVE_MVN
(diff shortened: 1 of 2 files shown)


linas:
> `sudo -u $USER mvn install:install-file`

Yuck!

> `rm -rf ~/.m2/repository/org/opencog/linkgrammar/@Version@`

Gag!  Clearly, part of the problem, not part of the solution.   Pollution of user home directories with random dot-files installed by random applications is a plague that shows no sign of abating. Its one of the ignored corners of the linux software ecosystem, piling up like garbage on the street.  From what I can tell (I'm not entirely sure), the *nix package manager is designed to avoid this mess.  But clearly we are years if not a decade away before *nix becomes popular. (I haven't used it yet)

In the meantime... clearly, maven install cannot happen at the time of link-grammar `make install`.  It seemed like a good idea, in relex/pull/282#discussion_r272276153 but now, not so much.

PR #931, opened 2019-04-21
Keeping the order of Disjuncts + a surprising speedup

ampli:
For the purpose of  a more efficient Tracon sharing (A WIP that results in a significant speedup) I had to preserve the order of Disjuncts, so it will remain according to the order of the Xnodes. To that end I modified `eliminate_duplicate_disjuncts()` to discard the duplicate disjuncts on the fly instead of rebuilding the Disjunct list from the hash table, thus preserving their order.

For that, I had to add a a `dup_table_next` field to the Disjunct struct. It made it 64 bytes instead of 56 bytes. These caused a surprisingly large speedup, especially for long sentences.
I guess this is a CPU cache effect, because with 56 bytes most Disjunct accesses use 2 cache lines (on X86_64).

I also introduced additional efficiency changes in this code, but most of the saving is still due to the increased Disjunct size.

A benchmark of 25 runs (avg 5 best of 5):
``` text
 2.096000 -   2.0000 =   -0.0960  +4.8%    data/en/corpus-basic.batch
 6.047000 -   5.8210 =   -0.2260  +3.9%    data/en/corpus-fixes.batch
 1.628000 -   1.5650 =   -0.0630  +4.0%    data/ru/corpus-basic.batch
71.680000 -  60.3600 =  -11.3200 +18.8%    data/en/corpus-fix-long.batch
```

### DIFF
--- a/link-grammar/disjunct-utils.c
+++ b/link-grammar/disjunct-utils.c
@@ -346,17 +346,16 @@ static gword_set *gword_set_union(gword_set *kept, gword_set *eliminated)
  */
 Disjunct *eliminate_duplicate_disjuncts(Disjunct *dw)
 {
-	unsigned int h, count;
-	Disjunct *dx;
+	unsigned int count = 0;
 	disjunct_dup_table *dt;
 
-	count = 0;
 	dt = disjunct_dup_table_new(next_power_of_two_up(2 * count_disjuncts(dw)));
 
 	for (Disjunct **dd = &dw; *dd != NULL; /* See: NEXT */)
 	{
+		Disjunct *dx;
 		Disjunct *d = *dd;
-		h = old_hash_disjunct(dt, d);
+		unsigned int h = old_hash_disjunct(dt, d);
 
 		for (dx = dt->dup_table[h]; dx != NULL; dx = dx->dup_table_next)
 		{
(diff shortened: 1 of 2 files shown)


ampli:
> [...] I had to preserve the order of Disjuncts, so it will remain according to the order of the Xnodes.

`setup_connectors()` currently reverses the order of disjuncts. But this doesn't matter.
The new Tracon sharing generation actually needs all the disjuncts that originate from a particular Xnode to remain in one block (i.e. without disjuncts that origin from other Xnodes between them).

linas:
Huh. Thanks!

linas:
I'm seeing no-change for corpus-basic, and 16% speedup for fix-long. This is on an older server-class AMD cpu, which typically had caches twice the size of intel for that era.

ampli:
I guess that this effect is stronger when there is a pressure on the cache, plus maybe a stronger effect on Intel processors (I use i7-4960X, which has a relatively big L3 cache).

linas:
I'm also tracking parse time for a cleaned up version of pride&prejudice, it improved by 11.5% with this commit. And also corpus-failures.batch is 11% faster.

Compared to a year ago, corpus-failures.batch is more than 4x faster!  which is amazing, I think. You've done a fantastic job on this, it's really impressive.

PR #939, opened 2019-04-26
Detecting "mutual" cross links - YA speedup for long sentences

ampli:
As you may remember, I have a WIP for some cross-link pruning. However, it handles only the `null_count==0` case.

This PR adds an extremely simple "mutual" cross link check that works for **any** `null_count`. It is effective mainly for long sentences (partly because it adds some small overhead).

**The speedup is ~20% for the `fix-long` corpus, and ~25% for the `failures` corpus.**

Interestingly, on 2 (similar) sentences in the `basic` corpus it reduces the number of linkages due to discarding all the disjuncts that would create some PP criterion link:
``` diff
*** 85133,85135 ****
  There is going to be an important meeting in January
! Found 3240 linkages (684 had no P.P. violations)
  	Linkage 1, cost vector = (UNUSED=0 DIS= 0.05 LEN=18)
--- 85133,85135 ----
  There is going to be an important meeting in January
! Found 4248 linkages (684 had no P.P. violations)
  	Linkage 1, cost vector = (UNUSED=0 DIS= 0.05 LEN=18)
***************
*** 100003,100005 ****
  There is going to be an important meeting next January
! Found 522 linkages (120 had no P.P. violations)
  	Linkage 1, cost vector = (UNUSED=0 DIS= 0.05 LEN=19)
--- 100003,100005 ----
  There is going to be an important meeting next January
! Found 642 linkages (120 had no P.P. violations)
  	Linkage 1, cost vector = (UNUSED=0 DIS= 0.05 LEN=19)
```

### DIFF
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -450,20 +450,28 @@ static bool possible_connection(prune_context *pc,
 	{
 		return false;
 	}
-	/* If the words are NOT next to each other, then there must be
-	 * at least one intervening connector (i.e. cannot have both
-	 * lc->next and rc->next being null).  But we only enforce this
-	 * when we think its still possible to have a complete parse,
-	 * i.e. before we allow null-linked words.
-	 */
 	else
-	if (!pc->null_links &&
-	    (lc->next == NULL) &&
-	    (rc->next == NULL) &&
-	    (!lc->multi) && (!rc->multi) &&
-	    !optional_gap_collapse(pc->sent, lword, rword))
 	{
-		return false;
+		/* If the words are NOT next to each other, then there must be
+		 * at least one intervening connector (i.e. cannot have both
+		 * lc->next and rc->next being null).  But we only enforce this
+		 * when we think its still possible to have a complete parse,
+		 * i.e. before we allow null-linked words.
+		 */
+		if (!pc->null_links &&
+			 (lc->next == NULL) &&
+			 (rc->next == NULL) &&
+			 (!lc->multi) && (!rc->multi) &&
+			 !optional_gap_collapse(pc->sent, lword, rword))
+		{
+			return false;
+		}
+
+		if ((lc->next != NULL) && (rc->next != NULL))
+		{
+			if (lc->next->nearest_word > rc->next->nearest_word)
+				return false; /* Cross link. */
+		}
 	}
 
 	return true;


linas:
I merged this, but I don't think it can be correct. The PP rules are just enforcing some English grammar constraints.   By contrast, the generation of linkages depends only on connectors/disjuncts. The number of linkages should not change: either a linkage is possible, or it is not.

Now, in the past, you described a different, related phenomenon: one where counting can sometimes report more linkages than the linkage-generator can find.  I don't quite remember why.  But if I try to check this (prior to your patch):
```
!limit=5000
limit set to 5000
linkparser> There is going to be an important meeting in January
Found 4248 linkages (684 had no P.P. violations)
```

linas:
Gahhh.  prior to your patch
```
!limit=5000
limit set to 5000
linkparser> !bad=1
bad set to 1
linkparser> There is going to be an important meeting in January
Found 4416 linkages (684 had no P.P. violations)
```

why does the number go up when I set `bad=1`?  I'm confused...

linas:
Oh, huh. So, *after* applying your patch, and setting `!limit=5000` and `!bad=1` then I also get `4416 linkages (684 had no P.P. violations)` just like before.  So, in fact, everything is working fine.

You can ignore my earlier comments.  It seems like its all good.

ampli:
> You can ignore my earlier comments. It seems like its all good.

I started to write a detailed explanation, but you saved me the effort to complete it...

linas:
> you saved me the effort to complete it...

You're welcome! Once upon a time, I knew the explanation, and it would take me a lot of time to remember it.

I'm amazed. once upon a time, unbounded time corpus-fix-long.batch took 23 minutes. After you fix some connector-id fix, it dropped to 5.  Then after pull #863 it dropped to 3 minutes. Today, its 96 seconds.  This is really quite amazing.

ampli:
BTW, I have some fixes in the queue that make it many tens of percents faster...
And also fixes that make parsing with null_count>0 more than twice faster (for long sentences).

I try to catch up with converting the "dirty" proof-of-concepts of these fixes to clean PRs, but all the time I find new ideas that I feel I must implement before I forget them...

linas:
>  new ideas that I feel I must implement before I forget them

This is indeed one of the best ways to make forward progress.  Almost the only way. (when I stop, months or half-a-year go by without any work.)

PR #945, opened 2019-05-03
Fix MAX_S_PRIME (missing closing paren)

ampli:
It is not in use now.
(Separated from the next PR).


ampli:
That's all - the tracon-id version is ready and I will submit it after this PR is applied (I would like to rebase in order to discard duplicate commits and ensure it applies cleanly).

PR #948, opened 2019-05-13
pack_sentence_for_*(): Fix crash w/ verbosity>=5 on short sentences

ampli:
Repeat by:
`echo x | link-parser -v=5`


ampli:
In the same occasion I added a commit to change the term "trailing hash" that I originally used for the tracon idea (and that is not appropriate anyway for the encoding that is done for pruning).
I still need to fix several comments related it. I will soon send a cleanup PR.

PR #951, opened 2019-05-28
Comment cleanup

ampli:
- Use "tracon" instead of "trailing connectors" etc.
- Fix typos.
- Fix comment rot.
- Add FIXME.

There is a remaining comment rot(and code)  related to the Disjunct size, to be fixed next.


linas:
thanks!

PR #952, opened 2019-05-29
Comment cleanup and expansion

ampli:
- More use of  "tracon" instead of "trailing connectors".
- Fix comment rot.
- Add more comments.
- id_table_check(): Rename to tlsz_check().


linas:
thanks!

PR #953, opened 2019-06-02
Add missing `./autogen.sh` command in "CREATING the system" section

zenflow:


linas:
Thanks, but I will turn this down, for two reasons:
1)  the tarball always comes with a working `configure`
2) users pretty much never have the correct `automake` environment set up, and running `autogen.sh` just corrupts thier system, and then it takes days to walk them out of the mess.   autogen is for the developers only working out of git.

(If you're a user, please download the tarball; it's guaranteed safe; using the git repo is dangerous.)

Issue #161, opened 2015-08-02
Any chances for node.js bindings?


lagleki:
It's been a while since the issue https://github.com/dijs/link-grammar/issues/5 was opened so any plans on making a node.js module for the master version of the library?

linas:
Someone would need to do the work ... patches are welcome. --linas

linas:
FWIW I believe that LG version 5.3.2 now resolves all Apple MacOSX build problems that were blocking a port of the node.js module.

zenflow:
@linas Instead of providing node.js bindings to a binary via FFI (foreign function interface), like `dijs/link-grammar` does, would it be possible to compile link-grammar into pure JavaScript using [emscripten](https://emscripten.org/docs/introducing_emscripten/about_emscripten.html)?

> Emscripten is an Open Source LLVM to JavaScript compiler. Using Emscripten you can:
>
>Compile C and C++ code into JavaScript
>Compile any other code that can be translated into LLVM bitcode into JavaScript.
> Compile the C/C++ runtimes of other languages into JavaScript, and then run code in those other languages in an indirect way (this has been done for Python and Lua)!

I think that would be better because the package would be more portable, and could even be used in web applications. Also I think it may be less complicated to use.

ampli:
Compiling  an linking the LG lib and `link-parser` to `.js` files according to the directions was straightforward and didn't produce any error or warning. I also produced a `link-parser.html` file.

However, when I tried to load `link-parser.html" in Chrome and got the error:
` link-parser.js:1 GET http://localhost/test/link-parser.wasm 404 (Not Found)`
So I guess I need to read some more about how to actually use it.

zenflow:
@ampli What did you do to compile `link-parser`? If you share the code I would be happy to work on this with you (maybe this week when I have a bit more time)

Would the js build be part of the build of this repository? Or would we need to create another repository to pull in the source code from this one, compile it to js, and publish as a package on the npm registry?

ampli:
> What did you do to compile `link-parser`?

Something like:
(Note: It is not an actual script, it is just an **approximate** copy/paste from my screen, also with `cd` commands and/or full paths **omitted**.)
```
source PATHTO/emsdk_env.sh
emconfigure  configure --disable-sat-solver --disable-java-bindings --disable-python-bindings
emmake make
emcc -O3 PATHTO/liblink-grammar.so -o PATHTO/liblink-grammar.js
[...]

PR #954, opened 2019-06-09
Add configuration for Travis CI

zenflow:
As per https://github.com/opencog/link-grammar/issues/161#issuecomment-500248571, adds Travis CI configuration

If this is merged, next steps to complete setting up Travis CI would be to:
1. Register/setup the `opencog` organization and the `link-grammar` repo on [travis-ci.com](https://travis-ci.com)
2. Set up the [Travis CI integration with GitHub](https://github.com/marketplace/travis-ci)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: c
+
+install:
+  - sudo apt-get update
+  - sudo apt-get install -y make m4 gcc g++ autoconf autoconf-archive pkg-config libtool  # link-grammar required dependencies
+  - sudo apt-get install -y swig flex ant graphviz  # link-grammar optional dependencies
+
+script:
+  - ./autogen.sh
+  - ./configure --disable-python-bindings  # TODO: fix tests that fail with python bindings enabled
+  - make
+  - make check
+  - sudo make install
+  - sudo ldconfig
+  - make installcheck
+  - echo "The needs of the many outweigh the needs of the few." | /usr/local/bin/link-parser
+
+notifications:
+  email: false


zenflow:
I had to add the `--disable-python-bindings` flag to the `./configure` step in the .travis.yml file, because otherwise I was getting failing tests. Someone should probably follow up on that!

Here is the Travis job log with the `--disable-python-bindings` flag removed from `the `./configure` step (in a separate branch).

For reference, here is the Travis job log for this PR's branch: https://travis-ci.org/zenflow/link-grammar/builds/543480609?utm_source=github_status&utm_medium=notification

ampli:
> For reference, here is the Travis job log for the latest commit in the branch of this PR: https://travis-ci.org/zenflow/link-grammar/builds/543480609?utm_source=github_status&utm_medium=notification

The tests errors are due to the unwanted "L" suffix difference:

AssertionError: 'LEFT-WALL(0L, 0L) this.p(0L, 4L) is.v(5L, 7L) a(8L, 9L) test.n(10L, 14L) RIGHT-WALL(14L, 14L)\nLEFT-WALL(0L, 0L) this.p(0L, 4L) is.v(5L, 7L) a(8L, 9L) test.n(10L, 14L) RIGHT-WALL(14L, 14L)\n' != 'LEFT-WALL(0, 0) this.p(0, 4) is.v(5, 7) a(8, 9) test.n(10, 14) RIGHT-WALL(14, 14)\nLEFT-WALL(0, 0) this.p(0, 4) is.v(5, 7) a(8, 9) test.n(10, 14) RIGHT-WALL(14, 14)\n' : at ./parses-pos-en.txt:8

I still don't know the reason (have not checked it yet). I will check and send a fix.

linas:
python3 bindings were working fine; I have not checked python2 in a year or so.

linas:
thanks!

ampli:
It seems the reason of this bug is:
``` c
/* Experimental API - subject to changes. */
link_public_api(size_t)
     linkage_get_word_byte_start(const Linkage linkage, WordIdx w);
link_public_api(size_t)
     linkage_get_word_byte_end(const Linkage linkage, WordIdx w);
link_public_api(size_t)
     linkage_get_word_char_start(const Linkage linkage, WordIdx w);
link_public_api(size_t)
     linkage_get_word_char_end(const Linkage linkage, WordIdx w);
```
On python (both 2&3) on fedora 30 (x86_64) `str(some_int)` never appens 'L' even for very big integers and is able to print integers > 2^64. I guess it is configured with the **LongInts**feature.

It seems python2 (at least) on the Travis machine doesn't support **LongInts**, and thus differentiate between `int` and `long`. Since the `get_word_*()` functions return `size_t`,  `long` is used (suffixed by  'L' when converted to string by `str()`).

I see 3 ways to fix this problem:
1. In tests.py, use `int()` to guarantee the result is not 'L' integer.
2. In `linkgrammar.py`, wrap these functions with `int()`.
3. In the C LG library, use `int` returned value instead of `size_t`.

**I'm for solution 3.**

I think it was my design error to use `size_t`, while the rest of the library functions use `int` for values that are usually positive. This is because they use `-1` for errors. The `linkage_get_word_char_*()` **wrongly** return 0 on error, while 0 is a valid position! They should retune -1.

``` c
size_t linkage_get_word_char_start(const Linkage linkage, WordIdx w)
{
        if (linkage->num_words <= w) return 0; /* bounds-check */
        int pos = (int)(linkage->wg_path_display[w]->start - linkage->sent->orig_sentence);
        char *sentchunk = strndupa(linkage->sent->orig_sentence, pos);
        return utf8_strlen(sentchunk);
}
```````````````````````

**So I propose to change the return value of he `linkage_get_word_char_*()` functions to `int`.**

(BTW, a similar problem exists with `size_t  utf8_strwidth(const char *)` which can not return a negative value on error. This can be fixed in this occasion, or be left as is since it is not an official API.)

**Comment by linas** [comment]
> I'm for solution 3.

seems reasonable to me.

**Comment by ampli** [comment]
> I'm for solution 3.

Regarding this solution, now that I come to change it, I note that some time ago (commit 92ce5d7379 some other functions got changed from returning `int` to returning `size_t`:
``` diff
 link_public_api(int)
      linkage_get_num_words(const Linkage linkage);
-link_public_api(int)
+link_public_api(size_t)
      linkage_get_num_links(const Linkage linkage);
-link_public_api(int)
-     linkage_get_link_lword(const Linkage linkage, int index);
-link_public_api(int)
-     linkage_get_link_rword(const Linkage linkage, int index);
+link_public_api(size_t)
+     linkage_get_link_lword(const Linkage linkage, size_t index);
+link_public_api(size_t)
+     linkage_get_link_rword(const Linkage linkage, size_t index);
 link_public_api(int)
      linkage_get_link_length(const Linkage linkage, int index);
 link_public_api(const char *)
-     linkage_get_link_label(const Linkage linkage, int index);
+     linkage_get_link_label(const Linkage linkage, size_t index);
 link_public_api(const char *)
-     linkage_get_link_llabel(const Linkage linkage, int index);
+     linkage_get_link_llabel(const Linkage linkage, size_t index);
 link_public_api(const char *)
-     linkage_get_link_rlabel(const Linkage linkage, int index);
+     linkage_get_link_rlabel(const Linkage linkage, size_t index);
 link_public_api(int)
      linkage_get_link_num_domains(const Linkage linkage, int index);
```
On bad `index`, they return `SIZE_MAX`.

Later (commit 552d6dd854) some of the returned `size_t` got changed to `WordIdx`:
``` diff
-link_public_api(int)
+link_public_api(size_t)
      linkage_get_num_words(const Linkage linkage);
 link_public_api(size_t)
      linkage_get_num_links(const Linkage linkage);
-link_public_api(size_t)
-     linkage_get_link_lword(const Linkage linkage, size_t index);
-link_public_api(size_t)
-     linkage_get_link_rword(const Linkage linkage, size_t index);
+link_public_api(WordIdx)
+     linkage_get_link_lword(const Linkage linkage, LinkIdx index);
+link_public_api(WordIdx)
+     linkage_get_link_rword(const Linkage linkage, LinkIdx index);
 link_public_api(int)
-     linkage_get_link_length(const Linkage linkage, int index);
+     linkage_get_link_length(const Linkage linkage, LinkIdx index);
```

I'm not sure the move to `size_t` instead of the original `int` was a good idea.
However, unless we revert it (which may not be a good idea for stability) it seems my proposed change would create an inconsistency.

So after a second thought it seems to me it would be more consistent to do the following for  `linkage_get_word_byte_*()`:
1. Change their returned value to `WordIdx` instead of `size_t`.
2. Change the returned error of  to `SIZE_MAX` instead of `0` (which is bogus in any case).
3. In the Python test code, change `str(number)` to `str(int(number))` so `L` will never get appended.
4. For later: Document which functions returning `size_t` / `WordIdx`. This is especially important for checking for a returned error (`SIZE_MAX` and not `-1`).

ampli:
> 4\. checking for a returned error (`SIZE_MAX` and not `-1`).

 Maybe the returned error should be `(WordIdx)-1` and not `SIZE_MAX`) so it follows any change in `WordIdx` definition.

ampli:
> python3 bindings were working fine; I have not checked python2 in a year or so.

Currently, if both python 2 & 3 are configured the tests run under python2.
It seems that now the default should (if available). 
I will change that too.

ampli:
> It seems that now the default should (if available).

 It seems that now the default should be python 3  (if available).

ampli:
I'm still not satisfied with my change of proposal....
I said:
>3\. In the Python test code, change `str(number)` to `str(int(number)` so L will never get appended.

But in Python we don't have to follow the`int vs. `size_t` conventions of the LG library, since we don't have i Python`signed` vs. `unsigned` issues like in C, and word indexes are small numbers and will remain so.

So it seems to me better to decide that all the numbers in the LG Python bindings are always `int` (they will anyway be converted  to the needed types when given as arguments). This will keep consistency between Python versions that use `int` & `long` and those that use **LongInts**. This may be a long term change and for the immediate fix I can just implement `str(int(number)`.

linas:
> I'm not sure the move to `size_t` instead of the original `int` was a good idea

It might not have been. About half-way through the effort, I realized that the real goal was to use `WordIdx` wherever we expected a word, `LinkIdx` wherever we expected a link, and so on.  That way, there would be some amount of type-checking.  To do type-checking, one would then be able to:
```
typedef size_t WordIdx;
```
compile, verify no errors, and then
```
typedef short WordIdx;
```
compile, verify that there are still no errors. Then, maybe in the final release, we would set 
```
typedef int WordIdx;
```
This project was never finished.

ampli:
I also miss a type like `sWord_Idx` for "storage WordIdx` to be used for `nearest_word` etc.
So if one would like to support sentences longest than 256 words,, a `typedef sWord_Idx short` could be used.

In any case, I think it will be a good idea to implement this:
> Maybe the returned error should be `(WordIdx)-1` and not `SIZE_MAX`) so it follows any change in `WordIdx` definition.

In the current intermediate state of these plan of checking link/word C types, we have a problem in language bindings that differentiate between `int` and `long`, like Python which is mot configured with LongInts. In the Python case we can fix that by using `int()` on the return values of all the functions that may return `size_t`.

So for now I propose that I just fix the tests to work, replace `SIZE_MAX` by `(WordIdx)-1` in code that I touch, and change the returned values of `linkage_get_word_*()` from `size_t` to `WordIdx` for consistency.

Maybe in version 6 we can return the the original `int` types in the C API.

linas:
Fix as you see fit. I can't provide any particular wisdom here.

ampli:
In `Sentence_s`, it seems `WordIdx sent` is needed instead of the current `size_t sent`.
This seems necessary particularly if "typedef int WordIdx` is to be allowed (else there is a mix of signed/unsigned in things like `w < sent->length`). But what about null_count?

Another WordIdx problem that I don't sure how to handle is function arguments.
For example:
``` c
some_func(int w);
...
WordIdx pos;
some_func(pos);
```
When compiling with `-Wconversion` (in order to check for integer changing sign/truncation) we may get undesired warnings unless the parameter is declared as `WordIdx w`. But for external or long functions (i.e. functions that are not inlined) this would be an unneeded overhead, so cats will be needed, which is cumbersome. Also, some word index variables may get negative value `-1` and this will make handling that value somewhat cumbersome and unreadable.

(So for now I don't have a good solution.)

linas:
> (So for now I don't have a good solution.)

And neither did I ... which is why the project ended, unfinished. 

For the most part, casts that are not inside of inner loops are fairly cheap; I'm willing to make things unmeasurably slower in exchange for better readability.

PR #959, opened 2019-06-13
Always pack the disuncts & connectors

ampli:
Currently, packing, connector encoding and connector sharing all are done together.
For very small sentences, this has a significant overhead so this step is skipped.
However, packing has the benefit that the disjuncts and connectors can be addressed as an array, and even when used as a lists, `uint32_t` indexes can be used instead of pointers  (reducing memory footprint on 64 bit CPUs).

Future code (WIP) takes benefit of this idea (for speedup).

In this PR:
- Always pack, disregarding sentence length. Connector encoding and connector sharing is still skipped for short sentences.
- Use the disjuncts as an array when enumerating them for short sentences.

This change happens to slightly increase the performance for short English sentences (by ~0.2%).

At the same occasion:
- Add comments and clarifications in existing code which is touched.
- Fix a debug printout in `pack_sentence_*()`.

### DIFF
--- a/link-grammar/disjunct-utils.c
+++ b/link-grammar/disjunct-utils.c
@@ -704,7 +704,7 @@ static Connector *pack_connectors(Tracon_sharing *ts, Connector *origc, int dir,
 			else
 			{
 				newc = *tracon;
-				if (NULL == tl)
+				if (!ts->is_pruning)
 				{
 					if (o->nearest_word != newc->nearest_word)
 					{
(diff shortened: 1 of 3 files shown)


ampli:
It crashes in the tests!
Please don't apply yet - I will force-push a fix.

ampli:
I fixed the problem and removed code in `classic_parse()` because packing is always done.
So This PR can be applied now.

PR #964, opened 2019-06-23
Implement power_table-hashing of unique uppercase parts

ampli:
Power table hash slots contain connectors that happen to have the same hash key, and they often have different UC part. This was the case with the fast-matcher tables until my fast-matcher rewrite.

This PR implements the same idea used in fast-matcher also in the power table. A main different implementation detail is that the power table should support entry removals. The solution implemented here is a common one for that case - deleted entry "tombstone".  Like in the rewritten fast-matcher, the fact that each slot contains only connectors with the same UC allows matching only the LC parts of the connectors.

A benchmark of 1500 runs (250 x (best of 6)) gave these minor speedups:
``` 
1.718000 -   1.7130 =   -0.0050  +0.3%    data/en/corpus-basic.batch
4.833000 -   4.8240 =   -0.0090  +0.2%    data/en/corpus-fixes.batch
```
(I didn't make a benchmark of this PR for long sentences yet.)

The motivation behind this PR is a fast-matcher rewrite I am now doing. For that end, I needed the exact matcher disjunct table sizes per word and direction. It turns out the power table has this info, but it is cumbersome and inefficient to extract it due to the said current mix of different UC parts.

BTW, the immediate motivation for totally rewriting the fast-matcher yet again was, in addition to removing a major bottleneck for long sentences, my WIP of power-prune per null-count (combined with incremental count tables - I totally reimplemented the code that I recently removed). But if power-prune is redone for each increased null-count, the fast-matcher tables need to be rebuilt again and again. The O(n^2) `nearest_word` sorting there turned out to be a bottleneck when there are a lot of disjuncts per word (like the case of the Russian sentence from issue #537 ). Among other things, I reimplemented it (still a WIP) as O(n).

### DIFF
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -166,6 +166,15 @@ Connector * connector_new(Pool_desc *, const condesc_t *, Parse_Options);
 void set_connector_length_limit(Connector *, Parse_Options);
 void free_connectors(Connector *);
 
+/**
+ * Compare only the uppercase part of two connectors.
+ * Return true if they are the same, else false.
+ */
+static inline bool connector_uc_eq(const Connector *c1, const Connector *c2)
+{
+	return (connector_uc_num(c1) == connector_uc_num(c2));
+}
+
 /* Length-limits for how far connectors can reach out. */
 #define UNLIMITED_LEN 255
 
--- a/link-grammar/parse/fast-match.c
+++ b/link-grammar/parse/fast-match.c
@@ -169,15 +169,6 @@ static Match_node * add_to_left_table_list(Match_node * m, Match_node * l)
 	return l;  /* return pointer to original head */
 }
 
-/**
- * Compare only the uppercase part of two connectors.
- * Return true if they are the same, else false.
- */
-static bool con_uc_eq(const Connector *c1, const Connector *c2)
-{
-	return (connector_uc_num(c1) == connector_uc_num(c2));
-}
-
 static Match_node **get_match_table_entry(unsigned int size, Match_node **t,
                                           Connector * c, int dir)
 {
@@ -188,7 +179,7 @@ static Match_node **get_match_table_entry(unsigned int size, Match_node **t,
 	if (dir == 1) {
 		while (NULL != t[h])
 		{
-			if (con_uc_eq(t[h]->d->right, c)) break;
+			if (connector_uc_eq(t[h]->d->right, c)) break;
 
 			/* Increment and try again. Every hash bucket MUST have
 			 * a unique upper-case part, since later on, we only
@@ -204,7 +195,7 @@ static Match_node **get_match_table_entry(unsigned int size, Match_node **t,
 	{
 		while (NULL != t[h])
 		{
-			if (con_uc_eq(t[h]->d->left, c)) break;
+			if (connector_uc_eq(t[h]->d->left, c)) break;
 			h = (h + 1) & (size-1);
 			if (h == s) return NULL;
 		}
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -14,15 +14,15 @@
 #include "api-structures.h"
 #include "connectors.h"
 #include "disjunct-utils.h"
-#include "dict-common/dict-common.h"     // For contable
+#include "dict-common/dict-common.h"    // contable
 #include "post-process/post-process.h"
 #include "post-process/pp-structures.h"
-#include "print/print.h"  // For print_disjunct_counts()
+#include "print/print.h"                // print_disjunct_counts
 
 #include "prune.h"
 #include "resources.h"
 #include "string-set.h"
-#include "tokenize/word-structures.h" // for Word_struct
+#include "tokenize/word-structures.h"   // Word_struct
 #include "tokenize/wordgraph.h"
 
 #define D_PRUNE 5 /* Debug level for this file. */
@@ -35,7 +35,7 @@
 #define ppdebug(...)
 #endif
 
-typedef Connector * connector_table;
+typedef Connector *connector_table;
 
 /* Indicator that this connector cannot be used -- that its "obsolete".  */
 #define BAD_WORD (MAX_SENTENCE+1)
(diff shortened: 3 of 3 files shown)


linas:
I'll look at this tomorrow.

I've also got some brand-new dictionaries with brand-new bottlenecks, I don't know where. The dicts are all auto-learned, quite large. They sometimes take a long time to run -- I don't know why, because its not always a combinatoric explosion. Or maybe it is. I can't quite tell.  The timeout code is unreliable; parsing often continues for many minutes after the timeout should have expired.  I'll send you one if interested.

ampli:
>  I'll send you one if interested

Of course I'm interested.

I will try to fix the said yet unknown bottleneck.
(I already know of one bottleneck of learning-dicts in converting expressions to disjuncts - a quadratic list concatenation behavior that I can turn into a linear one. The expression simplification can also be enhanced to produce simpler expressions.)

If it is trivial I will also try to fix the timeout to be more exact (in any case it is part of the fix for #785 that I intend to implement).

ampli:
I wrote:
> I already know of one bottleneck of learning-dicts in converting expressions to disjuncts - a quadratic list concatenation behavior that I can turn into a linear one.

@linas wrote:
> I've also got some brand-new dictionaries with brand-new bottlenecks, I don't know where.

I checked the `micro-disinfo` and  `micro-discrim-unk` dicts.

The problem for the `micro-disinfo` dict  happens in `catenate()` (`build-disjuncts.c`).
(There is no timeout because there is no timeout check while building disjuncts.)
Most of the time in `catenate()` goes on just traversing the lists and checking for the terminating `NULL`.

More than one undefined word is needed for the pruning to leave bigger expressions.
 
Maybe one thing that may contribute to that (I'm not sure in that yet) is unnecessary depth of OR expressions, as you can see with:
`echo 'u1 u2' | link-parser micro-disinfo -v=9 -debug=tokenize.c`
Note the huge number of parens:
```
Debug: string='u1' expr=(((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((([(TY- or TZ-) & TAA+]-5.818 or [TAB+]1.605) or [TAC+ or TAD+]-1.508) or [TAE- & (TAF+ or TAA+)]-6.156) or ...
```
I will try to solve it in several places:
- Improve `catenate()` (by remembering the list ends).
- In `read-sql.c`, add OR expressions at the same level of previous ones.
- For the purpose of regular dicts efficiency, maybe modify `expression_prune()` to remove unnecessary levels or parens.
- Maybe add timeout check somewhere in `build-disjuncts.c`.

The `micro-discrim-unk` is somewhat simpler, and although build_disjuncts() also takes mmuch time for each expression, it gets to the parsing stage.

Due to the way timeouts are implemented (just returning 0 count on lookup miss) it may take very much more than the timeout value to actually terminate the passing (since table lookups are still done). Maybe a "hard timeout" (e.g twice the given timeout) can be introduced, in which all table lookups will return 0 and thus the parsing will be terminated faster (but still it will take some more time, unless implemented by a `longjump`).

linas:
> hard timeout

Don't bother with a hard timeout or anything fancy. These dictionaries are "known broken", they're pathological. The code to merge together connectors into classes has not yet been written, and so there are tens of thousands or hundreds of thousands of these, when there really should be only hundreds.  It will take me several weeks, maybe a month to get the connector merging done right.  The goal was to be able to evaluate "disinfo" to see how it compares, but maybe I need a different strategy for that, right away.

> More than one undefined word is needed

? The `disinfo` should have had 91 different "unknown word" tokens in it.

ampli:
> ? The disinfo should have had 91 different "unknown word" tokens in it.

With ` micro-discrim-unk`, if you have only one unknown word, the expression pruning discards everything.  If you have more and more unknown words, it seems it discards less and less, so building the disjuncts and the parsing take more time.

linas:
micro-discrim should have had 124 unknown words  The idea was that it worked just like UNKNOWN-WORD.n UNKNOWN-WORD.v UNKNOWN-WORD.a in the English dict, except there are more of them.

... I could reduce all these down to just one.... its not really central (at least, not yet), but a nice feature. Just like in the English dict, it tells you which kind it guessed, in the end ... I was planning on using that for feedback ...

Issue #718, opened 2018-03-29
Expression structures are insane


linas:
The construction of the `Exp` structure is nutty, such as the use of the `E_list`structure which is a list that is always exactly zero or one long. That is, `E_list` is not really needed; it could be folded into the Exp struct. I tried to do that in the `left-right` branch in my git repo but ran into all sorts of ugly problems, because the original design was ... insane. Its surprisingly hard to make this seemingly simple change.   It currently fails to build clauses; because of the crazy way that optional connectors are handled in the code.

Basically, I give up; lost too much time on this; this bug report is a fix-me-later report  Half-done code is in `left-right` branch, plus some debugging prints in `lr` branch.

linas:
Turns out that the SAT solver code really likes the n-ary version of the OR_type expression. If `join_alternatives` is altered to create a binary tree, instead of a list, then the SAT solver fails; re-writing `SATEncoder::generate_satisfaction_for_expression()` to work with the binary form of the OR_type makes it ugly.   The `unroll` branch contains this work; if `join_alternatives` is unrolled, then SAT fails, without this deeper redesign.

So, maybe this is a bad idea, in the end.

linas:
Closing. I spend two days hacking on this, and it was an unsatisfying experience.

ampli:
Is there any other benefit for doing such a change, besides being more "sane"?
BTW, rewriting the SAT-parser way of handling expressions is also in my list, so if I get to it there will not be a need for this conversion there.

ampli:
I have just did it in a slightly different way.  I unified `E_list` and `Exp`, i.e. added the `next` field of `E_list` to `Exp` and eliminated `E_list`. (At this step I have already eliminated the original `next` field of `Exp` by changing the dict code to use memory-pool (about 5% speedup on reading dicts. There is also a per-sentence speedup.  I will send PR soon.) As a result of this unification, much of the code of handling `E_list` got removed. Also, expressions then need yet (in addition to saving the original `next` of `Exp`) less total memory because the Exp pointer (that was in `E_list`) got eliminated.

Here is the new `Exp_struct`:
``` c
/**
 * The Exp structure defined below comprises the expression trees that are
 * stored in the dictionary. The expression has a type (OR_type, AND_type
 * or CONNECTOR_type). If it is not a terminal "operand" points to its
 * list of operands (when each of them points to the next one through
 * "next"). Else "condesc" is the connector descriptor, when "dir"
 * indicates the connector direction.
 */
[...]

PR #969, opened 2019-07-10
Dict expression memory pool

ampli:
Implement a dict expression memory pool.

Benefits:
1. Speedup.
1.1. The dict read/close is faster. It can be seen by running `tests.py`, which is dominated by dicts read/close, and runs about 6% faster.
2.2. The rest of the code also runs faster, maybe because the Exp memory is more packed and continuous, and the Exp struct is smaller (`next ` field removed).

A benchmark based on 3000 runs (the dict opening is a small part of the total CPU, especially for English, in which most of the speedup is from the rest of the code):
```
Parameters:   -n 600 -r 5
 1.726000 -   1.6990 =   -0.0270  +1.6%    data/en/corpus-basic.batch
 4.929000 -   4.9060 =   -0.0230  +0.5%    data/en/corpus-fixes.batch
 1.325000 -   1.2990 =   -0.0260  +2.0%    data/ru/corpus-basic.batch
```
2. It allows to add a pointer field in the Exp struct without exceeding 32 bytes (A WIP that eliminates `E_list` uses that).
3. It allows using common functions for expression handling for dict and rest of code, thus allowing code sharing. One of my WIPs concerns expression simplifier, and the function that worked on expression copy could not work on original dict expressions due to their different memory management. Several other functions are different due to historical reasons and still need a unification (TODO).

### DIFF
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -537,7 +537,11 @@ void sentence_delete(Sentence sent)
 	pool_delete(sent->Exp_pool);
 	pool_delete(sent->X_node_pool);
 	if (IS_DB_DICT(sent->dict))
+	{
 		condesc_reuse(sent->dict);
+		pool_reuse(sent->dict->E_list_pool);
+		pool_reuse(sent->dict->Exp_pool);
+	}
 
 	free(sent);
 }
--- a/link-grammar/dict-common/dict-common.c
+++ b/link-grammar/dict-common/dict-common.c
@@ -247,36 +247,6 @@ int delete_dictionary_words(Dictionary dict, const char * s)
 }
 #endif /* USEFUL_BUT_NOT_CURRENTLY_USED */
 
-/**
- * The following two functions free the Exp s and the
- * E_lists of the dictionary.  Not to be confused with
- * free_E_list in dict-utils.c.
- */
-static void free_Elist(E_list * l)
-{
-	E_list * l1;
-
-	for (; l != NULL; l = l1) {
-		l1 = l->next;
-		free(l);
-	}
-}
-
-void free_Exp_list(Exp_list * eli)
-{
-	Exp * e1;
-	Exp * e = eli->exp_list;
-	for (; e != NULL; e = e1)
-	{
-		e1 = e->next;
-		if (e->type != CONNECTOR_type)
-		{
-		   free_Elist(e->u.l);
-		}
-		free(e);
-	}
-}
-
 static void free_dict_node_recursive(Dict_node * dn)
 {
 	if (dn == NULL) return;
@@ -289,7 +259,8 @@ static void free_dictionary(Dictionary dict)
 {
 	free_dict_node_recursive(dict->root);
 	free_Word_file(dict->word_file_header);
-	free_Exp_list(&dict->exp_list);
+	pool_delete(dict->Exp_pool);
+	pool_delete(dict->E_list_pool);
 }
 
 static void affix_list_delete(Dictionary dict)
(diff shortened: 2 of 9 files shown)


link-grammar/api.c
```diff
@@ -537,7 +537,11 @@ void sentence_delete(Sentence sent)
 	pool_delete(sent->Exp_pool);
```
linas: above, you deleted what you're reusing!?

linas:
Thanks.

> A WIP that eliminates E_list

Be aware that this is currently a part of the external API; tehre's only one user of that API, and it can be changed as needed. As long as its possible to have access to Exp's or anything like them, its OK.

link-grammar/api.c
ampli: As I couldn't find `pool_reuse(sent->Exp_pool)` in the code, I guess you referred to line 543:
`pool_reuse(sent->dict->Exp_pool)`.
But this is the dict Exp memory-pool.

BTW:
1. **All** the sentence memory pools are currently deleted at `sentence_delete()`.
But the dict pools survive - for a file dict they are just left untouched, and for a db dict they are reused.
2. I guess it is possible to save CPU by reusing memory-pools over sentences and deleting them on thread exit only.
This can be done by reusing them on `sentence_delete()` instead of deleting them, and deleting them using a registered exit function (I didn't test that so I don't know how much it would save).
A drawback is that if a single sentence happened to use a lot of memory, this memory would not get freed until exit.
I remedy could be adding an additional argument to `pool_reuse()` for how many memory-pool entries to retain.

ampli:
> Be aware that this is currently a part of the external API; tehre's only one user of that API, and it can be changed as needed. As long as its possible to have access to Exp's or anything like them, its OK.

I did the change in a compatible way - my new `Exp_struct` is actually what used to be the E_list:
The eliminated `E_list struct` includes a `next` pointer (for the next operand) and an Exp pointer.
Instead of the Exp pointer I used the full Exp struct, and renamed the said `next` pointer to `operand_next` as follows:
```
     46 struct Exp_struct
     47 {
     48         Exp *operand_next; /* Next same-level operand. */
     49         Exp_type type; /* One of three types: AND, OR, or connector. */
     50         char dir;      /* The connector connects to the left ('-') or right ('+'). */
     51         bool multi;    /* TRUE if a multi-connector (for connector). */
     52         union
     53         {
     54                 Exp *operand_first; /* First operand (for non-terminals). */
     55                 condesc_t *condesc; /* Only needed if it's a connector. */
     56         };
     57         double cost;   /* The cost of using this expression. */
     58 };
```
(Note that the original `next` pointer of `Exp_struct`, used only for memory management, got eliminated in the the recent commit bae7d591e that introduced a dict Exp memory pool).

For the reference, this is the eliminated `E_list_struct`:
```
struct E_list_struct  // renamed to Exp_struct
{
        E_list * next;    // renamed to operand_next
        Exp * e;           // the full previous Exp_struct is used instead
};
```
In addition, for readability I renamed `u.l` to `operand_first` (seems clearer than `operand_list` due to `operand_next`).

So the code conversion is rather straightforward, but in the LG lib most of the specific E_list_struct processing code got eliminated.

**Proposal:**
Add additional API call to make Exp future proof to internal changes (that may be needed according to my WIPs and TODOs):
```
Exp *lg_exp_operand_first(Exp *);
Exp *lg_exp_operand_next(Exp *);
```
When Exp is an opaque structure pointer.

ampli:
Discussion on eliminating E_list moved to #718.

PR #970, opened 2019-07-10
Clean up SYM_AND dict reading

ampli:
Note: This PR is based on #969, which has not been merged yet.
It includes the changes of this #969, and thus **shouldn't be merged** without verifying that GitHub rebases it automatically after this #969 is actually merged. In cased it is not so please tell me and I will force-update it after rebasing.

---
Fix code bloat in the dict reading of SYM_AND.

(I detected that in my WIP of eliminating `E_list` and submit it here separately to make this future PR clearer, since its SYM_AND reading handling contains  a special change related to the `E_list` elimination).


linas:
it didn't autorebase, but I could read the last commit easily enough.  (it should autorebase...)

PR #975, opened 2019-07-20
Remove the corpus statistics code

ampli:
Per discussion at PR #971.


ampli:
I did some sloppy work in removing the corpus stat stuff. A compilation failure resulted.
I just forced-push a fix.

linas:
thanks!

PR #976, opened 2019-07-20
Remove remains of corpus statistics and INFIX_NOTATION

ampli:
EDIT: This fixes also the broken language bindings.

### DIFF
--- a/bindings/lisp/link-interface.cl
+++ b/bindings/lisp/link-interface.cl
@@ -391,17 +391,6 @@
     ((po :foreign-address))
   :returning :int)
 
-
-(ff:def-foreign-call parse_options_set_display_senses
-    ((po :foreign-address)
-     (param :int fixnum))
-  :returning :void)
-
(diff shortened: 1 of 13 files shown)


linas:
thanks!

ampli:
Some corpus stats stuff still remained in `data/tools` and I didn't handle it.
Some of that can maybe converted to manipulate SQL dicts. 
I also found that `wiktionary.pl` needs updating to work fine. It can also be made much better.

linas:
> data/tools

I completely forgot about that directory; haven't used the tools in many many years.  Anyway, I just now removed the cluster-tools. I probably should have remove `wiktionary.pl` but I didn't.  It was never used for anything. It's unusable in its current form - one cannot hand-check the 80K words in the English dict, and automation is unreliable due to the large number of exceptional cases.

linas:
I removed wiktionary.pl just now.

PR #977, opened 2019-07-20
Cost cleanup

ampli:
Mostly per [my post](https://github.com/opencog/link-grammar/issues/783#issuecomment-509979884) at PR #783:
>1. Define cost type (currently "double") so it could be easily changed to float if desired.
I propose `cost_t`.

Not implemented. If needed,  the cost type in `Exp` or `Disjunct` can be changed to `float`, so these are the only places in which cost_t would be of use. (There is no benefit to change the cost variables that are used for calculation (as opposed to storage) to `float`).
Changing all of them to `short` (scaled, so cost 1 is represented internally by 1024) may have enough benefit only if more `Exp` or `Disjunct` space will be needed and there will be no other alternative (not  the case with my current or planned WIPs).

>2. Define print format (currently 2,3,4,6 digits after the decimal point are printed in various places).
I propose `const unsigned int cost_precision = 3;`

Implemented as `cost_max_dec_places`.
I also implemented here `cost_stringify()`, that prints the cost in the shortest possible floating point notation (up to `cost_max_dec_places` decimal places). These is how costs mostly appear in the `en` dict.
E.g.:

current |this PR
-- | --
0.500|.5
0.650|.65
-0.100|-.1

>3. Define function costs_eq(cost1, cost2) (currently comparing cost1-cost2)to an inconsistent value 1E-4 or 1E-5).
I propose `const cost_t cost_epsilon = 1E-5;`.

Implemented as `const double cost_epsilon = 1E-5;`.

I don't know whether to put `cost_stringify()` in the public API along with the "`lg_exp_*()` stuff.
In that case it should maybe called `lg_exp_cost_stringify()`.
For now it is not in the public API.  I will also take `expression_stringify()` out of the API.
But if expression traversing is part of the API, maybe also printing expressions should be.
So if desired in the API I think its API name should be `lg_exp_stringify()`.
I will decide shortly (next PR) but of course I can do it either way if according to what you think.

### DIFF
--- a/link-grammar/dict-common/dict-structures.h
+++ b/link-grammar/dict-common/dict-structures.h
@@ -35,6 +35,9 @@ typedef enum
 	CONNECTOR_type
 } Exp_type;
 
+static const int cost_max_dec_places = 3;
+static const double cost_epsilon = 1E-5;
+
 /**
  * The Exp structure defined below comprises the expression trees that are
  * stored in the dictionary. The expression has a type (OR_type, AND_type
--- a/link-grammar/dict-common/dict-structures.h
+++ b/link-grammar/dict-common/dict-structures.h
@@ -61,6 +61,7 @@ struct Exp_struct
 };
 
 bool cost_eq(double cost1, double cost2);
+const char *cost_stringify(double cost);
 
 /* API to access the above structure. */
 static inline Exp_type lg_exp_get_type(const Exp* exp) { return exp->type; }
--- a/link-grammar/dict-common/dict-structures.h
+++ b/link-grammar/dict-common/dict-structures.h
@@ -60,6 +60,8 @@ struct Exp_struct
 	double cost;   /* The cost of using this expression. */
 };
 
+bool cost_eq(double cost1, double cost2);
+
 /* API to access the above structure. */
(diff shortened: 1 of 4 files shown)


linas:
I don't understand why the cost_stringify function is needed. What's wrong with the printf approach?

ampli:
It is for internal use only (not user API).

The printf approach has these problems:
1. Inconsistent number of decimal places (4 variations were in use).
2. It always prints the longest format for the same number of decimal places.
E.g. (when 3 decimal places used): `0.100` when the dict uses `.1`.
3. Consistent cost format changes need a change in several places.

linas:
The other cleanup is nice, but I don't see the benefit of the printing change. It's easy to read 0.100 -- it doesn't hurt the eyes (at least, not my eyes.)

ampli:
So I can just change cost_stringify() to print it in the full decimal places + leading zero.
Then the benefits of (1) and (3) will still remain, plus the code using `cost_stringify()` already exists so no extra work for getting these small benefits.

I will push-force a fix right away, unless you think there is still a better way to do it.

ampli:
> So I can just change cost_stringify() to print it in the full decimal places + leading zero.

I have forced-pushed a fix to do just that.

PR #978, opened 2019-07-20
Exp cleanup

ampli:
- Fix comments and printouts.
- Rename `expression_stringify()` to `lg_exp_stringify()`.
- Make it to return `const char *` to a static buffer that is freed on sentence delete.
- Add all the lg_exp_*() functions to link-grammar.def. They will still not be exported on Windows (need to
use `link_public_api` for that.)

I intend to later to add `lg_exp_stringify()` to Python in order to implement basic dict expression checks and  testing the dialect stuff (issue #402). BTW, the dialect stuff is already done but needs some cleanup and porting to the new dict code (that got slightly modified from then).

### DIFF
--- a/link-grammar/dict-sql/read-sql.c
+++ b/link-grammar/dict-sql/read-sql.c
@@ -257,11 +257,8 @@ db_lookup_exp(Dictionary dict, const char *s, cbdata* bs)
 
 	if (es != s) free(es);
 
-	if (verbosity_level(D_SQL+1))
-	{
-		printf("Found expression for class %s: ", s);
-		print_expression(bs->exp);
-	}
+	lgdebug(D_SQL+1, "Found expression for class %s: %s\n",
+	        s, lg_exp_stringify(bs->exp));
 }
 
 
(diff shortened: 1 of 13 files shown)


link-grammar/dict-sql/read-sql.c
```diff
@@ -381,8 +378,8 @@ static Dict_node * db_lookup_wild(Dictionary dict, const char *s)
 	{
 		if (bs.dn)
 		{
-			printf("Found expression for glob %s: ", s);
-			print_expression(bs.dn->exp);
+			printf("Found expression for glob %s: %s\n",
+	        s, lg_exp_stringify(bs.dn->exp));
```
linas: indent

link-grammar/prepare/build-disjuncts.c
```diff
@@ -15,8 +15,8 @@
 
 #include "build-disjuncts.h"
 #include "connectors.h"
-//#include "dict-common/print-dict.h"      // print_expression()
-#include "dict-common/dict-structures.h"   // Exp_struct()
+//#include "dict-common/dict-structurtes.h" // lg_exp_stringify
```
linas: ??

ampli:
Both problems due to cut&paste type of code conversion...
Fixed in pr #979.

PR #984, opened 2019-08-02
eliminate_duplicate_disjuncts() performance enhancement

ampli:
While checking whether it is possible to eliminate the related `Disjunct` fields for reuse (it is possible, but with about 10% performance degradation in `eliminate_duplicate_disjuncts()`, so not implemented) I found 2 things that improved the performance (~1% on the English /basic/fixes batches).

### DIFF
--- a/link-grammar/disjunct-utils.c
+++ b/link-grammar/disjunct-utils.c
@@ -120,10 +120,10 @@ static inline unsigned int old_hash_disjunct(disjunct_dup_table *dt, Disjunct * 
 	unsigned int i;
 	i = 0;
 	for (Connector *e = d->left ; e != NULL; e = e->next) {
-		i = (5 * (i + e->desc->uc_num)) + (unsigned int)e->desc->lc_letters + 7;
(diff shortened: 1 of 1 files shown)


link-grammar/disjunct-utils.c
```diff
@@ -371,13 +371,13 @@ Disjunct *eliminate_duplicate_disjuncts(Disjunct *dw)
 				gword_set_union(dx->originating_gword, d->originating_gword);
 
 			count++;
-			*dd = d->next; /* NEXT - set current disjunct to the next one. */
+			prev->next = d->next;
```
linas: prev will be null if the very first disjuncts has a non null dx ... maybe you want to initialize `prev=dw`;` instead?

ampli: Note what I wrote in the commit message.
I guess a comment would be a good idea:
``` c
    350         Disjunct *prev = NULL;  /* The first disjunct is never eliminated. */
```

linas: The comment would be good; I rarely read commit messages, unless there's a problem.

PR #985, opened 2019-08-02
Cleanup t7

ampli:
Mainly comment addition and fixing.
Also add a comment and remove the TLS configuration option per discussions.


linas:
thanks ...

PR #988, opened 2019-08-09
read-sql.c: Extending OR - fix a bug introduced in "Eliminate E_list"

ampli:
Fix a bug in `exp_cb()` introduced in PR #971.


linas:
thanks

PR #993, opened 2019-08-23
Cleanup1+

ampli:
Extracted commits from my current WIP:

- Fix an assert() bug (add fflush()).
- Remove the now unused `string-id.[ch]`.
- Minor efficiency changes.
The `string-set/tracon-set: Centralize the table-full criteria` has a speedup of ~1% on the `basic` batch benchmark.
- Sentence_s: Add num_disjuncts.
It is needed for the upcoming PRs.
- Code cleanups.


linas:
You removed `string-id.c/h` from the makefiles, but the files are still there in git.

PR #996, opened 2019-08-29
Long sentence speedup by connector-word cache

ampli:
This is a hopefully cleaner implementation of the idea demonstrated in PR #861.
It is a reimplementation from scratch - I didn't even look at the original code beforehand.

Now the code seems verifiable by inspecting it (as opposed to the proof-of-concept code of PR #861).
But I also did extensive checks of generating up to 150K linkages and comparing the first 27K of them, for the English `basic` and `fixes` batches, and the Russian `basic` batch. Of course I also checked the `fix-long` and `failure` batches (but not in a detailed automatic checks like I did for the other mentioned batches). 

Results:
-----------
With this patch, the `fix-long` batch runs about 3.5x - 4x times faster, and the `failures` batch about 3.5x faster. Short sentences are not being sped up (but also not slowed down). This may get improved if I implement the FIXMEs in the comments added in this PR (due to a reduced caching overhead). But such implementation is expected to mainly cause an additional speed up for long sentences.

I didn't add a `ChangeLog` entry because I have several more significant speedup fixes (one of them for short sentences too).

### DIFF
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -178,6 +178,20 @@ static Table_lrcnt *find_table_lrcnt_pointer(count_context_t *ctxt,
 	/* Disregard caching in case the table is full - this shouldn't happen. */
 	if (ctxt->table_lrcnt_available_count == 0) return &table_full;
 
+	/* Panic mode: Return a parse bypass indication if resources are
+	 * exhausted.  checktimer is a device to avoid a gazillion system calls
+	 * to get the timer value. On circa-2018 machines, it results in
+	 * several timer calls per second. */
+	ctxt->checktimer ++;
+	if (ctxt->exhausted || ((0 == ctxt->checktimer%(1<<22)) &&
+	                       (ctxt->current_resources != NULL) &&
+	                       //fprintf(stderr, "T") &&
+	                       resources_exhausted(ctxt->current_resources)))
+	{
+		ctxt->exhausted = true;
+		return &lrcnt_cache_zero;
+	}
+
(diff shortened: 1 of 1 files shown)


ampli:
Forced push to fix 2 comment typos.

linas:
Wow. I just measured this. Wow.

linas:
Maybe  in a few weeks, I should put together another release? Or do you have something else you want to push out?

ampli:
I think it should be released even sooner.

I saw a request to remove  LG from Ubuntu 19.04 because it totally malfunctions on 32-bit (x86 and arm). The tests on 32-bit processors detect that - many tests fail.

(I responded, see https://bugs.launchpad.net/ubuntu/+source/link-grammar/+bug/1842134).

I check and there is a major initialization bug that causes random results due to a field that happens to remain uninitialized on 32 bit (my bug). There are also some more minor 32-bit problems, and while checking it on a 32-bit Ubuntu 19.04 I also encountered an unclean compilation with several problems.

I fixed all of that and intent to send a PR soon.

BTW, I didn't get his Python and SQLiite problems. I will need to inspect his logs more carefully. In particular, he gets a SWIG compilation warning that I don't get, even though I did my tests on the exact same system version  version, and only 2 days after his test. Maybe their build system adds more warning flags.

linas:
aieee! It would be nicer if he had reported the bugs upstream.  Let me know when you're ready; I'll need 24-48 hours to turn things around.  Please post updates to launchpad, I'm out of town right now and can't log into it to post comments.

PR #997, opened 2019-08-30
link-parser option related fixes

ampli:
- command-line.c: Don't allow values > 1 for booleans
It applies to things like:
``` text
linkparser> !walls=2
walls set to 2
```
- command-line.c: Use a pre-defined help message
It just looks better.

- link-parser.c: Don't print version on -verbosity=0
Because it is requested to be with zero verbosity.

- link-parser.c: Add --quite
See: https://www.gnu.org/prep/standards/standards.html#Option-Table
This is a common GNU program flag. Add also --silent as described there.

### DIFF
--- a/link-parser/link-parser.c
+++ b/link-parser/link-parser.c
@@ -760,11 +760,14 @@ int main(int argc, char * argv[])
 
 	check_winsize(copts);
 
-	prt_error("Info: Dictionary version %s, locale %s\n",
-		linkgrammar_get_dict_version(dict),
-		linkgrammar_get_dict_locale(dict));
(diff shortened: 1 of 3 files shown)


link-parser/link-parser.c
```diff
@@ -698,6 +698,8 @@ int main(int argc, char * argv[])
 		language = argv[1];
 	}
 
+	/* Process options used by GNU programs. */
+	int quite_start = 0; /* Iff > 0, inhibit the initial messages */
```
linas: quiet, not quite ...

linas:
Thanks!

PR #1000, opened 2019-09-03
Fix 32-bit problems

ampli:


ampli:
I tested these fixes on 32-bit Ubuntu 19.04.
(The last fix - for fchdir(), is not related to 32-bit but it fixes a warning on Ubuntu 19.04.)

I was not able to configure the system minisat2. Need to check why, but this is not critical.
The bundled one got configured after installing `zlib1g-dev`. It will be a good idea to remove these dependency from it because it is actually unused in the minisat2 library (it is needed in a supplied program that uses the library).

ampli:
Just appended Ubuntu(/Debian) related README fixes.

ampli:
I have just posted an update on https://bugs.launchpad.net/ubuntu/+source/link-grammar/+bug/1842134 .

linas:
Those changes seem harmless; it's hard to imagine how they lead to 32-bit bugs. But whatever.

linas:
I'd like to keep the ARM build; there is someone who is trying to port opencog to "termux", which is some kind of linux environment for android.

ampli:
> Those changes seem harmless; it's hard to imagine how they lead to 32-bit bugs. But whatever.

The following bug was fatal and caused random results. Many of the tests failed.
- connector_new(): Bug fix: "shallow" is uninitialized on 32-bit systems

A side effect of fixing it is a slight speedup for short sentences.

It turns out I had a problem in my try to post yesterday an update to launchpad.
So I posted an updated update just now, please see in the link above.

ampli:
> I'd like to keep the ARM build;

The OP asked to remove it from `armhf` which is 32-bit.

>                                                   there is someone who is trying to port opencog to "termux", which is some kind of linux environment for android.

Will he use android on a 32-bit CPU?

Any any case I asked not to remove it but just rebuild it from the upcoming release in a few days.
However, I didn't test if the plenty of linkage errors on `armhf` effect the LG library operation (still need to install an `armhf` guest machine.)

I also would like to recheck the current master repository on Windows and Mac, because there have been many chnages froim the last time I did that. I will update you later today.

linas:
> `.strtab` too small  or toolchain bug

I'd say both.  Linkers job is to link, and that means setting the right size for segments.

> Will he use android on a 32-bit CPU?

Don't know. I'll ask.

ampli:
> > `.strtab` too small  or toolchain bug
> 
> I'd say both. Linkers job is to link, and that means setting the right size for segments.

The one who said that also suggested that embedded processors have small memory so by default the linker may try to leave enough space for possible other things that need to run, and this may be a mater of adjusting linker flags for big images.
But then I don't understand how the resulted binaries work anyway (as appears from posts on that matter). In our case many of the tests on `armhf` succeed, which means it runs despite of these alarming linkage problems.

ampli:
> I also would like to recheck the current master repository on Windows and Mac, because there have been many changes from the last time I did that. I will update you later today.

It turns out the Windows compilation is broken in several places, and this happened in previous releases. I'm preparing a PR to fix that. After that I will check it on MacOS.

PR #1003, opened 2019-09-04
Fixes for Windows

ampli:
My fixes and additions in the previous release broke several things in the Windows build.
Here I fixed all of them.
I also fixed some warnings that we don't enable on GCC (or maybe GCC doesn't care, like missing "restrict" in function declaration).
In addition I fixed the Windows "~" support that I recently added.
I validated that it passes the Python tests.
I didn't test the others because there is no build/run system to do that yet.

I still need to test on Mac and will update on that soon (with a PR if needed).
Reference: PR #1000.

### DIFF
--- a/link-grammar/utilities.c
+++ b/link-grammar/utilities.c
@@ -49,7 +49,8 @@
 #ifdef _WIN32
 void lg_strerror(int err_no, char *buf, size_t len)
 {
-	return strerror_s(buf, len, err_no);
+	if (strerror_s(buf, len, err_no) != 0)
+		strerror_s(buf, len, errno); /* errno got set by previous strerror_s() */
 }
 #else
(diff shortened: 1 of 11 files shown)


link-grammar/utilities.c
```diff
@@ -49,7 +49,8 @@
 #ifdef _WIN32
 void lg_strerror(int err_no, char *buf, size_t len)
 {
-	return strerror_s(buf, len, err_no);
+	if (strerror_s(buf, len, err_no) != 0)
+		strerror_s(buf, len, errno); /* errno got set by previous strerror_s() */
 }
```
linas: OK, but `errno` is an unprotected global, and so this can race in threads. Of course, very unlikely but the principle of it ...

PR #1006, opened 2019-09-07
More Windows

ampli:


ampli:
I just forced-push a fix. I hope that now there are no conflicts.

PR #1007, opened 2019-09-07
Fix a bug in setting UNLIMITED-CONNECTORS

ampli:
- set_condesc_length_limit(): Fix off-by-1 bug in restart_cn

This is a bad bug when it has an effect. See the commit message for details.
However, it most probably didn't affect Linux systems with the standard libc.
Also, I don't know the probability that this bug will occur on a particular `qsort()` implementation, maybe it is small. In any case, it didn't happen up to now on the systems on which I tested LG, including Windows 8 and 10, Cygwin, FreeBSD, several Linux distributions, MinGW32 and MinGW64 and oldest MacOS systems, and somehow it doesn't happen on 5.6.2 even on the current Mac machine I use (OS version Mojave).

More extensive test suites may detect such bugs. This one could be detected by a test suite that verifies that UNLIMITED-CONNECTORS are set properly (this need some help from the library, in that case there is `link-parser` -v=101` that prints all the connectors and their length limit).

META info:
This PR is intended to be applied after the #1006 (More Windows), because both update the ChangeLog. It also contains 2 commits from #1006 with the same hash and I hope they will get automatically cancelled out. In addition it is not rebased on the current master, but it doesn't suppose top touch affected files.
The ChangeLog in this PR also refers this PR itself (#1007) and I hope I will not create a horrible GitHub loop.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,8 @@ Version 5.7.0 (XXX 2019)
  * Incompatible change to the Exp traversal API.
  * Remove the obsolete and unsupported "corpus statistics" code.
  * Major performance improvement (3x-4x) for long sentences. #996
+ * Fix broken build on Windows.
+ * Convert Windows build to Visual Studio 2019.
 
 Version 5.6.2 (24 June 2019)
  * Bug-fix the SQL-backed dictionary.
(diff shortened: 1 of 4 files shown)


ampli:
If there are conflicts after #1006 is applied, please don't apply this PR #1007, and I will rebase it and make a forced-push.

PR #1008, opened 2019-09-13
Cleanup t8

ampli:
Last cleanups for now.
The rest of fixes are left to the release after the next one...

Included here:
- ChangeLog: Update on the 32-bit fix

On issue #1000 I said:
> I have just posted an update on https://bugs.launchpad.net/ubuntu/+source/link-grammar/+bug/1842134

Now a new release can be done and I will then update on launchpad.


linas:
Thanks. New release in an hour or two, then.

Issue #1009, opened 2019-09-19
LinkGrammar.getLinkageSense and LinkGrammar.getLinkageSenseScore were removed from JNI but still present in Java API declaration


vsbogd:
`LinkGrammar.getLinkageSense` and `LinkGrammar.getLinkageSenseScore` java API method implementations were removed by commit https://github.com/opencog/link-grammar/commit/5249ddb99fb8fd33731b518029c31f316c1c552c#diff-acf5e8f599bb0954fb0aa69ba6293ce1 but they are still present in Java API declaration: https://github.com/opencog/link-grammar/blob/c304c4398557b8849da42e5a40c72cdb13db0758/bindings/java/org/linkgrammar/LinkGrammar.java#L120-L122

Because of this Java compiler cannot find that methods are absent during compilation, see https://github.com/opencog/relex/issues/283

vsbogd:
In addition probably it makes sense to remove `storeSense` option from `LGConfig` API:
https://github.com/opencog/link-grammar/blob/c304c4398557b8849da42e5a40c72cdb13db0758/bindings/java/org/linkgrammar/LGConfig.java#L71-L78

linas:
closing; the correct fix is to rip this stuff out of relex.

PR #1010, opened 2019-09-19
Java bindings: Remove senses API

ampli:
This addresses issues #1009 and opencog/relex#283.


ampli:
A new release may be in order.
I can add some other non-critical cleanups if it is desired to release something with more changes...

linas:
thanks. I don't think this is urgent enough to warrant a new release. relex hasn't used the sense-stuff since forever.

vsbogd:
Thanks @ampli , yes, release is not needed as @linas remove usages from Relex code base.

PR #1016, opened 2019-10-09
Remove count sharing

ampli:
More and more bugs got found when implementing a more aggressive pruning
(no examples of problematic existing batch sentences in the latest release).
Fixing them increased the overhead and more bugs have discovered.
It seems that the idea is not sound.
It is also hard to understand and maintain.

Its speedup potential is limited (tens of percents), and it prevents
implementing changes that would increase speedup several times more than
that.

**So this idea is dropped until a full-proof way to do that is found.**

**EDIT**: The commit message has more details.

---
I found other ways to speed up parsing, and also to speed up parsing with nulls.
Some of them I already implemented (WIP) and intend to send PRs soon.

### DIFF
--- a/link-grammar/disjunct-utils.c
+++ b/link-grammar/disjunct-utils.c
@@ -473,7 +473,6 @@ void word_record_in_disjunct(const Gword * gw, Disjunct * d)
 	}
 }
 
-
 /* ================ Pack disjuncts and connectors ============== */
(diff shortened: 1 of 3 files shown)


ampli:
Comments:
- The commit message is more detailed.
- This is the second time I discard this idea. The first one was at PR #900.

PR #1017, opened 2019-10-09
Update docs

ampli:
1. In case someone would like to help in the LG library development, here is an update of the debug read readme.
2. In the main readme:
- An update on the problem with Anaconda on Mac.
- Remove most references to python2.
3. In the same occasion add a pseudo-test `@` to suppress the test warning message, in order to prevent the need to comment out this message in `link-parser.c` when running benchmarks on features that are enabled by `-test=`.


linas:
thanks!

PR #1019, opened 2019-10-12
Farthest word

ampli:
Changes here:
- Add comments on the linkage count optimization.
- Convert `length_limit`to word number (`farthest_word`).
- Optimize `start_word`/`end_word`.

Benchmark results on the various batches (speedup multiplier):
`en/basic`/`en/fixes`: ~-1%
`ru`: ~+2%
`en/fix-long`: ~+10%
`en/failures`: ~+18%

The slowness for `en/basic`/`en/fixes` is unfortunate. However, old WIPs contain trimming of `length_limit` (now `farthest_word`) in `power_prune()` and elsewhere, and it is slightly more deficient to trim  `farthest_word` than `length_limit` so hopefully this will get improved when these WIPs are converted to PRs.

### DIFF
--- a/link-grammar/parse/preparation.c
+++ b/link-grammar/parse/preparation.c
@@ -29,16 +29,22 @@
 
 /**
  * Set c->nearest_word to the nearest word that this connector could
- * possibly connect to.  The connector *might*, in the end,
- * connect to something more distant, but this is the nearest
- * one that could be connected.
+ * possibly connect to.
+ * The connector *might*, in the end, connect to something more distant,
+ * but this is the nearest one that could be connected.
+ * Also recalculate length_limit to be the farthest word number that could
+ * be connected.
  */
-static int set_dist_fields(Connector * c, size_t w, int delta)
+static int set_dist_fields(Connector * c, size_t w, int delta, int sent_len)
 {
 	int i;
 	if (c == NULL) return (int) w;
-	i = set_dist_fields(c->next, w, delta) + delta;
+	i = set_dist_fields(c->next, w, delta, sent_len) + delta;
 	c->nearest_word = i;
+	int farthest_word = w + (delta * c->length_limit);
+	if (farthest_word < 0) farthest_word = 0;
+	if (farthest_word >= sent_len) farthest_word = sent_len - 1;
+	c->farthest_word = farthest_word;
 	return i;
 }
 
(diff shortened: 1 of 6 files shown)


link-grammar/parse/preparation.c
```diff
@@ -29,16 +29,22 @@
...
+ * be connected.
  */
-static int set_dist_fields(Connector * c, size_t w, int delta)
+static int set_dist_fields(Connector * c, size_t w, int delta, int w_clamp)
 {
 	int i;
 	if (c == NULL) return (int) w;
-	i = set_dist_fields(c->next, w, delta) + delta;
+	i = set_dist_fields(c->next, w, delta, w_clamp) + delta;
 	c->nearest_word = i;
```
linas: This seems to be a simple recursive call. If it's performance critical, then it could be converted into a loop.

(It's almost a tail call; does gcc ever convert tail calls into loops?)

link-grammar/parse/preparation.c
```diff
@@ -58,8 +64,9 @@ static void setup_connectors(Sentence sent)
 		for (Disjunct *d = sent->word[w].d; d != NULL; d = xd)
 		{
 			xd = d->next;
-			if ((set_dist_fields(d->left, w, -1) < 0) ||
-			    (set_dist_fields(d->right, w, 1) >= (int) sent->length))
+			if ((set_dist_fields(d->left, w, -1, 0) < 0) ||
+			    (set_dist_fields(d->right, w, 1, (int)(sent->length-1)) >=
+			     (int) sent->length))
 			{
```
linas: This `if` is always false, because `set_dist_fields` cannot return anything less than 0 or greater than sent length.

linas:
Off-topic, but I noticed that, on my system, some of my tests spend 5% to 15% cpu in system time, instead of user.  This makes no sense to me, as LG makes almost no system calls ... except for some "infrequent" calls to get the clock (for timeouts).  Did all of these perf improvements mean that we are checking the timer much too often?

link-grammar/parse/preparation.c
ampli: Note that what is clamped to the range [0, sent_length) is `farthest_word`, while the returned value is `i` (actually `nearest_word`), which may get out of this range in the statement:
`i = set_dist_fields(c->next, w, delta, w_clamp) + delta;`

link-grammar/parse/preparation.c
ampli: gcc may convert tail calls into loops, but it depends on many things and inspecting the result is needed in order to find out.
In our case, `set_dist_fields()` is not converted to loop, and the first X calls are not even getting inlined (as I falsely hoped).

In the various batch profiling it takes 0.4-0.7% of the total CPU, and this percentage will get enlarged (by much) when further optimizations to the rest of the code are implemented. So eventually converting it to a loop (and optimizing it further) may a good idea.

ampli:
Indeed the timer checks are problematic. For  some sentences they are too infrequent, causing a too slow panic resolution. For some other sentences it is too high, resulting in an unneeded overhead.

Possible solutions (maybe both together):
1. Make it even less frequent and use `longjump() on timeout.
2. Dynamic check interval setting to make the check frequency a predefined constant.

linas:
How about 
3.  Use a timer interrupt. That way, we can eliminate *all* calls to check the elapsed time. .. and if we're doiong that, remove the calls to check mem usage.

link-grammar/parse/preparation.c
linas: oh right, I misread the `delta`.

Issue #1020, opened 2019-10-12
Remove time and mem usage calls


linas:
Following discussion on #1019:

on my system, some of my tests spend 5% to 15% cpu in system time, instead of user. This makes no sense to me, as LG makes almost no system calls ... except for some "infrequent" calls to get the clock (for timeouts). Perhaps we are checking the timer much too often? Hard to say.

Propose -- Use a timer interrupt. That way, we can eliminate all calls to check the elapsed time. .. and if we're doing that, we may as well remove the calls to check mem usage.

Some notes about RAM usage:
1) RAM is no longer a constraint on modern systems (the original LG code dates to 1995, when it was.)
2) In the highly unlikely event that LG runs on a memory-constrained system (100's of MB or less) then having a special malloc library for dealing with mem constraint makes more sense.  Also, rather than checking for `nullptr == malloc()` I instead I envision such a special lib would throw an exception which LG would need to catch and cleanup. Which  would require new code & design, but really I strongly doubt that anyone will ever run LG on a small RAM system.  

So basically, this proposal is to remove `resources.c` and `resources.h` entirely.

ampli:
> Perhaps we are checking the timer much too often? Hard to say.

It turns out this is an old problem in `post_process_lkgs()` and `prepare_to_parse()`.

In `post_process_lkgs()`:
``` c
1176	if ((63 == in%64) && resources_exhausted(opts->resources)) break;
...
1203  if ((15 == in%16) && resources_exhausted(opts->resources)) break;
```
**For now, these numbers should just be increased.**
Eventually these checks there should be dropped altogether - when I finish my total rewrite of post processing. The current code there is written in a very wasteful way and I hope to get there 3 order of magnitude less CPU, and maybe move part of it to the pruning stage (I think I found how to implement that), a thing that will solve other problems too.

In  `prepare_to_parse()`:
``` c
    150         for (i=0; i<sent->length; i++)
    151         {
    152                 sent->word[i].d = eliminate_duplicate_disjuncts(sent->word[i].d);
    153 
    154                 /* Some long Russian sentences can really blow up, here. */
    155                 if (resources_exhausted(opts->resources))
    156                         return;
    157         }
```
[...]

PR #1022, opened 2019-10-13
Reduce timer check rate

ampli:
This PR is per issue #1020.

The batch benchmark shows speedup of 2-3% for short `en` sentences,
and negligible saving for `ru` and long `en` sentences.

I didn't have a workload (dict + corpus) in which the system time is anything significant to begin with, so I don't know how much the saving is in such a case.

### DIFF
--- a/link-grammar/parse/preparation.c
+++ b/link-grammar/parse/preparation.c
@@ -151,9 +151,18 @@ void prepare_to_parse(Sentence sent, Parse_Options opts)
 	{
 		sent->word[i].d = eliminate_duplicate_disjuncts(sent->word[i].d);
 
+#if 0
+		/* eliminate_duplicate_disjuncts() is now very efficient and doesn't
+		 * take significant time even for millions of disjuncts. If a very
+		 * large number of disjuncts per word or very large number of words
+		 * per sentence will ever be a problem, then a "checktimer" TLS
+		 * counter can be used there. Old comment and code are retained
+		 * below for documentation. */
+
 		/* Some long Russian sentences can really blow up, here. */
 		if (resources_exhausted(opts->resources))
 			return;
+#endif
 	}
 	print_time(opts, "Eliminated duplicate disjuncts");
 
--- a/link-grammar/post-process/post-process.c
+++ b/link-grammar/post-process/post-process.c
@@ -1156,6 +1156,8 @@ void post_process_lkgs(Sentence sent, Parse_Options opts)
 		return;
 	}
 
+#define TCD 512 /* timer checking divisor */
+
 	/* (optional) First pass: just visit the linkages */
 	/* The purpose of the first pass is to make the post-processing
 	 * more efficient.  Because (hopefully) by the time the real work
@@ -1173,7 +1175,7 @@ void post_process_lkgs(Sentence sent, Parse_Options opts)
 
 			post_process_scan_linkage(pp, lkg);
 
-			if ((63 == in%64) && resources_exhausted(opts->resources)) break;
+			if (((TCD-1) == in%TCD) && resources_exhausted(opts->resources)) break;
 		}
 	}
 
@@ -1200,7 +1202,7 @@ void post_process_lkgs(Sentence sent, Parse_Options opts)
 		N_linkages_post_processed++;
 
 		linkage_score(lkg, opts);
-		if ((15 == in%16) && resources_exhausted(opts->resources)) break;
+		if (((TCD-1) == in%TCD) && resources_exhausted(opts->resources)) break;
 	}
 
 	/* If the timer expired, then we never finished post-processing.
--- a/link-grammar/resources.c
+++ b/link-grammar/resources.c
@@ -142,15 +142,18 @@ bool resources_memory_exhausted(Resources r)
 #define RES_COL_WIDTH 40
 
 /** print out the cpu ticks since this was last called */
-static void resources_print_time(int verbosity_opt, Resources r, const char * s)
+GNUC_PRINTF(2,0)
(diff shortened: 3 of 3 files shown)


linas:
This made no difference at all in the user/sys ratio: its still about 10%  My test corpus is a cleaned up variant of Austen's Pride & Prejudice The zipped copy is 165K let me see if I can share it somewhere...

linas:
https://file.io/2QAIqn don't know how long it will be there.

linas:
oh it's deleted after the first download. its a bzip2 text file

ampli:
Do you just use:
`link-parser < pandp-union.txt >/dev/null`

i.e. with default short_length, timeout and limit options?

ampli:
>This made no difference at all in the user/sys ratio

Yes, I observed the 2%-3% speedup **only** for the short sentence batches.

With `link-parser < pandp-union.txt >/dev/null`, I get the following:
`167.20u 32.39s 3:34.31e 377624Kb`

This is about 20% system time.

`strace` show  something like the following for each sentence which is not very short:
```
mmap(NULL, 134221824, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fad5f4a0000
...
munmap(0x7fad5f4a0000, 134221824
```

This is the allocation of  `Table_connector` in `count.c`.
What happens is that `malloc()` uses `mmap()` for big allocations, and `free()` always returns them to the system using `munmap()`. This has a high system overhead.

As a test, I used a `static TLS` variable in `init_table()` to hold the biggest allocation, so it is freemalloptd only upon `link-parser` exit.  As a result, the system time got to be less than 0.5%:
`158.72u 0.45s 2:39.68e 369856Kb`

Very strangely, even though the system time for the long sentence batches is very low,  with this hack the batch benchmarks show 6%-7% speedup on `fix-long` and  `failures`. Is this a cache effect due to the kept table? (BTW, no speedup at all for the `basic`/`fixes` batches).

This means that with this change, once the library parses a sentence with more than 47 tokens, it starts to permanently hold 132MB until it exists.

I guess that this may be fine. So I will send a PR. It looks like a hack, but if needed maybe it can be refined.

linas:
> default short_length, timeout and limit options?

Yes.

> mmap

Wow!

> Very strangely, even though the system time for the long sentence batches is very low, with this hack the batch benchmarks show 6%-7% speedup on fix-long and failures

Rather than caching, I am guessing it is a TLB effect: the TLB entries needed to map that RAM, once loaded, just stay there.  I'm guessing that most of the system time is actually due to TLB misses in the kernel itself.  (I don't know why the kernel needs to touch that memory. Maybe to clear it? Maybe spectre/meltdown work-arounds??)

> it starts to permanently hold 128MB until it exists.

Seems acceptable.

> It looks like a hack

Well, add a comment explaining why it's needed.

----
If you have time, you can play with hugepages. `sudo sysctl vm.nr_hugepages=12` and then maap yourself: 
```
void * mem = mmap(0, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB | MAP_POPULATE, 0, 0);
```
I suspect we should NOT do this in LG, but it could be interesting to see what happens...

ampli:
> Maybe to clear it?
> Maybe to clear it?

The kernel clears mmap'ed memory as a security measure.
`calloc()` of big memory allocations are said to consider that and be faster than `malloc()`+`memset()` (but I couldn't get speedup with that - I still need to investigate way).

> Well, add a comment explaining why it's needed.

I have just added such a comment.

> it could be interesting to see what happens.

What may happen? Do you expect it will have much less overhead than using regular TLB size?

BTW, there are still plenty of similar big allocations of 32-64MB each, but for some reason they have very small overhead. Maybe this may hint on the nature of the overhead in 128MB allocations.

linas:
My machine has 1024 dTLB emtries, and so, with 4K pages, there will be dTLB misses when the working set is larger than 4MB.   Which is kind-of crazy, because my L3 cache is 16MB. 

I tried the `HUGETLB` trick (after #1024) and I get a 2.5% speedup on `pandp-union` and maybe a 0.5% speedup on `corpus-failures`.

Implementing `mmap(HUGETLB)` correctly would add a lot of complexity, since not all OS'es support HUGETLB, and when they do, its disabled unless the sysadmin enables `vm.nr_hugepages` ...

linas:
FYI, this was the diff on top of #1024 
```
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -12,6 +12,7 @@
 /*************************************************************************/
 
 #include <limits.h>
+#include <sys/mman.h>
 
 #include "link-includes.h"
 #include "api-structures.h"
@@ -89,7 +90,7 @@ static void init_table(count_context_t *ctxt, size_t sent_len)
 
        if (ctxt == NULL)
        {
-               if (kept_table) free(kept_table);
+               if (kept_table) munmap(kept_table, 1UL<<log2_table_size);
                kept_table = NULL;
                log2_table_size = 0;
                return;
@@ -124,8 +125,10 @@ static void init_table(count_context_t *ctxt, size_t sent_len)
        {
                log2_table_size = shift;
 
-               if (kept_table) free(kept_table);
-               kept_table = malloc(ctxt->table_size * sizeof(Table_connector*));
+               if (kept_table) munmap(kept_table, 1UL<<log2_table_size);
+               kept_table = mmap(0, ctxt->table_size * sizeof(Table_connector*),
+                       PROT_READ | PROT_WRITE,
+                       MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB | MAP_POPULATE, 0, 0);
 
                // atexit(free_kept_table);
        }
```

PR #1025, opened 2019-10-17
Extended power prune, part 1

ampli:
This PR implements 2 new ideas:
1. If the pruning indicates that there cannot be a parse with the requested number of nulls, then skip the parse. Without improved pruning, it doesn't cause a speedup because it skips parsing for sentences with a "lightweight" parse anyway. As an slight optimization, the pruning stops when this condition is detected. The sentences that are affected are relatively short. This optimization is disabled with `-test=parse-always`.
2. In power-prune, add a `possible_connection()` step of finding how many nulls will be created due to cross-link considerations.

Speedup batch benchmarks (in parens, increase due to the parse-skipping optimization of (1) above,
found by parallel runnings with and w/o  `-test=parse-always`).
`fix-long`: ~18% (~1%, only 3 short sentences are affected).
`failures`: ~35% (~6.5%)
`pandp-union`: ~15% (~4.5%)

This PR is "part 1". There are 3 additional power-prune extra pruning ideas that are in a WIP stage,
with some promising results but nothing solid to talk about.

A problem with this optimizations is that I fear only I understand them due to not enough documentation (and I may forget the details over time even if now they seem to be obvious). I included the comments from the proof-of-concept code, which uses the ideas in some other manner (hopefully included in the next parts after refined). They include an ASCII diagram. Real diagrams, one for each optimization case, would be much better. However, I don't have any experience in producing such diagrams.

@linas, if some optimization is not clear after reading the comments, please ask me for clarifications, and maybe we will be able to improve the comments.

### DIFF
--- a/debug/README.md
+++ b/debug/README.md
@@ -54,7 +54,7 @@ messages of the lower ones.
        A length limit of 0 means the value of the `short_length` option is
        used.
 
-* 102: Print all disjuncts before pruning.
+* 102: Print all disjuncts before and after pruning.
 
 * 103: Show unsubscripted dictionary words and subscripted ones which share
        the same base word.
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -183,6 +183,18 @@ static inline bool connector_uc_eq(const Connector *c1, const Connector *c2)
 	return (connector_uc_num(c1) == connector_uc_num(c2));
 }
 
+/*
+ * Return the deepest connector in the connector chain starting with \p c.
+ * @param c Any connector
+ * @return The deepest connector (can be modified if needed).
+ */
+static inline Connector *deepest_connector(const Connector *c)
+{
+	for (; c->next != NULL; c = c->next)
+		;
+	return (Connector *)c; /* Note: Constness removed. */
+}
+
 /* Length-limits for how far connectors can reach out. */
 #define UNLIMITED_LEN 255
 
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -114,6 +114,7 @@ struct Connector_struct
 	                         this could ever connect to.  Initialized by
 	                         setup_connectors(). Final value is found in
 	                         the power pruning. */
+	uint8_t prune_pass;   /* Prune pass number (one bit could be enough) */
 	bool multi;           /* TRUE if this is a multi-connector */
 	int tracon_id;        /* Tracon identifier (see disjunct-utils.c) */
 	const condesc_t *desc;
--- a/link-grammar/disjunct-utils.c
+++ b/link-grammar/disjunct-utils.c
@@ -1050,6 +1050,12 @@ Tracon_sharing *pack_sentence_for_parsing(Sentence sent, unsigned int dcnt,
 
 	Tracon_sharing *ts = pack_sentence(sent, dcnt, ccnt, false, keep_disjuncts);
 
+	if (verbosity_level(D_SPEC+2))
+	{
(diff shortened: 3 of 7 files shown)


linas:
Thanks, I merged with only a cursory review, mostly because its late-night here and I'm tired, and I didn't want to hold you up. The more documentation the better.  For example, its not at all clear what the "per-word minimum/maximum link distance" is supposed to be, at least not when reading the file from top to bottom.  I guess its discovered during pruning. I guess the "old comment" describes it. But yes, write down anything obvious esp when you think you might forget it.

I'm pretty sure that, when you get to the end of this, you should write a paper, meant for publication, that describes all of the optimizations you've implemented.  There's been a 10x speedup, and you might be closing on a 20x speedup, as compared to a year or two ago: this is quite unusual.  I think it's worth publishing.

ampli:
> For example, its not at all clear what the "per-word minimum/maximum link distance" is supposed to be

I added the following comments about this table:
``` c
/* Per-word minimum/maximum link descriptor.
 * The dimension of the 2-element arrays below is used as follows:
 * [0] - left side; [1] - right side.
 * nw is the minimum nearest_word of the shallow connectors.
 * fw is the maximum farthest_word of the shallow connectors.
 *
 * The connection of the shallow connector is to a greater distance than
 * the connections from the deepest ones. For word w, words in the ranges
 * (nw[0], w) or (w, nw[1]) cannot connect to words outside its
 * corresponding range without crossing a link to a shallow connector of
 * w. In addition, words in (nw[0], w) cannot connect to words before
 * fw[0] (and similarly for the other range).
 *
 * These values are computed by build_mlink_table() after the first
 * power_prune() call, and before each additional call to power_prune(). */
typedef struct
{
	WordIdx_m nw[2];   /* minimum link distance */
#if FW
	WordIdx_m fw[2];   /* maximum link distance - not implemented yet */
#endif
} mlink_t;
```

Because I still remember the subtle details and this matter is complicated, it may not be obvious for me what needs a further clarification. Just tell me what is still unclear here and in other places and I will try to fix that.

linas:
OK, How can we know when a connector is shallow, and when it is deep?  Is it
```
Disjunct *d;
d->shallow->deeper->deepest;
```
or is it `d->deepest->shallower->shallowest;`?  Maybe the ordering depends on `d->left` vs `d->right` ?  By looking at `deepest_connector` I can infer that the first answer is the correct one, and ordering does not depend on left/right.  Add a comment to explicitly say that...

ampli:
This is explained in an original comment in `prune.c`:
```
    163            (A connectors is deep if it is not the first in its list; it
    164            is shallow if it is the first in its list; it is deepest if it
    165            is the last on its list.)
```
I will add this explanation to `Connector_struct` and append it to this PR:
Something like:
``` c
struct Connector_struct
{
...
   bool shallow;  /* TRUE if this is a shallow connector.
	           * A connectors is shallow if it is the first in its list on its
	           * disjunct. It is deep if it is not the first in its list; it is
	           * deepest if it is the last on its list. */
...
}
```

BTW, I also fixed the comment block that includes this original comment, as it got a deep comment rot. The block before it is even more comment-rotten (fixed too). I will send the PR after this PR is applied to ensure everything is applied cleanly.

linas:
thanks. I just merged #1026.   Can you look at #1024 ? I think it simplifies things, but maybe you can see something I don't.

ampli:
> Can you look at #1024 ? I think it simplifies things, but maybe you can see something I don't.

Yes. I have commented there about a problem. I will also add a comment on an aspect in which your fix is better.

PR #1026, opened 2019-10-17
Add comments to #1025

ampli:
Add comments to the patch of #1025.


ampli:
About this commit:
- prune.c: Add a FIXME for the big comment rot that happened over the years.

These comments don't describe the current code in this file.
The initial ones describe what is done now in the expression pruning, and can be updated and moved to there.
The last section has an idea that I still have to check if it is covered by the current code or by my current prune WIPs.
The rest is not relevant any more and should be removed.
I will send a PR.

I will update these comments.

PR #1027, opened 2019-10-18
Fix prune comments

ampli:
Fix a deep comment rot on the original comments in `prune.c`.
The initial ones now actually describe the expression_prune algo so convert them and move to `exprune.c`.
I also added in `exprune.c` a FIXME that I would like to implement (inspired by the SAT code).
I appended the "shallow, deep, deepest" comment here since PR #1026 has already been applied.

### DIFF
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -129,51 +129,6 @@ struct prune_context_s
 };
 
 /*
-  FIXME: These comments are old and are not relevant in their
-  current form. Convert them to relevant comments.
-
-  The algorithms in this file prune disjuncts from the disjunct list
-  of the sentence that can be eliminated by a simple checks.  The first
-  check works as follows:
-
-  A series of passes are made through the sentence, alternating
-  left-to-right and right-to-left.  Consider the left-to-right pass (the
-  other is symmetric).  A set S of connectors is maintained (initialized
-  to be empty).  Now the disjuncts of the current word are processed.
-  If a given disjunct's left pointing connectors have the property that
-  at least one of them has no connector in S to which it can be matched,
-  then that disjunct is deleted. Now the set S is augmented by the right
-  connectors of the remaining disjuncts of that word.  This completes
-  one word.  The process continues through the words from left to right.
-  Alternate passes are made until no disjunct is deleted.
-
-  It worries me a little that if there are some really huge disjuncts lists,
-  then this process will probably do nothing.  (This fear turns out to be
-  unfounded.)
-
-  Notes:  Power pruning will not work if applied before generating the
-  "and" disjuncts.  This is because certain of it's tricks don't work.
-  Think about this, and finish this note later....
-  Also, currently I use the standard connector match procedure instead
-  of the pruning one, since I know power pruning will not be used before
-  and generation.  Replace this to allow power pruning to work before
-  generating and disjuncts.
-
-  Currently it seems that normal pruning, power pruning, and generation,
-  pruning, and power pruning (after "and" generation) and parsing take
-  about the same amount of time.  This is why doing power pruning before
-  "and" generation might be a very good idea.
-
-  New idea:  Suppose all the disjuncts of a word have a connector of type
-  c pointing to the right.  And further, suppose that there is exactly one
-  word to its right containing that type of connector pointing to the left.
-  Then all the other disjuncts on the latter word can be deleted.
-  (This situation is created by the processing of "either...or", and by
-  the extra disjuncts added to a "," neighboring a conjunction.)
-
-*/
-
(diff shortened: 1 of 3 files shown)


link-grammar/parse/prune.c
```diff
@@ -129,99 +129,59 @@ struct prune_context_s
...
+*
+* The kinds of constraints it checks for are the following:
+*
+*  1) successive connectors on the same disjunct have to go to
+*     nearer and nearer words.
+*
+*  2) two deep connectors cannot attach to each other
+*     (A connectors is deep if it is not the first in its list; it
+*     is shallow if it is the first in its list; it is deepest if it
+*     is the last on its list.)
```
linas: This claim: "two deep connectors cannot attach to each other" seems to be immediately contradicted below: "on two adjacent words, a pair of connectors can be used only if they're the deepest" ... !?  I'll merge this pull req, but this neds fixing.

link-grammar/parse/prune.c
```diff
@@ -129,99 +129,59 @@ struct prune_context_s
...
+*
+*  3) on two adjacent words, a pair of connectors can be used
+*     only if they're the deepest ones on their disjuncts
+*
+*  4) on two non-adjacent words, a pair of connectors can be used only
+*     if not [both of them are the deepest].
```
linas: "only if both of them are not the deepest"

link-grammar/parse/prune.c
ampli: It is not a contradiction, because the deepest connector can be a shallow one, while a deep connector cannot be shallow.

Maybe a clarification can be added:
```
    144 *  3) on two adjacent words, a pair of connectors can be used
    145 *     only if they're the deepest ones on their disjuncts.
    146 *     NOTE: the deepest connector can be shallow, while a
    147 *     deep connector cannot be shallow.

File: link-grammar/parse/prune.c (hunk 2 above)
Comment (ampli) [reply link-grammar/parse/prune.c]: Your wording has a different interpretation.

The clam in (4) is that at least one of the connectors should not be the deepest. This means that one of them **can** be the deepest.
On the other hand, your wording means that even not one of them can be the deepest.

Comment (linas) [reply link-grammar/parse/prune.c]: OK, then, "only if at least one of them is not the deepest."

**Comment by ampli** [comment]
(4) has a slight comment rot which is not related to your post above.
I didn't check the history of this comment but it looks as if it originates form the times before the idea of parsing with nulls. It is correct only when the null count must be zero.

File: link-grammar/parse/prune.c (hunk 1 above)
Comment (linas) [reply link-grammar/parse/prune.c]: I find the clarification even more confusing. So, in 2), I guess it is saying that every deep connector must connect to a shallow one? So that:
```
2) two deep connectors cannot attach to each other; every deep connector must 
    connect to a shallow one.
```

Comment (linas) [reply link-grammar/parse/prune.c]: I tweaked wording above.

Comment (ampli) [reply link-grammar/parse/prune.c]: The implication from "two deep connectors cannot attach to each other" as a principal exclusion rule is slightly wider:
1. A deep connector can only attache to a shallow one. (As you wrote.)
2. A shallow connector can attach to any connector.

So maybe we can change it to:
```
2) two deep connectors cannot attach to each other; i.e. a deep
connector can only attache to a shallow one, and a shallow 
connector can attache to any connector.
```

link-grammar/parse/prune.c
ampli: > OK, then, "only if at least one of them is not the deepest."
 This seems to me fine.
Considering my later comment below, I propose:
 "only if at least one of them is not the deepest (supposing null links are not allowed)."
￼

link-grammar/parse/prune.c
linas: that works

link-grammar/parse/prune.c
linas: "only if at least one of them is not the deepest (ignoring null-linked words)."
or better yet:

"only if at least one of them is not the deepest (ignoring null-linked words, which behave as if they are entirely absent)."

ampli: "only if at least one of them is not the deepest (ignoring null-linked words, which behave as if they are entirely absent)."

I'm still struggling to understand the exact meaning of the parenthesized part. This may hint that maybe it is not clear...

Maybe this is clearer:
"on non-adjacent words with no intervening null-linked words, a pair of connectors can be used only if at least one of them is not the deepest"

Or this (the same with added parenthesizes):
"on non-adjacent words (with no intervening null-linked words), a pair of connectors can be used only if at least one of them is not the deepest"

linas: Yes, you're right the last one is the best.

PR #1029, opened 2019-10-19
Various batch fixes

ampli:
Two interesting ones:
- Remove `verbsoity=1` so a different verbosity can be used on the command line.
- Add initial blank in `* This is an item`.


linas:
Thanks!

PR #1031, opened 2019-10-21
Forgotten cleanup commits

ampli:
- Fix format signed/unsigned inconsistency
It looks I mad it after the initial PR push and forgot to force-push it.

- Add `pruning_pass_end()` for common pruning pass ending code
The omission of this one is a mystery - I intended to include it in a cleanup PR, but it got disappeared. Included here is a rewrite...

### DIFF
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -904,6 +904,27 @@ static bool check_null_word(prune_context *pc, int w)
 	return false;
 }
 
+/**
+ * Return \c true if a pass has not made any changes.
+ * Also print pruning pass stats.
+ */
+static bool pruning_pass_end(prune_context *pc, const char *pass_dir,
+                             int *prune_total)
+{
+	int total = pc->N_deleted[0] + pc->N_deleted[1];
+
+	lgdebug(D_PRUNE, "Debug: %s pass changed %d and deleted %d (%d+%d)\n",
+	        pass_dir, pc->N_changed, total, pc->N_deleted[0], pc->N_deleted[1]);
+	if (pc->ml != NULL)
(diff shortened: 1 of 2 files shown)


link-grammar/parse/prune.c
```diff
@@ -904,6 +904,25 @@ static bool check_null_word(prune_context *pc, int w)
...
+
+	lgdebug(D_PRUNE, "Debug: r->l pass changed %d and deleted %d (%d+%d)\n",
+	        pc->N_changed, total, pc->N_deleted[0], pc->N_deleted[1]);
+	if (pc->ml != NULL)
+		lgdebug(D_PRUNE, "Debug: xlink: %d\n", pc->N_xlink);
+
+	pc->N_changed = pc->N_deleted[0] = pc->N_deleted[1] = pc->N_xlink = 0;
+
+	*prune_total += total;
+	return ((pc->N_changed == 0) && (total == 0));
```
linas: you  set `pc->N_changed` to zero up above.  The old code did this test before setting it to zero.

link-grammar/parse/prune.c
```diff
@@ -904,6 +904,25 @@ static bool check_null_word(prune_context *pc, int w)
 	return false;
 }
 
+/**
+ * Return \c true if a pass has not made any changes.
```
linas: stray backslash-c

ampli: This is by intention, and I already used it several times.
It is a Doxygen escape for 'code'.
I also used several times `\p` (for parameter).
(These were the only Doxygen escapes I used by now.)

ampli:
I have just force-pushed a fix.

linas:
ok

PR #1033, opened 2019-10-25
Extended power prune: Disable code which is ahead of its time

ampli:


linas:
whoops, I did the commits out of order, and now there is a conflict ...

ampli:
I have just force-pushed here the resolution.

linas:
OK, thanks, sorry.

PR #1036, opened 2019-10-27
pp_prune(): Add handling of selectors with '*'

ampli:
This is something that the original authors thought it is impossible.
<details><summary>Original comment</summary>
Consider the foo part.  Consider a connector C.  When does
   foo match C for our purposes?  It matches it if every possible link
   name L (that can result from C being at one end of that link) results
   in post_process_match(foo,L) being true.  Suppose foo contains a "*".
   Then there is no C that has this property.  This is because the *s in
   C may be replaced by some other subscripts in the construction of L.
   And the non-* chars in L will not post_process_match the * in foo.
</details>

It is indeed impossible to do in all cases, but I found that in particular cases that happen frequently enough it is still possible. This is because these "some other subscript" are frequently not  found in sentences, especially short ones.

The result is that with this PR, many sentences now have significantly less linkages, but of course they have still always exactly the same number of linkages after postprocessing. For example, in the `fixes` batch this effects about 375 sentences.

Here is an example:
`linkparser> It is deadlocking right here.`

Original:
```
Found 104 linkages (40 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS=-0.61 LEN=7)
```
After this PR:
```
Found 52 linkages (40 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS=-0.61 LEN=7)
```

Since this effects mostly short sentences, and the functions in `count.c` + `fast-match.c` now take about 4% CPU for the `fixes` batch run, there is no significant CPU saving from reducing the parsing overhead  (for short sentences) by better pruning, so there is no significant speedup in the batch benchmarks.

### DIFF
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -1374,6 +1374,51 @@ static bool mark_bad_connectors(multiset_table *cmt, Connector *c)
 	return false;
 }
 
+/**
+ * Validate that no connector in the sentence that matches the
+ * candidate trigger connector \p t may create a link x so that
+ * post_process_match(s, x)==false.
+ *
+ * If so, this ensures that the candidate trigger connector may form
+ * only links which satisfy the selector. (This bypasses the limitation
+ * on selectors with '*' that is mentioned in the original comments).
+ *
+ * The algo validates that the '*'s in the selector \p s and in an each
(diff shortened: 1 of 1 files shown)


ampli:
Force-pushed a fix of using `easy_match_desc()` instead of `easy_match()` (in the new code).

PR #1038, opened 2019-10-28
Imprve the efficiency of intersect_strings()

ampli:
Speed up `intersect_strings()` in a factor of about 3, by using connector encoding.

The main speedup is because in absolutely most of the cases  a pointer into the connector strings is returned without entering the label constructing loop, and by avoinf `strlen()`.

Other changes:
- Move MAX_LINK_NAME_LENGTH to `connectors.h`.
- Use a dynamic buffer length for the result.

`MAX_LINK_NAME_LENGTH` is now only used  in `lisjuncts.h`, in which a buffer overflow cannot happen (a truncation would happen instead, but this is not a likely cases due to the big buffer).

The benchmark of `fix-long` and `failures` shows ~2% speedup. It seems it depends on the number of linkages + the linkage limit.

On the `pandp-union.txt` there is no speedup, although the function benchmark still shows ~3x speedup for `intersect_strings()`. It seems this is due to the low `linkage_limit` which is used there (1000) and the mix of sentence length which include a lot of short ones. When I increased it to 10000, I got a speedup of ~2% too.

BTW, my motivation of improving  `intersect_strings()` is to use it for an improved pruning idea in pp_prune() without introducing too much overhead (no results yet).

### DIFF
--- a/link-grammar/connectors.c
+++ b/link-grammar/connectors.c
@@ -244,8 +244,11 @@ static bool connector_encode_lc(const char *lc_string, condesc_t *desc)
 
 	for (s = lc_string; '\0' != *s; s++)
 	{
-		lc_value |= (lc_enc_t)(*s & LC_MASK) << ((s-lc_string)*LC_BITS);
-		if (*s != WILD_TYPE) lc_mask |= wildcard;
+		if (*s != WILD_TYPE)
+		{
+			lc_value |= (lc_enc_t)(*s & LC_MASK) << ((s-lc_string)*LC_BITS);
+			lc_mask |= wildcard;
+		}
 		wildcard <<= LC_BITS;
 	};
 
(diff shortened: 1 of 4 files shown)


ampli:
In the new `intersect_strings()`, no buffer overflow in the result is ensured by:
```
	/* Note: LC_BITS * MAX_CONNECTOR_LC_LENGTH < sizeof(lc_enc_t).
	 * So after MAX_CONNECTOR_LC_LENGTH shifts lc_label must be 0. */
```

I intend to start to add static assertions for such things.

ampli:
I used here `alloca()` for the result, to automatically accommodate for any connector string length.
However, this function is called in a loop. GCC  would either not inline it, or (most probably) inline it but not free the so allocated memory until the end of the caller function, which contribute to max. stack size usage. Both of these things are not so good.

(There may be more cases like that, and it may be a good idea to fix them too.)

In the case of this function I propose the following instead of using `alloca()`:
1. In `connector.h', change `MAX_LINK_NAME_LENGTH` to a higher value, say 20. This will support long "connector" names in the affix file (if not exempted, see (2)).
2. In `read-dict.c`, check for connector name length violation. The affix file may be exempted from this check if desired (can be detected by `dict->lang==NULL`).

PR #1043, opened 2019-11-07
Prune length limit

ampli:
- Trim `farthest_word` in the power-pruning.
- Add new type of pruning: "must have" links.
- Simplify/improve readability/clarify code.
- Rename `mlink_t` and `cross_mandatory_link_prune()`.
- Add additional "mlink" pruning step.

The speedup is minimal: 1-3% (depending on sentence length).
More work needed on trimming  `farthest_word`. However, it is not clear if it can be improved. This is due to the memory sharing of connectors that is done according to tracons, while the minimal length_limit is according to "leacons".

### DIFF
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -1878,7 +1878,9 @@ static unsigned int cross_mandatory_link_prune(Sentence sent, mlink_t *ml)
 				}
 #endif /* FW */
 
-				shallow_c->farthest_word = MIN(ml[w].nw[1], shallow_c->farthest_word);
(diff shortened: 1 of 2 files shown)


link-grammar/parse/prune.c
```diff
@@ -62,10 +69,8 @@ typedef uint8_t WordIdx_m;     /* Storage representation of word index */
 typedef struct
 {
 	WordIdx_m nw[2];   /* minimum link distance */
-#if FW
 	WordIdx_m fw[2];   /* maximum link distance - not implemented yet */
```
linas: The comment is no longer accurate - `not implemented yet.`

PR #1047, opened 2019-11-08
Improved PP pruning - part 2

ampli:
The current (original)  check for criterion links is "permissive" - if connectors that **maybe** can create criterion links are found this is considered a rule satisfaction.

From the original comment:
``` text
   For the criterion parts, we need to determine if there is a
   collection of connectors C1, C2,... such that by combining them you
   can get a link name L that post_process_matches bar or baz.  Here's a
   way to do this.  Say bar="Xabc".  Then we see if there are connector
   names that post_process_match "Xa##", "X#b#", and "X##c".  They must
   all be there in order for it to be possible to create a link name
   "Xabc".
```

So only the the necessary condition for criterion links is checked. What is not checked is the sufficient condition - that at least one link they may create actually PP-matches the criterion rule.

**Add this check.**

This causes a better pruning that significantly reduces the number of linkages of many sentences (before preprocessing - after preprocessing the number of linkages is of course the same).
Regretfully, due to the current pp-pruning algo, this mostly doesn't affect long sentences.

[en-basic-diff.txt](https://github.com/opencog/link-grammar/files/3825709/en-basic-diff.txt)
[en-fixes-diff.txt](https://github.com/opencog/link-grammar/files/3825710/en-fixes-diff.txt)

The batch benchmark speedup (`fix-long`, `failures`) is about 1-2% (less on `pandp-union.txt`).

### DIFF
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -1286,6 +1286,22 @@ static bool can_form_link(const char *s, const char *t, const char *e)
 	return true;
 }
 
+#ifdef DEBUG_PP_PRUNE
+static const char *connector_signs(Cms *cms)
+{
+	static char buf[3];
+	size_t i = 0;
+
+	if (cms->left) buf[i++] = '-';
+	if (cms->right) buf[i++] = '+';
+	buf[i] = '\0';
+
+	return buf;
+}
+#else
+#define connector_signs(x) NULL /* For YCM IDE */
+#endif /* DEBUG_PP_PRUNE */
+
 /**
  * Returns TRUE iff there is a connector name in the sentence
  * that can create a link x such that post_process_match(pp_link, x) is TRUE.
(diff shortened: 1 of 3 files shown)


link-grammar/parse/prune.c
```diff
@@ -1283,18 +1313,22 @@ static bool match_in_cms_table(multiset_table *cmt, const char *pp_link,
...
+			ppdebug("MATCHED %s%s\n", connector_string(cms->c),connector_signs(cms));
+			cms->last_criterion = true;
+			found = true;
+			continue;
 		}
-		ppdebug("NOT-MATCHED %s \n", connector_string(cms->c));
+		ppdebug("NOT-MATCHED %s%s\n", connector_string(cms->c),connector_signs(cms));
 	}
 
+	if (found) return true;
```
linas: `return found;` is simpler.

PR #1048, opened 2019-11-08
Cleanup for PR #1047

ampli:
- match_in_cms_table(): Simplify function return
Per review comment at PR #1047.


linas:
thanks.  Would it have been better to not merge the earlier pull req, and wait?

ampli:
I think it is better to merge unless you suspect there is some problem.

PR #1059, opened 2019-12-12
Clang static-analyzer 2

ampli:
- Fix some FP messages.
There are many other such messages, that eliminating them needs a significant number of assert() calls. Maybe the solution is to make them ifdef'ed on using a static analyzer.

- Fix non-fatal bugs.
The most interesting one is in `free_Exp()` (using stale memory). First I didn't understand how come it hasn't been detected by ASAN. Then I realized this function is not being used any more because expressions are now allocated from memory pools. For now I ifdef'ed it out.


linas:
Huh. interesting

PR #1061, opened 2019-12-16
Remove Python2 support + configuration cleanup and bug fixes

ampli:
Removing Python2 support:
- Configuration and build.
- Bindings.
- Code support  (no point to pretend supporting Python2 in the code when this is not tested anymore).
- Documents.

Configuration bug fixes, enhancements and cleanup:
- gcc warning flags didn't actually get checked - all get accepted.
- gcc C-only flag was mentioned as possible C++ flag.
- Fix help message for `--enable-sat-solver`.
- Remove old stuff from configuring Java.
- Provide a clear error if autoconf-archive is not installed.
However, the error says to install the `autoconf-archive` package, while it is named differently 
on some distributions.
- Add -Wformat-signedness for %u/%d consistency.
- autogen.sh: Fix whitespace.

The ChangeLog update here conflicts with the `dialect` PR (#1060).
I can force-push a fix there after you apply this PR.


linas:
wow.

PR #1062, opened 2019-12-17
intersect_strings(): Use a fixed array instead of alloca()

ampli:
alloca() interferes with the inlining of this function and a reduced performance results.
With this fix, the short-sentence batches runs 2-3% faster.
(The speedup for very long sentences is negligible because their parse time dominants.)


linas:
thanks

PR #1064, opened 2019-12-17
Poetic contractions, elisions, colloquialisms

linas:
All of these are tagged with the 'colloquial" tag. 

This seems to damage performance a bit, not sure why.

### DIFF
--- a/data/en/4.0.affix
+++ b/data/en/4.0.affix
@@ -45,7 +45,9 @@ _ ‐ ‑ ‒ – — ― ～ – ━ ー -- - ‧
 % "real" English prefix: y' w/
 % Y'gotta Y'gonna
 % coffee w/milk
-y' w/: PRE+;
+% i'faith
+% th'door
+y' w/ i' th': PRE+;
 
 % The below is a quoted list, used during tokenization. Do NOT put
 % spaces in between the various quotation marks!!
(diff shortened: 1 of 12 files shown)


data/en/4.0.affix
```diff
@@ -39,13 +39,17 @@ _ ‐ ‑ ‒ – — ― ～ – ━ ー -- - ‧
 -- ‒ – — ― "(" ")" "[" "]" ... ";" ±: MPUNC+;
 
 % Suffixes
-'s 're 've 'd 'll 'm ’s ’re ’ve ’d ’ll ’m: SUF+;
+'s 're 've 'em 'd 'll 'm 'd've 'r 't
+’s ’re ’ve ’em ’d ’ll ’m ’d’ve ’r ’t: SUF+;
```
ampli: Duplication here.

data/en/4.0.affix
```diff
@@ -39,13 +39,17 @@ _ ‐ ‑ ‒ – — ― ～ – ━ ー -- - ‧
...
 
 % Prefixes
 % "real" English prefix: y' w/
 % Y'gotta Y'gonna
 % coffee w/milk
-y' w/: PRE+;
+% i'faith
+% th'door
+i' o' t' th' wi' y' w/
+i’ o’ t’ th’ wi’ y’: PRE+;
```
ampli: Also here.

data/en/4.0.affix
linas: they're not the same mark; one is vertical, one is slanted. ox27 and 0xe28099=U+2019

it would be nice if either could be used, automatically...

PR #1065, opened 2019-12-17
Dialect + general bug fixes / cleanup

ampli:
Bug fixes:
- free_cost_table(): Nullify cost_table to designate no caching
It causes a double-free crash.
- dictionary_create_default_lang(): Don't open "en" a second time
It causes double error messages if the 'en' dict contains an error.

- get_label(): Fix a bad and truncated error message if a name is missing
- require_delimiter(): Don't print the whole file after the point of error

Python / tests suite:
- Python binding ParseOptions(): Add parse_options_*_dialect()
- Python binding ParseOptions(): Notify real variable names on errors
- tests.py: Add a dialect test

Cleanup:
- Define DIALECT_OPTION: The name of the dialect option in error messages
- Dictionary_s: Remove unused field "Connector_set"
- !help dialect: Document a current limitation in user variable values
- DEBUG mode: Fix format warnings due to -Wformat-signedness
- Dialect code: Add '.' at end of error messages
I now think that messages must end with a period. I started with the recent ones.

### DIFF
--- a/bindings/python-examples/parses-dialect-en.txt
+++ b/bindings/python-examples/parses-dialect-en.txt
@@ -1,0 +1,10 @@
+% Dialect tests
+
+-dialect='headline'
+IThieves rob bank
+O
+O    +------->WV------>+
+O    +--->Wd---+---Sp--+--Os--+
+O    |         |       |      |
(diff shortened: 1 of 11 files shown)


linas:
thanks

PR #1068, opened 2019-12-17
display_word_split(): Avoid the need to free the cost table

ampli:
I again got a `free()` crash (this time on a "bad pointer").
Copying parse options cause subtle memory management with dialect_info.
So avoid copying `opts`.

### DIFF
--- a/link-grammar/dict-common/print-dict.c
+++ b/link-grammar/dict-common/print-dict.c
@@ -234,7 +234,6 @@ static char *display_word_split(Dictionary dict,
                                char * (*display)(Dictionary, const char *))
 {
 	Sentence sent;
-	struct Parse_Options_s display_word_opts = *opts;
 
 	if ('\0' == *word) return NULL; /* avoid trying null strings */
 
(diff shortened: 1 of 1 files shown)


link-grammar/dict-common/print-dict.c
```diff
@@ -247,17 +246,18 @@ static char *display_word_split(Dictionary dict,
...
-	if (0 == sentence_split(sent, &display_word_opts))
+	if (0 == sentence_split(sent, opts))
 	{
 		/* List the splits */
 		print_sentence_word_alternatives(s, sent, false, NULL, NULL);
 		/* List the disjuncts information. */
 		print_sentence_word_alternatives(s, sent, false, display, NULL);
 	}
 	sentence_delete(sent);
-	free_cost_table(&display_word_opts); /* XXX */
```
linas: It would have been enough to remove this one line.  The struct copy does not copy things that get pointed at.

ampli:
The problem is that `setup_dialect()` that is called in `sentence_split()` may allocate `cost_table` inside the Parse_Options object (in this case inside `display_word_opts`).  If nothing frees it this causes a memory leak. It may also not allocate this memory in case it was already allocated (cached `cost_info`). In that case it may be freed in the original `opts`. It is hard to solve this in a clean way and the best is to avoid the Parse_Options duplication as I did here.

linas:
oh ok

PR #1070, opened 2019-12-24
Cleanup: Unify Exp functions + more

ampli:
The new Exp tag field and also an additional WIP field there remained uninitialized in ZZZ nodes added by `add_empty_word()`.

Fix that by using there `Exp_create()`.
In the same occasion:
- Unify the sentence Exp function with the dict ones.
- Replace the SAT parser private Exp functions by these unified functions.

### DIFF
--- a/bindings/python-examples/parses-pos-en.txt
+++ b/bindings/python-examples/parses-pos-en.txt
@@ -51,6 +51,6 @@ P
 % the word position of these combined splits correctly.
 
 IAs no linkage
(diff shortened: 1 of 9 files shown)


ampli:
Added:
- Fix test sentence (test failure due to a recent dict change).
- Fix sentence-check.py.
  Ironically, adding "Note: This only runs with Python3!" caused it not to run with Python3 because this addition creates a non-comment (the big docstring) before a `__future__ import` statement.

PR #1071, opened 2019-12-27
Dict: Flag an empty word list in as an error

ampli:
Encountered in a WIP.
There is currently no use of an empty word list so it indicates an error.
So instead of special-case it I flagged it as an error.


linas:
Your sort order is different than mine ..!?

linas:
ugh. There doesn't seem to be a consistent sort order

ampli:
A consistent sort order can be implemented there, but I even didn't add that to my TODO list...

PR #1076, opened 2020-01-04
print.c cleanup

ampli:
Initial general changes from the fake-xlink branch.
Due to using `qsort()` instead of a per-word link_array loop, massive diagram printouts became faster by a few percents. A modified sorting is used later (WIP) for printing cross links.

### DIFF
--- a/link-grammar/print/print.c
+++ b/link-grammar/print/print.c
@@ -298,7 +298,7 @@ build_linkage_postscript_string(const Linkage linkage,
 {
 	int link, i,j;
 	int d;
-	bool print_word_0, print_word_N;
+	bool print_word_0 = true, print_word_N = true;
 	int N_links = linkage->num_links;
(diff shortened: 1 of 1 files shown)


ampli:
It may be a time to fix the postscript code too, since the fake cross-links will confuse it farther.
For now it cannot print sentences longer than page width, or with words that get wrapped over lines, and also it cannot handle non-ASCII words.
Note that the Singularity's LL uses the PS output (the part that functions fine for now) for result analysis, so maybe we should provide a flag to disable the cross-link display output (in case they use it  with the `en` dict for comparisons). Such a flag may be needed anyway for compatibility, but I'm not sure how to implement it (special function call?). Maybe the `bool print_ps_header` parameter can be extended to an `int` flag field in a compatible way (that will have a `DISPLAY_CROSS_LINKS` bit) so existing library users will get the traditional result.

PR #1078, opened 2020-01-05
Remove the experimental regex-tokenizer code

ampli:
Remove the demo of tokenizing by a regex.
(I have ideas for replacing the current tokenizer but it may take some time to implement that.)


linas:
Changelog should probably make note of this.

ampli:
> Changelog should probably make note of this.

I'm not sure, as the changelong has never been updated about the regex tokenizer demo.

linas:
oh ok.

PR #1079, opened 2020-01-08
Dialect bug fix and improvements

ampli:
- Dialect implementation bug fix: realloc() may change address.
This is a design bug which may cause a wild pointer reference crash if there are > 100 different dialect components.
- Ensure that Parse_Options' dialect info refers to the sentence's dict.
Indicate an error in case the user apples a parse-option with dialect that has previously applied to another dict. (Maybe just invalidating the dialect cache could be better in such a case, I'm not sure and this may not bee too important).
- Add a test suite check that the initial dialect definitions get applied.
- exptag_add(): Rename to exptag_dialect_add()
This is a preparation to adding another tag type.

An additional known bug remains: `!!word` doesn't apply the dialect costs as was initially intended.
This is fixed in the next PR (that implements `!!word/regex/`).


linas:
thanks. Sorry for very long delay

PR #1081, opened 2020-01-17
Additional pruning in is_cross_mlink()

ampli:
This patch adds a significant pruning for null_count=0 but only a little pruning for null_count>0.
The CPU saving for long sentences is relatively low (1-2% maybe).


ampli:
Somehow old commits have been pushed here, and I cannot delete them by force-pushing again.
Anyway, I improved this patch, so I closed this one and will create a new one.

PR #1082, opened 2020-01-17
Additional pruning in is_cross_mlink()

ampli:
This patch adds a significant pruning for any null count.
It implements a combination of new ideas:
- Using minimum nearest_word without taking into account unidirectional disjuncts.
- Using minimum nearest_word of only unidirectional disjuncts.

And most important:
- Using `*_table_search()` for a check of `possible_connection()`.
`*_table_search()` calls `possible_connection()` again, but the recursion stops in that because it returns near its start. I intend to test a full recursive version of `power_prune()`.

For short sentences, there is a drastic reduction in the parsing time, but since it is only about 2.5% for the `basic` and `fixes` batches, the result is only a slight speedup for them.
For long sentences there is 6-8% speedup (`failures` and `pandp-union` batches, the last one with `!short=20` to make it comparable to the other batches).

### DIFF
--- a/link-grammar/parse/prune.c
+++ b/link-grammar/parse/prune.c
@@ -52,6 +52,7 @@ typedef uint8_t WordIdx_m;     /* Storage representation of word index */
 /* Per-word minimum/maximum link distance descriptor.
  * The dimension of the 2-element arrays below is used as follows:
  * [0] - left side; [1] - right side.
+ *
  * nw is the minimum nearest_word of the shallow connectors.
  * fw is the maximum farthest_word of the shallow connectors.
  *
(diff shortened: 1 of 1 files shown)


ampli:
Again the first two commits repeat previous ones. I will check what can be done about that.

ampli:
Something was severely confused in my `master` branch, which I'm **only** updating using `git pull`, as it was different in the first commit here from `upstream/master` (the opencog/link-grammar master)!

I have no idea how such a thing may happen. It could be that something in my forced pushes of the PRs that you lastly updated confused GitHub (i.e. there is a big there).

I have refreshed my `upstream/master` copy with `git pull -f upstream master:master` and forced-push this PR again. It may be applied now.

PR #1084, opened 2020-01-21
is_expression_optional(): Fix comparison to Exptag_none

ampli:
The infamous bug of assignment instead of comparison has hit me here (preventing nice display of optional connectors).
Many times I use `constant == variable`, especially if the old code already uses that.
But since I somewhat dislike this, I sometimes use in new code `variable == constant`.
The compiler can warn on an assignment in something like `if (variable = constant)`, but not when it has extra parens for "clarity" like in this case.


linas:
But this is exactly why `constant == variable` is so nice.  When your finger slips, and you have a single-byte-error in your source, then the compiler complains.    (I learned this trick decades ago, when there was a contractor who made this error, and it took him 3 weeks to find it. Of course, it didn't help that he would go out and smoke pot for lunch.)

PR #1086, opened 2020-01-25
Even more !!word/ functionality

ampli:
Mainly per discussion at PR #1085.
Main changes:
- Replace <--> by <>.
- Add flag 'a' for connector matching in any order.
- Add 'f' for full disjunct specification string (as copied fro !disjuncts).
The flag is needed to prevent a lot of output from partial matches.
- Add "r" flag to request a regex match. This is not normally needed as the code added here automatically detects regular expression and automatically quote `*` and `+` when a regular expression is not intended. So no need to use`\Q` any more.

More additions:
- Add flag summary on unknown flag.
- Fix the error message when the word argument is missing (the current error message on `!!/` is misleading).
- Prevent an empty line before first connector macro (for consistent spacing).
- Update the help text.

Edit:
Pushed more commits:
1. Fix false positive of Clang static analyzer.
2. Fix ignored flag warning.
3. Fix a NULL dereference on normal disjunct printing (e.g. using `-v=102`).

Edit:
Forced push to fix 3 bugs.

### DIFF
--- a/data/command-help-en.txt
+++ b/data/command-help-en.txt
@@ -338,13 +338,26 @@ The above command is more useful for a single disjunct (1234 is an example
 for a disjunct number, see below for disjunct print format):
   !!test.n/1234/m
 
+
+Show selected disjuncts according to the supplied string (* and +
+are automatically escaped if no other regex meta characters in the string):
+  !!test.n/Ds**x+/
+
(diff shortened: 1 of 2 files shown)


linas:
thanks!

PR #1087, opened 2020-02-01
A simple change with a big pruning

ampli:
EDIT:
Forced-push to simplify the check and improve the commit description.

- possible_connection(): Check multi connectors nearest_word.
 The original condition assumes a multi connector may provide connectivity to the checked word range.  Actually validate that using its nearest_word.
  This change adds a significant pruning for any null count.
  The speedup for the long sentences batch benchmarks (`failures` and ` pandp-union`) is about 5%.
- build_mlink_table(): Don't calculate final nw when unneeded.
  The speedup here is very small.

Edit:
Forced-push to update the block comment.


linas:
thanks!

Issue #1088, opened 2020-02-16
unit test `dict-reopen` is failing


linas:
Running the unit test prints:
```
Opening the dictionary for the 4'th time
link-grammar: Info: Dictionary found at ../data/en/4.0.dict
link-grammar: Error: Dialect setup belongs to a different dictionary.
Fatal error: 
Assertion (NULL==sent->last_word) failed at ../../link-grammar/tokenize/tokenize.c:2806: wordgraph_create(): wordgraph exists
Segmentation fault
```

ampli:
I have just submitted PR #1089 to address this crash.
However, an additional fix may be needed.

What happens is that the `Parse_Options` that was used to split a sentence with one dictionary, is then used to split a sentence with another dictionary. Since the dialect setting is cached in this object, it may be inappropriate for another dictionary.

The code now just notifies that without farther handling, and this is wrong (since then incorrect cached info is used). The fix here reset the said cache to enable it to be redone with the new dictionary.

But a problem remains: How to identify a new dictionary.
Possibilities:
1. Dictionary_s address.
This is what is done now. The problem with that is that a new dictionary may get the same address as the old one.
2. The proposed fix in the XXX comment - use a unique dictionary ordinal number.
The problem is that it needs to be done is a thread safe way, i.e. with tread locking, a thing that we try to avoid in the library.
3. Use a time stamp. However, it is problematic to ensure different time stamps for "concurrent" dictionary opens using portable code.
4. Use a dictionary ordinal number along with the stack frame address at the time of grabbing it.
It seems this combination must be unique.

I propose to implement possibility 4.

linas:
Option 5: store the language in `Parse_Options` and compare it to the language in `Dictionary_s` -- if the two do not match, then there's a hard error. 

(I was thinking of creating a new release; are there pending fixes you want to get in before then?)

ampli:
> Option 5: store the language in `Parse_Options` and compare it to the language in `Dictionary_s` -- if the two do not match, then there's a hard error.

This will not detect a the problem when the dicts happen to be named identically, e.g. if the user opens the system `en` dict and also a private `en` dict, or reopens the same dict after updating it. Since a crash may result, I think a full-proof method is needed. Option 4 just consists of incrementing a static variable and storing it along with a stack-related address. It is a few-lines solution so I think it is better.

linas:
I don't quite understand the problem then.  Is there some cached pointer to the dict that becomes invalid when the dictionary is deleted?

[...]

PR #1089, opened 2020-02-17
Reset dialect cache belonging to a different dictionary

ampli:
For the problem report see issue #1088.
A unique open-dictionary ID is still needed to be totally safe.


linas:
Thanks!

PR #1091, opened 2020-02-20
Dictionary_s: Add dialect_info for additional dialect cache verification

ampli:
Use dialect_info as an additional verification of the relevancy of the `Parse_options` Dialect cache.

With these two verifications it seems very unlikely that an irrelevant cache will remain undetected (for now I don't have an example for that).

In case a fool-proof solution will be needed, I documented my proposal in a comment.

To verify the using of the dialect cache in normal cases:
`link-parser -v=9 -debug=dialect.c`
Enter two sentences one after the other. For the first one the dialect cache will be created. The second one will just use it.
Then set a new dialect: `!dialect=headline`
Now enter something like `this is test`. The dialect cache will be recalculated due to the new dialect.
A more complex verification requires an added test in tests.py that maybe captures debug messages.

### DIFF
--- a/link-grammar/dict-common/dialect.c
+++ b/link-grammar/dict-common/dialect.c
@@ -236,11 +236,14 @@ bool setup_dialect(Dictionary dict, Parse_Options opts)
 	if (dinfo->cost_table != NULL)
 	{
 		/* Cached table. */
-		if (dinfo->dict != dict)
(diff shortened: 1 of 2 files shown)


linas:
thanks!

linas:
I'll publish a new release "any day now".

Issue #804, opened 2018-09-12
Incorrect parsing due to locale settings


noskill:
Following program return different result depending on locale settings:

```
#include <locale.h>
#include <stdio.h>
#include "link-includes.h"


int main()
{
    Dictionary    dict;
    Parse_Options opts;
    Sentence      sent;
    Linkage       linkage;
    char *        diagram;
    int           i, num_linkages;
    const char * sentence = "Is the bed white?";
    setlocale(LC_ALL, "");
    opts = parse_options_create();
    dict = dictionary_create_default_lang();
    if (!dict) {
        printf ("Fatal error: Unable to open the dictionary\n");
        return 1;
    }

    prt_error("Info: Dictionary version %s, locale %s\n",
            linkgrammar_get_dict_version(dict),
            linkgrammar_get_dict_locale(dict));
    prt_error("Info: Library version %s. Enter \"!help\" for help.\n",
            linkgrammar_get_version());

 
    sent = sentence_create(sentence, dict);

                        parse_options_set_min_null_count(opts, 0);
                        parse_options_set_max_null_count(opts, 0);
                        parse_options_reset_resources(opts);

    
    num_linkages = sentence_parse(sent, opts);
    printf("Num linkages %i found\n", num_linkages);
    for(int i=0; i<num_linkages; i++)
    {
        linkage = linkage_create(i, sent, opts);
        printf("%s\n", diagram = linkage_print_diagram(linkage, true, 800));
        fprintf(stdout, "cost vector = (UNUSED=%d DIS=%5.2f LEN=%d)\n",
                                linkage_unused_word_cost(linkage),
                                linkage_disjunct_cost(linkage),
                                linkage_link_cost(linkage));

        linkage_free_diagram(diagram);
        linkage_delete(linkage);
    }
    sentence_delete(sent);

    dictionary_delete(dict);
    parse_options_delete(opts);
    return 0;
}
```
correct parse:

```
LC_NUMERIC=C ./a.out | head
link-grammar: Info: Dictionary found at ./data/en/4.0.dict
link-grammar: Info: Dictionary version 5.5.2, locale en_US.UTF-8
link-grammar: Info: Library version link-grammar-5.5.1. Enter "!help" for help.
link-grammar: Info: Freeing dictionary en/4.0.dict
link-grammar: Info: Freeing dictionary en/4.0.affix
Num linkages 12 found

    +--------------Xp-------------+
    |       +-------Pa-------+    |
[...]

PR #1093, opened 2020-04-17
Fix SunOS build

dhgutteridge:
SunOS otherwise satisfies the HAVE_LOCALE_T check, but does not provide strtod_l(3), so this needs to fall back to strtod(3) instead. (At any rate, this is true for recent releases of illumos, OmniOS, and SmartOS.)

### DIFF
--- a/link-grammar/utilities.c
+++ b/link-grammar/utilities.c
@@ -757,7 +757,7 @@ bool strtodC(const char *s, float *r)
 {
 	char *err;
 
-#ifdef HAVE_LOCALE_T
+#if defined(HAVE_LOCALE_T) && !defined(__sun__)
 	double val = strtod_l(s, &err, get_C_LC_NUMERIC());
 #else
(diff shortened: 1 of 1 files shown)


dhgutteridge:
We could modify the configure script to check for the presence of strtod_l(3), which would be more involved and intrusive, but, of course, avoids platform-specific macros. Since this source file already has a bunch of platform-specific macros in it, and this is the quickest solution for me, that's what I went with. You may prefer something different.

linas:
Thanks!  Simplicity is almost always better.  There might actually be a one-line autoconf check for `strtod_l` but it would take an hour or two to find it, if it exists, so ... easier-> more time for other things.  Again, thanks.

ampli:
The use of `strtod_l()` addresses issue #804 (Incorrect parsing due to locale settings).
A better fix for this issue could be to add a function to convert a scaled integer string to float.
Such a function is several times faster than `strtod()/strtod_l()`, but I assumed that modern OSs (that have `localet`) also have `strtod_l()` and invested my time in other things...

PR #1102, opened 2020-08-16
Upstreaming Gentoo patches

SoapGentoo:
Hi @linas 
we have a number of patches in Gentoo for link-grammar.

1. Allow building link-grammar out-of-source. We build multiple python bindings, and that requires building everything out-of-source.
2. Do not inject pure user flags. You should *never* inject flags that affect optimizations and such, which should be left to the user. I usually compile with `-O2`, only to notice that `-O3` gets added by the `configure` script.

### DIFF
--- a/bindings/python/Makefile.am
+++ b/bindings/python/Makefile.am
@@ -4,8 +4,8 @@
 # Created by Linas Vepstas February 2014
 #
 #
-SWIG_SOURCES = ../swig/link_grammar.i
-SWIG_INCLUDES = ../../link-grammar/link-includes.h
+SWIG_SOURCES = $(top_srcdir)/bindings/swig/link_grammar.i
+SWIG_INCLUDES = $(top_srcdir)/link-grammar/link-includes.h
 built_c_sources = lg_python_wrap.cc
 built_py_sources = $(top_builddir)/bindings/python/clinkgrammar.py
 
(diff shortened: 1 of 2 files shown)


linas:
Thanks!  I merged ... however, I'm thinking of putting back the -O3 flag .. not using it is an 8% performance hit, which is a rather nasty price to pay. @ampli has put in a huge amount of effort to get good performance, and to surrender this much seems fairly criminal. This stuff doesn't "grow on trees", as they say, you have to work for it.

SoapGentoo:
@linas before you put it back, I have suggestion: let's override the Autoconf default of `-O2 -g` if and only if the user did not specify CFLAGS. That way, I dont have to hack around it, and anyone using the sources gets `-O3` by default. Does that sound like a plan?

linas:
Sure, but I have no idea how to over-ride. 

Related: why the heck would you *want* to use -O2 ?  This is a 20-year old flag; way-back when, it made sense to switch from -O to -O2, and now it makes sense to switch from -O2 to -O3 and I don't see the point of using -O2 for anything at all ... I think -O3 has been stable for a decade+ now ... !?

SoapGentoo:
This is about debuggability. What if I want to get sane backtraces? Backtraces with `-O3` are absolutely useless, as everything has been inlined at that stage and the useful line annotations are gone. With `-O1` or `-O0`, I get useful backtraces. What if GCC 11 jacks up the optimizer and breaks code with `-O3` instead of `-O2`? Being able to change optimization levels is a key part of the GNU conventions.

Overriding `./configure` defaults is a well-documented and supported feature of Autoconf. The PR #1103 makes `-O3` the default, but makes it such that `./configure CFLAGS="-O2"` means I can force `-O2` if I need to for debuggability or other reasons.

linas:
OK, I merged #1103 (side remark: I'm mostly not concerned about the debuggability, at this point; I stick to old-school techniques.)

SoapGentoo:
sure, but you have `-O3` by default, so we're both happy :smiley:

PR #1103, opened 2020-08-16
Specify `-O3` as overridable default in configure.ac

SoapGentoo:
* Inspired by
  https://lists.gnu.org/archive/html/autoconf/2006-04/msg00022.html

@linas 
overriding `CXXFLAGS`:
```
./configure CXXFLAGS="-O1"
[...]
checking whether configure should try to set CFLAGS/CXXFLAGS... no
[...]
C compiler:                     gcc  -D_DEFAULT_SOURCE -std=c11 -D_BSD_SOURCE -D_SVID_SOURCE -D_GNU_SOURCE -D_ISOC11_SOURCE 
C++ compiler:                   g++  -D_DEFAULT_SOURCE -std=c++11 -O1 -Wall
```

Without overriding (the default):
```
./configure
[...]
checking whether configure should try to set CFLAGS/CXXFLAGS... yes
[...]
C compiler:                     gcc  -D_DEFAULT_SOURCE -std=c11 -D_BSD_SOURCE -D_SVID_SOURCE -D_GNU_SOURCE -D_ISOC11_SOURCE -O3
C++ compiler:                   g++  -D_DEFAULT_SOURCE -std=c++11 -O3 -Wall
```
as you can see, with a vanilla `./configure` call, `-O3` is back.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@ Version 5.8.1 (XXX 2020)
  * English dict: fix numerical identifiers used as adjectives.
  * English dict: fix post-posed Latin adjectival modifiers.
  * Merge upstream gentoo patches. #1102
+ * Make -O3 default for CFLAGS/CXXFLAGS, but overridable by the user.
 
 Version 5.8.0 (28 February 2020)
  * Java bindings: Remove the obsolete senses API.
--- a/configure.ac
+++ b/configure.ac
@@ -7,6 +7,20 @@
 AC_INIT([link-grammar],[5.8.0],[https://github.com/opencog/link-grammar],,
 		  [https://www.abisource.com/projects/link-grammar])
 
+dnl Check whether we want to set defaults for CFLAGS and CXXFLAGS
+AC_MSG_CHECKING([whether configure should try to set CFLAGS/CXXFLAGS])
+AS_IF([test "x${CFLAGS+set}" = "xset" || test "x${CXXFLAGS+set}" = "xset"], [
+    enable_flags_setting=no
+    : ${CFLAGS=""}
+    : ${CXXFLAGS=""}
+  ], [
+    enable_flags_setting=yes
+    dnl Set to -O3 here so AC_PROG_CC and AC_PROG_CXX do not add -g -O2
+    CFLAGS="-O3"
+    CXXFLAGS="-O3"
+  ])
+AC_MSG_RESULT([${enable_flags_setting}])
+
 DISCUSSION_GROUP=https://groups.google.com/d/forum/link-grammar
 AC_SUBST(DISCUSSION_GROUP)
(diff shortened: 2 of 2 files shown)


linas:
OK, thanks!

SoapGentoo:
@linas thanks for merging, do you think it would be possible to get a 5.8.1 release soon, so we can remove all the patches we have to carry right now?

linas:
Yes, I was thinking that it's about time for that; I'm hoping that @ampli  gets to take a look at everything ... I'm thinking a new releaase in 1-3 days...

ampli:
I have one bug fix and 2 speedup changes.
I would like to include at least this bug fix in the next version, and maybe one of the speedups (PRs soon).

Regarding -Ox and debugging: I sometimes do not hand-optimize code if this reduces clarity, after checking (by looking at the assembly code) that -O3 optimizes it anyway (but -O2 may not).  I also removed "inline" from many functions that used to have it, because -O3 in ines anyway.

For debugging using a debugger -O3 is indeed useless, but anyway I use -O0 (along with UBSAN and LSAN on the non-dictionary code).

N.B.
1. Maybe I should have used -Og, which is 60% faster for English (it is 30% slower for Russian).
2. Regretfully maintainers on other distinctions don't usually send us their patches. I intend to send PRs for the ones I located (for version 5.8.2).

linas:
OK, thanks, I can wait :-)

ampli:
Using -O2, there is this warning:
``` c
constituents.c:1166:26: warning: ‘m’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 1166 |    last_child = n->child = m;
```
The problem is the use of a private `assert()` that is not defined as `noreturn`.
Theoretically, `noreturn` should also enable some optimizations because then the compiler knows that the asserted condition cannot be true in the code after the `assert()`. Defining the current private `assert()` as a function (instead of macro) will also reduce the code size. So I would also like to include a PR for that (the code yet to be written).

ampli:
With this patch, CPU time profiling using `ppref` (from `gpreftools`) gives a total garbage output.
If I remove this patch then everything is fine again.
Since it is not apparent to me why it is so, if anyone has any hint this may save me a lot of time...

SoapGentoo:
how do you invoke the profiler?

ampli:
> how do you invoke the profiler?
I have just found the problem, and it is not related the invocation (which is basically `pprof  link-parser profile`).

The problem is that `pprof` needs the symbol table.
Originally the default flags were "-g -O2 -O3" (which is actually `g -O3`).
Now it is just `-O3`.
Of course I can just configure with `CFLAGS="-g -O3"`, but I like that libraries and binaries include the symbol table, like done  in many (most?) distributions. So I will send a PR to make the default  `g -O3` instead of just `-O3`.

SoapGentoo:
see #1105

PR #1105, opened 2020-08-17
Don't drop `-g` when specifying CFLAGS/CXXFLAGS

SoapGentoo:
@linas @ampli 
sorry for the churn, I've decided to change it slightly: I've added `-g` to the default `-O3`, so when you run default `./configure` you get `-O3 -g`, which helps with profiling.


ampli:
Thanks!

SoapGentoo:
@ampli coudl you give it a try and merge it if it works for you?

ampli:
Yes, it works fine!
BTW, you can see the actual compilation flags using:
`link-parser --version`

linas:
Thanks!

PR #1108, opened 2020-08-28
Fix assert() problems

ampli:
Fix old commit 5c2aa75352851dc55460653bd0ef85b4647e91fb:
- Fix wrong preprocessor symbol.
- Call `abort()` instead of `0()`.
The rest are fixes to the recent PR #1106.

Possible FIXMEs:
- Use `AX_GCC_BUILTIN(__builtin_trap)`.
- Use debug_break (GitHub `debugbreak`) if avaliable.

### DIFF
--- a/link-grammar/error.c
+++ b/link-grammar/error.c
@@ -413,7 +413,7 @@ const char *feature_enabled(const char * list, ...)
 #elif defined __GNUC__ || defined __clang_analyzer__
 	#define DEBUG_TRAP __builtin_trap()
 #else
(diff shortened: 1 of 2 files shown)


linas:
Do you expect more in the next few days/week, or should I roll a new version? I'm happy to wait a few more weeks.

ampli:
I am working now on several different speed-ups.
They are in a stage of proof-of-concept and I try to bring them to production quality.
I would like to include in the new version at least one of the speed-ups.

linas:
OK!

PR #1113, opened 2020-09-10
Improve efficiency of setting farthest_word and eliminate connector length_limit 

ampli:
Currently, length_limit is set per each new connector that is created. This takes ~1% of CPU.
The expression pruning then computes from it the farthest word for each connector that is inserted into the connector table.
Before power pruning, the connectors length_limit is converted in place to farthest_word.

This patch adds a farthest_word field to `Exp` and first set it there. This setting involves order of magnitude fewer connectors.
Setting it later into Disjunct connectors involved only copying it (instead of a full computation per connector). The expression pruning code is then able to directly use it.

On the same occasion of looking into the expression pruning code, I found a way to save a potential branching per iteration in the connector set matching loop.

### DIFF
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -632,6 +632,12 @@ int sentence_parse(Sentence sent, Parse_Options opts)
 	if (IS_DB_DICT(sent->dict))
 		condesc_setup(sent->dict);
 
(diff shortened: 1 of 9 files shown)


linas:
My apologies for long delay!

PR #1117, opened 2020-09-16
Exact power_table sizing

ampli:
The current power pruning is a bottleneck when there are a vast number of different uppercase connector parts in each disjunct.
This is because their number is overestimated by a large factor. This is not so important for the current dicts because this number is low for them.

For testing, I used a dict with ~500K different uppercase connectors and a sentence with length ~190 and 8 nulls when parsed.
After expression pruning, there were a few thousands of different upper case parts on most disjuncts, in each direction.
W/o this patch, it took about 100 seconds to get a result, and with this patch only 15 seconds (`power_prune()` was about ~10 times more efficient in this test).

BTW:
This doesn't solve the huge slowness that is seen in some language learning runs. It is due to two additional things (one of them or both at once): 1. Inefficiencies in build_disjuncts(). 2. The parsing `Table_connector` is too small for some sentences.
Both can be fixed. (My dynamic `Table_connector` resizing patch is very old but was not important until the long-sentence speedup and the language-learning huge dictionaries. It needs to be adapted to the current `count.c`. In addition, by now I have other related ideas that I still have to test.)

Important: I need the current PRs and this one to be applied before I can send additional ones.

### DIFF
--- a/link-grammar/dict-file/dictionary.c
+++ b/link-grammar/dict-file/dictionary.c
@@ -249,6 +249,7 @@ failure:
 /**
  * Use filenames of six different files to put together the dictionary.
  */
+NO_SAN_DICT
 Dictionary
 dictionary_six(const char * lang, const char * dict_name,
                const char * pp_name, const char * cons_name,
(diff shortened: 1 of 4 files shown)


linas:
> I used a dict with ~500K
> This doesn't solve the huge slowness that is seen in some language learning runs.

Wow!  Thank you for paying attention to this!

Issue #1120, opened 2020-12-29
test multi-thread is unreliable: sometimes fails with exit status: 139


jonassmedegaard:
The `multi-thread` test sometimes fails.  Seems to happen more frequently on slower architectures, and may succeed when rebuilding exact same code.

Excerpt of build-log from failed `make -j2 check VERBOSE=1  || { cat test/test-suite.log; false; }`:
```
make  check-TESTS
make[4]: Entering directory '/<<PKGBUILDDIR>>/tests'
make[5]: Entering directory '/<<PKGBUILDDIR>>/tests'
PASS: multi-dict
../test-driver: line 109: 1743663 Segmentation fault      "$@" > $log_file 2>&1
FAIL: multi-thread
PASS: mem-leak
PASS: dict-reopen
PASS: multi-java
==============================================
   link-grammar 5.8.0: tests/test-suite.log
==============================================

# TOTAL: 5
# PASS:  4
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: multi-thread
==================

link-grammar: Info: Dictionary found at ../data/en/4.0.dict
link-grammar: Info: Dictionary found at ../data/ru/4.0.dict
link-grammar: Info: ru: Spell checker disabled.
FAIL multi-thread (exit status: 139)
```
Full build-log: https://buildd.debian.org/status/fetch.php?pkg=link-grammar&arch=ia64&ver=5.8.0-4&stamp=1609277205&raw=0
Full build-log of repeated succesful build: https://buildd.debian.org/status/fetch.php?pkg=link-grammar&arch=ia64&ver=5.8.0-4&stamp=1609282640&raw=0

ampli:
Hi @jonassmedegaard,
Thank you for the problem report.
As I don't know how to repeat this failure, will you be able to produce a call trace?

jonassmedegaard:
please elaborate - what more precisely do you want me to do? Compile with some specific configure options? Run some custom tools?

ampli:
You will need to locate the core file of the failed `multi-thread` and inspect it using `gdb` as follows:
`gdb multi-thread COREFILE` (when COREFILE is the core file name).
At the `gdb` prompt, issue the command `bt` to obtain a trace.

Before you start the build process,  make sure the core file size limit is set to "unlimited".
[...]

PR #1122, opened 2021-01-01
Fix the multi-thread test

ampli:
Fixes:
1. Use half of the threads for the `ru` dict.
2. Use a separate Parse_options object for each thread.

See issue #1120.

Also, validate that something got parsed so we know a bug doesn't cause it to be a no-op.

### DIFF
--- a/tests/multi-thread.cc
+++ b/tests/multi-thread.cc
@@ -125,9 +125,6 @@ static void parse_sents(Dictionary dict, Parse_Options opts, int thread_id, int 
 int main(int argc, char* argv[])
 {
 	setlocale(LC_ALL, "en_US.UTF-8");
-	Parse_Options optsa = parse_options_create();
-	Parse_Options optsb = parse_options_create();
-	parse_options_set_spell_guess(optsb, 0);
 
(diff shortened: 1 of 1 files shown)


tests/multi-thread.cc
```diff
@@ -137,30 +138,41 @@ int main(int argc, char* argv[])
...
 	printf("Creating %d threads, each parsing %d sentences\n",
 		 n_threads, niter);
 	std::vector<std::thread> thread_pool;
 	for (int i=0; i < n_threads; i++)
 	{
 		Dictionary dict = dicte;
-		if (0 == i%3) dict = dictr;
-
-		Parse_Options opts = optsa;
-		if (0 == i%2) opts = optsb;
```
linas, line 152: Ah hah! This is the bug (maybe): the intent was that the Russian dict would have spell-guessing disabled. And so the old otsb was supposed to always be used with russian. The russian dict was %3 but the opts was %2 and it also should have been %3.

It looks like you fix this ... does the test now always pass?

ampli: This was not enough to always pass the test, but after using a separate `Parse_Option` per thread the test always passes.

PR #1127, opened 2021-01-08
Add diary entries to document current CPU profiling

ampli:
It is interesting to see the CPU percentage taken by the counting parsing stage for various batches.
I didn't normalize for the dictionary read time, so it is slightly more, especially for the short-sentence batches.

| Batch filename | count.c | fast-match.c/fast-match.h |
|-|-|-|
| en/corpus-basic.batch | 0.8% |  0.0% |
| en/corpus-fixes.batch  | 1.2% |  0.6% |
| en/corpus-fix-long.batch  | 27.4% |  6.0% |
| en/corpus-failures.batch  | 24.9% |  6.8% |
| ru/corpus-basic  | 0.6% |  0.4% |
| pandp-union.txt | 15.6% |  3.8% |

It can be seen that for short sentences there is currently no significant room to improve the parsing performance.
However, there is a potential to improve their performance by improving the efficiency of the post-processing (for `en`), memory-allocation, and to a less extent building disjuncts. Some simple post-processing changes already achieved about a 20% performance increase.

For long sentences, there is still much room to improve the parsing performance, in addition to the areas mentioned for short sentences. More pruning can also be added (in the works).


linas:
>  post-processing

Last time I looked (4-5 years ago?) almost half of the post-processing rules were never triggered. If these were removed, speed is likely to improve.

Many rules are very rarely triggered. Perhaps these can be eliminated to changes to the dictionary?

Perhaps the same is true for the most-frequently triggered rule?

Of all the batch files, the one that triggers the most rules is `corpus-basic`.

ampli:
> Last time I looked (4-5 years ago?) almost half of the post-processing rules were never triggered. If these were removed, speed is likely to improve.

From what I remember from the time I checked it, most of these are not triggered because the relevant connectors are removed by `pp_prune()`.  

> Many rules are very rarely triggered. Perhaps these can be eliminated to changes to the dictionary?

If there are rules that are not triggered of course they can be removed. But my intended changes are said to make their impact negligible. I will check that again when I get again to work on that branch.

linas:
version 5.8.1 is out

PR #1128, opened 2021-01-11
Dictionary #define

ampli:
See issue #785.

- Add dict support for `#define name value;`.
- Convert the domain and locale definitions to use `#define`.
- Accept also the old-style domain and locale definition support (for backward compatibility).
- Add support for reading the max-cost from the dict.
- Define max-cost only for the `en` dict for now. The other dicts are to be updated later.

The correct naming of some variables and definitions (as I mentioned in issue #785) was not clear to me and may need updates.

### DIFF
--- a/data/ar/4.0.dict
+++ b/data/ar/4.0.dict
@@ -2,7 +2,7 @@
 %
 % Copyright (c) 2006 Warren Casbeer and Jon Dehdari
 %
-#define dictionary-version-number 5.0.1;
+#define dictionary-version-number 5.9.0
 #define dictionary-locale C; % Only the English alphabet is in use here
 
 
--- a/data/ar/4.0.dict
+++ b/data/ar/4.0.dict
@@ -3,7 +3,7 @@
 % Copyright (c) 2006 Warren Casbeer and Jon Dehdari
 %
 #define dictionary-version-number 5.0.1;
-<dictionary-locale>: C+; % Only the English alphabet is in use here
+#define dictionary-locale C; % Only the English alphabet is in use here
 
 
 % Link Types:  [Forms in brackets roughly equate with English Link Grammar]
--- a/data/ar/4.0.dict
+++ b/data/ar/4.0.dict
@@ -2,8 +2,7 @@
 %
 % Copyright (c) 2006 Warren Casbeer and Jon Dehdari
 %
-% Dictionary version number is 5.0.1 (formatted as V5v0v1+)
-<dictionary-version-number>: V5v0v1+;
+#define dictionary-version-number 5.0.1;
 <dictionary-locale>: C+; % Only the English alphabet is in use here
 
 
--- a/data/de/4.0.dict
+++ b/data/de/4.0.dict
@@ -7,7 +7,7 @@
  %                                                                           %
  %***************************************************************************%
 
-#define dictionary-version-number 5.0.2;
+#define dictionary-version-number 5.9.0;
 #define dictionary-locale         de_DE.UTF-8;
 
 % NOUNS, PRONOUNS
--- a/data/de/4.0.dict
+++ b/data/de/4.0.dict
@@ -7,8 +7,7 @@
  %                                                                           %
  %***************************************************************************%
 
-% Dictionary version number is 5.0.2 (formatted as V5v0v2+)
-<dictionary-version-number>: V5v0v2+;
+#define dictionary-version-number 5.0.2;
 <dictionary-locale>: DE4de+;
 
 % NOUNS, PRONOUNS
--- a/data/de/4.0.dict
+++ b/data/de/4.0.dict
@@ -8,7 +8,7 @@
  %***************************************************************************%
 
 #define dictionary-version-number 5.0.2;
-<dictionary-locale>: DE4de+;
+#define dictionary-locale         de_DE.UTF-8;
 
 % NOUNS, PRONOUNS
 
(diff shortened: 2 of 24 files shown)


data/ar/4.0.dict
```diff
@@ -2,9 +2,8 @@
 %
 % Copyright (c) 2006 Warren Casbeer and Jon Dehdari
 %
-% Dictionary version number is 5.0.1 (formatted as V5v0v1+)
-<dictionary-version-number>: V5v0v1+;
-<dictionary-locale>: C+; % Only the English alphabet is in use here
+#define dictionary-version-number 5.0.1;
```
linas, line 5: Should be 5.9.0 -- all dicts with this change will be 5.9.0, since, obviously, these dicts are no longer compatible with older versions of LG ... older versions of the binary will error out, if used with new dicts.

data/de/4.0.dict
```diff
@@ -7,8 +7,7 @@
  %                                                                           %
  %***************************************************************************%
 
-% Dictionary version number is 5.0.2 (formatted as V5v0v2+)
-<dictionary-version-number>: V5v0v2+;
+#define dictionary-version-number 5.0.2;
 <dictionary-locale>: DE4de+;
```
linas, line 11: locale, too

data/en/4.0.dict
```diff
@@ -10,9 +10,11 @@
  %                                                                           %
  %***************************************************************************%
 
-% Dictionary version number is 5.8.1 (formatted as V5v8v1+)
-<dictionary-version-number>: V5v8v1+;
-<dictionary-locale>: EN4us+;
+#define dictionary-version-number 5.8.1;
+#define dictionary-locale         en-US;
```
linas, line 14: I'm pretty sure locales always use underscore.If you don't do that, you might find yourself writing dash-to-underscore converters. ... Oh, I see below you are using `sscanf` so I guess it doesn't really matter... but still, compatibility would be best.

Note also: on my system, `LANG=en_US.UTF-8` -- we don't need to support this, but, for compatibility, it should be harmless ...

linas (approved):
Thanks! Looks good. Let me know when you want this merged.

data/de/4.0.dict
ampli: In `he` I forgot that too....

data/en/4.0.dict
ampli: I intentionally used `-` so it doesn't look like a valid POSIX locale `en_US` which is a `latin1` locale.
Because it may still be misleading, I can change it to use the full name of a `UTF-8` locale (just more new code instead of essentially copy-paste and slightly modify old code...).

linas: We don't have to support full POSIX locales. Just being POSIX-like is good enough, and if someone ever complains, then its a new bug report. Simplicity, and approximate compatibility are good things.

ampli:
I added `#define max-disjunct-cost 2.7;` only to the `en` dict as documentation and test.
Should I add this definition to all the dicts or we can just depend on the hard-coded default  value?

data/ar/4.0.dict
ampli: I added a commit for that.

ampli:
I have just force-pushed a fixed PR according to the discussion at issue #785.

Here are the list of names to ease the consistency check (quoting whole lines for reference):
``` c
#define dictionary-version-number 5.9.0;
#define dictionary-locale         en_US.UTF-8;
% The default largest disjunct cost to consider during parsing.
#define max-disjunct-cost 2.7;

/* If the maximum disjunct cost is yet uninitialized, the value defined in the
 * dictionary (or if not defined then DEFAULT_MAX_DISJUNCT_COST) is used. */
static const double UNINITIALIZED_MAX_DISJUNCT_COST = -10000.0;
static const double DEFAULT_MAX_DISJUNCT_COST = 2.7;
/* We need some of these as literal strings. */
#define LG_DISJUNCT_COST                        "max-disjunct-cost"
#define LG_DICTIONARY_VERSION_NUMBER            "dictionary-version-number"
#define LG_DICTIONARY_LOCALE                    "dictionary-locale"

linkgrammar_get_dict_define(Dictionary, const char *);
linkgrammar_get_dict_max_disjunct_cost(Dictionary);
```

Changes from the initial PR:
- Names.
- Use a POSIX UTF-8 locale name. Non-UTF-8 locale names are not supported and produce an error.
- Change dicts version to `5.9.0`.
- `LG_DISJUNCT_COST` also uses C #define, for consistency with `LG_DICTIONARY_VERSION_NUMBER` and `LG_DICTIONARY_LOCALE`.
- Construct the dictionary `#define`s using a function `add_define`, to ease adding `#define` to the SQL DB if desired.
- Move the setting of all the `#define`s into `dictionary_setup_defines()`.
- Fix a crash on opening the SQL DB.
- Fix a memory leak.
- Warn on redefinition. (Last value is used. The line number in the warning is w/offset +1 as usual - FIXME).

If all the names are consistent now, this PR may be applied.

linas:
Thanks!

> Should I add this definition to all the dicts or we can just depend on the hard-coded default value?

Yes!  At this time, I don't see that this matters. BTW, sometimes, I don't answer questions for two reasons: (a) I didn't see the question, or (b) the answer doesn't seem to be important. I didn't want to be rude, and not answer this question,. but ... answering it also does not seem important. I'm just sayin, in case you find something annoying, it might be because of case (a)...

PR #1130, opened 2021-01-17
Panic mode fixes + SIGWINCH handling

ampli:
See issue #785.
The main changes in this PR are related to panic mode.
Handling SIGWINCH is also implemented here (it may be difficult to separate).
Some other minor changes are not related to the panic mode but are interleaved with its changes.
The `linkgrammar_get_configuration()`  commits are not directly related but I wanted to see the feature_test_macros that expose `sigsetaction()`.
Another non-related change is supporting displaying the correct max-cost even before the first parse. This change also supports its correct default value display.

`!help panic_timeout` (or `!help` on any other `panic_*` variable) shows all the settable `panic_*` variables and the rules that govern their usage in panic mode. `!panic_<TAB>` also shows the `panic_*` variables.

I also updated the `!help` text. As usual, it may need English polishing.

Other changes:
- Add `panic-max-disjunct-cost` in the `en` and `ru` dicts.
-  Add `max-disjunct-cost` in the `ru` dict for documentation of what is actually used.
- Set panic_cost-max from the dict if available.
- Fix memory leak.
- Implement accurate timeout. This doesn't include a timeout during link extraction (can be fixed).
- In case of a panic, don't display partial results, as they are inconsistent and different for each run.
- Fix 2  "No complete linkages found" notification problems.
- resources_exhausted(): Print indication on verbosity > 1. Make it easier to track panic handling with `-verbosity=2`.
- Bug fix: Don't try the 2nd step if 1st one timeouts (this fixes 3 problems).
- Support `!bad` also for `!nulls=1`.

Implementation notes:
- The panic mode for `!nuls=1` is now done using the "one step" parse.
- In `link-parser` I retained, for now, the original 2-parse step even though the library can do it in "one step", unless  `-test=one-step-parse` is used. (On the language binding, a "one step" parse is done when `min_null_count==0 && max_null_count>0`, but the user is of course free to do a 2-step parse if desired.)
- There is currently no infrastructure to test the correct behavior of timeouts in various places (correct parsing and no memory leaks). I tested it by inserting (one at a time) `do {} while (!resources_exhausted(r))` before timeout checks and using `-timeout=1 -panic_timeout=1`.
- Panic mode for SAT is still not implemented (more details at issue #785).
- resources_exhausted() (and the related parse option)  is still being used for a timeout check. To be replaced by the timer check related calls in a future cleanup.

Note that the introduction document is not up to date regarding max-cost: https://www.abisource.com/projects/link-grammar/dict/introduction.html#:~:text=only%20connectors,morphology.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -88,7 +88,7 @@ AC_LANG([C])
 
 AC_C_CONST
 
-AC_CHECK_FUNCS(strndup strtok_r)
+AC_CHECK_FUNCS(strndup strtok_r sigaction)
 AC_CHECK_FUNCS(aligned_alloc posix_memalign _aligned_malloc)
 
 AC_FUNC_ALLOCA
--- a/data/command-help-en.txt
+++ b/data/command-help-en.txt
@@ -423,7 +423,7 @@ Here are their names, roles and initial values, as shown by
  Variable             Controls                              Default value
  --------             --------                              -------------
  panic_short          Max length of all links                          12
- panic_cost-max       Largest cost to be considered                  4.00
+ panic_cost-max       Largest cost to be considered                  4.00 [*]
  panic_limit          The maximum linkages processed                 1000
  panic_max-null-count Max number of null links allowed                 10
  panic_spell          Up to this many spell-guesses per unknown word    0
@@ -431,12 +431,13 @@ Here are their names, roles and initial values, as shown by
 
 Their values are used in panic mode according to the following rules:
 
- !panic_short          - replaces !short if its value is lower.
- !panic_cost-max       - replaces !cost-max if its value is higher [*].
+ !panic_short          - replaces !short if its value is lower. [**]
+ !panic_cost-max       - replaces !cost-max if its value is higher.
  !panic_limit          - replaces !limit if its value is lower.
  !panic_max-null-count - the maximum allowed null links.
  !panic_spell          - replaces !spell if its value is lower.
  !panic_timeout        - the timeout of the panic mode parsing.
 
-[*] The library all_short_connectors parse option is always set in panic
-mode, so !short determines the maximum length of all the links.
+[*]  Unless set in the dictionary to another value.
+[**] The library all_short_connectors parse option is always set in panic
+mode, so !panic_short determines the maximum length of all the links.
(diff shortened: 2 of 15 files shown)


linas:
> the original 2-parse step 

* Would it run faster if it was changed to one-step parse?
* Would the code in `link-parser` be smaller/simpler?

I assume the answers are "no" and "yes".  Making `link-parser` code smaller and simpler is a good thing, as it is our defacto "main example" of how to use the C API. Thus, it should be as simple as ... reasonable.  It might be best if the parsing code was just in one small-ish file, while all the options and other non-parsing stuff is in a different file.

ampli:
> * Would it run faster if it was changed to one-step parse?

It would run faster for parses with nulls.

> * Would the code in `link-parser` be smaller/simpler?

It would be simpler by a little bit. However, it will then be a good idea to add -test=two-steps-parse, in order to be able to check the library for correctness by validating validate that both methods give the same result. So the complexity at the end would be exactly the same (just reversing the if/else).

> It might be best if the parsing code was just in one small-ish file, while all the options and other non-parsing stuff is in a different file.

This can be done independently and I will eventually check that  (I have a lot of other code to submit).

linas:
> check the library for correctness 

This kind of check belongs in a unit test, not in the application...

ampli:
> This kind of check belongs in a unit test, not in the application...

It is very hard to do this kind of test using our test suite.
This is the test I do:
I compare the results of two branches, the master branch and the WIP one.

For each, I generate graphical linkages (up to 27K linkages per sentence, using `-limit=150000`) with null links allowed, of the basic batches of all the languages + en/fixes is several ways: 
- With and w/o islands
- With `-test=min-len-encoding:0` and `:254`
- With and w/o `-test=one-step-parse`
- With and w/o `-test=always-parse`

and then compare them using "diff". Multi-gigabyte files are generated and compared. Even one character difference is not allowed (except in the P.P. violation count).

This test uses multiple shell scripts, that run an 8 `link-parser` process in parallel (so it finishes in a reasonable time).
However, I can do it using a python script that mimics the current link-parser.
For testing `link-parser` itself, I have a WIP that uses a Python test suite for shell commands (that I don't use for actual tests since it is not yet in production state).

To sum up: I will add to my tasks simplifying `link-parser` to use only "one-step-parse" and separating it as you suggested.

linas:
>  I generate graphical linkages 

Wow!  Well, uhh, do not let me get in your way, or make suggestions that make your life harder!  When I made my original comment, I was assuming a unit test that is 50 lines of code long, and did some basic compares, and nothing more.

PR #1132, opened 2021-01-18
Speed up counting by 1. word-vector 0-count cache; 2.Boyer-Moore-like word skips

ampli:
The changes here have the following speedups on my computer:
|Batch file|~Speedup|
|--|--|
|fix-long|9%|
|failures|5%|
|pandp-union|1.2%|

The actual speedup of `do_count()` in several times this figure since `do_count()` itself takes <30% of the total CPU (see the table in the first comment of PR #1127).

The speedup is due to two main changes:
1. Implementation of the 0-count tracon-word cache by a word-vector indexed by the tracon number. Instead of the current cache which is a hash table, this cache is an array, indexed by the tracon number. The vector includes elements only from/until the farthest_word of the left/right tracon. One cache access is needed to locate the prediction for all the words relevant for each tracon, as opposed to the current hash access per word. This implements a FIXME.
2. A Boyer-Moore style word skipping is implemented to skip over a range of words that are known to yield a zero count due to previously cached results (the idea got mentioned in the first comment of the demo PR #887). This saves word-vector cache lookups.

The last commit updates a comment with a detailed explanation of how these changes work.

This PR also introduces `pool_alloc_vec()`, which is used for the allocation of the word-vector cache elements.

### DIFF
--- a/link-grammar/memory-pool.c
+++ b/link-grammar/memory-pool.c
@@ -13,7 +13,7 @@
 
 #include "error.h"
 #include "memory-pool.h"
-#include "utilities.h"                  // MIN/MAX, aligned alloc, lg_strerror
+#include "utilities.h"                  // MIN/MAX, aligned_alloc, lg_strerror
 
 /* TODO: Add valgrind descriptions. See:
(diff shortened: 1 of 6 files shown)


linas:
Wow.

PR #1137, opened 2021-02-09
Disable maintainer mode (fix for #649)

### LINKED ISSUE #649: Requires automake after configure is done
Once the supplied ```configure``` has been run, there should be no need for ```automake```.
All ```Makefile.in``` files should be present along with ```configure```.

Error:
```
        Swig interfaces generator:      no
        Perl interfaces:                no
        Perl install location:          
        Python2 interfaces:             no
        Python3 interfaces:             no
        ASpell spell checker:           no
        HunSpell spell checker:         yes
        HunSpell dictionary location:   /usr/share/myspell/dicts
        Boolean SAT parser:             no
        SQLite-backed dictionary:       yes
        Viterbi algorithm parser:       no
        Corpus statistics database:     no
        RegEx tokenizer:                no

===========================================================================
=======================<phase: build          >============================
===>  Building for link-grammar-5.4.3
gmake[1]: Entering directory '/wrkdirs/usr/ports/textproc/link-grammar/work/link-grammar-5.4.3'
 cd . && /bin/sh /wrkdirs/usr/ports/textproc/link-grammar/work/link-grammar-5.4.3/missing automake-1.15 --foreign Makefile
/wrkdirs/usr/ports/textproc/link-grammar/work/link-grammar-5.4.3/missing: automake-1.15: not found
WARNING: 'automake-1.15' is missing on your system.
         You should only need it if you modified 'Makefile.am' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'automake' program is part of the GNU Automake package:
         <http://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <http://www.gnu.org/software/autoconf>
         <http://www.gnu.org/software/m4/>
         <http://www.perl.org/>
```

linas:
This fixes #649 .. @ampli can you please review? I'm not entirely clear what will happen after doing this.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,7 +52,10 @@ AC_SUBST(LINK_MINOR_VERSION)
 AC_SUBST(LINK_MICRO_VERSION)
 
 AM_INIT_AUTOMAKE([foreign subdir-objects])
-AM_MAINTAINER_MODE([enable])
+
+# When maintainer-mode is enabled, downstream users sometimes trip
+# trip over automake issues. See bug #649.
+AM_MAINTAINER_MODE([disable])
 
 AC_CONFIG_MACRO_DIR([m4])
 


ampli:
This was the original default (besides not using explicit `disabled`).
When not enabling it by default, maybe we should add in the main `README.md` that tarball users who would like to modify the build system, language binding, or the lex file, are advised to run  `configure --enable-maintainer-mode` to generate makefiles that support that.

linas:
Isn't it enough to just run `./autiogen.sh` by hand? I am assuming that anyone smart enough the modify source code will also be smart enough to work with the build system. (The README is kind of too long already; the recent email discussion about link-grammar json formats revealed that 4-5 different people didn't read the README.)

ampli:
>Isn't it enough to just run ./autiogen.sh by hand?

It is enough.

>I am assuming that anyone smart enough the modify source code will also be smart enough to work with the build system.

I assume this too.

However, as @bapt mentioned, package maintainers should also know they can use `--disable-maintainer-mode` to avoid the regeneration-attempt problem. So it is not clear to me, after all, why it is better to use `AM_MAINTAINER_MODE([disable])`.

linas:
> So it is not clear to me, after all, why it is better

Oh that's easy: the tarballs that I make have maintainer mode enables, and on rare occasions, this causes problems for users. That's what #649 documents. The goal is to minimize issues for tarball users.

ampli:
>The goal is to minimize issues for tarball users.

For that end this change may be fine.

linas:
thanks :-)

PR #1142, opened 2021-02-22
Memory handling fixes 

ampli:
The main fixes here:
1. Fix a severe error in the combination of `pool_alloc_vec()` and `pool_next()`.
Problem description:
In the recent change of `pool_alloc()` to `pool_aloloc_vec()`, I introduced an off-by-one error that causes it not to allocate the last component of any block. This is (very) slightly wasteful but it is not harmful by itself. However, `pool_next()` iterates over all the elements of a block, including its last component (which has not been allocated and thus contains random values).
Currently, `pool_next()` is only used when growing the count table. This puts a random element every 16K elements (the number of elements in the allocated blocks there) in the table.
However, it is unlikely to cause a parsing problem because its table position needs to match the hash value, and in addition `do_count()` needs to look up this tracon combination (and also 0 is a likely value of the count field and absolutely most lookups find 0-count anyway). So it is not urgent to issue a new version immediately just due to this problem.

2. Fix a stale-memory problem when using the SQL-dict in a typical way from Python.
Problem description:
When running the test suite with ASAN and using the fake memory-pool allocator (used for memory allocation debugging), a stale-memory reference is detected in the SQL-dict parse of "This is another test" when the code produces the diagram for the first linkage of that sentence. Here is the place that causes the stale memory:
https://github.com/opencog/link-grammar/blob/9f7c4fb429c4ff4bb787306d43f66c98846cbb6d/bindings/python-examples/tests.py#L1147-L1148 
In line 1148, the linkage variable, which at this point holds the linkage of the previous sentence, is overwritten. This destroys the last reference to the Sentence object (of the **previous** sentence), which deletes it and calls  `sentence_delete() for it. Here is the relevant C code:
https://github.com/opencog/link-grammar/blob/9f7c4fb429c4ff4bb787306d43f66c98846cbb6d/link-grammar/api.c#L529-L533

The problematic line is 531. It frees the `condesc` table, which contains the connector strings. However, at this point, it already contains the connectors of the linkage of the **current** sentence, which has got parsed at line 1147 in the above Python code.
Just after that, the diagram-creating function called at line 1168 (see below) refers to this stale memory. 
https://github.com/opencog/link-grammar/blob/9f7c4fb429c4ff4bb787306d43f66c98846cbb6d/bindings/python-examples/tests.py#L1168-L1168
This problem doesn't happen (typically) when the SQL-dict is used from C, because sentences are deleted just after they are parsed (and not after an additional sentence is parsed).
\
The fix was to just disable this C line (531).

3. Fix the fake memory-pool allocator (that got miss-fixed last time I touched it). This time I made extensive tests using it.

### DIFF
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -132,7 +132,7 @@ struct Sentence_s
 	size_t length;              /* Number of words */
 	Word  *word;                /* Array of words after tokenization */
 	String_set *   string_set;  /* Used for assorted strings */
-	Pool_desc * fm_Match_node;  /* Fast-matcher Match_node memory pool */
+	Pool_desc * Match_node_pool;
 	Pool_desc * Table_connector_pool; /* Count memoizing memory pool */
 	Pool_desc * wordvec_pool;   /* For tracon-word zero-count memoizing */
 	Pool_desc * Exp_pool;
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -521,7 +521,7 @@ void sentence_delete(Sentence sent)
 	lg_exp_stringify(NULL);
 
 	global_rand_state = sent->rand_state;
-	pool_delete(sent->fm_Match_node);
+	pool_delete(sent->Match_node_pool);
 	pool_delete(sent->Table_connector_pool);
 	pool_delete(sent->wordvec_pool);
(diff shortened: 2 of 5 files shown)


linas:
> So it is not urgent to issue a new version immediately just due to this problem.

Well, then, there should still be a new version "soon" ...

PR #1143, opened 2021-02-22
count.c fixes

ampli:
The main fix here is fixing an uninitialized variable in this commit:
* is_lrcnt(): Fix returning without assigning a value to null_start

It looks like it could have bad implications. However, with a detailed parsing with nulls of all the "basic" corpuses + "en/fixes" (with `limit=150000`) it makes absolutely no difference. Strangely, ASAN doesn't show any problem - I found it with valgrind.

Another important change (for debugging only) is adding a Table_connector validation:
* table_stat(): Validate the table

The uninitialized variable fix can be yet another reason to issue a new version "soon" (but it still not "urgent").
However, with the stress tests of using the `en` dict for sentence generation, I may detect some more problems.

---
The next thing I would like to fix in `count.c` is the Table_connector size.
Currently, the maximum size is 2^24.  But it seems there is no reason to limit it since it can still consume more memory by added buckets, which consume most of its memory anyway. Limiting it just causes a vast slowness if the table becomes very loaded (long bucket lists).
However, it uses an `unsigned int` for its size and its size is a power-of-2, so it cannot go over 2^31 entries without additional changes. But the main issue is that using `size_t` hash values will enlarge the bucket size (memory inefficiency) since it holds the hash value (for efficient table growth) and will make it (the bucket size) not a power-of-2 (indexing inefficiency). This can be solved by making the count 24-bit (I have an old patch for that but I will need to apply it by hand...).
Another thing that I would like to try is allocating a table size of a few bytes less than a power-of-2 (say by 16 bytes). Currently `malloc()` may allocate internal double allocations since it needs a few bytes for bookkeeping. This will allow a table size of 2^32-16 entries (the last 16 entries can be `folded` over other entries). Preventing double allocation for Table_connector may not be so important for systems that allow overcommitted memory (some don't allow), but the same problem exists for pool allocation, which is potentially more problematic.

### DIFF
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -312,6 +312,7 @@ static inline unsigned int pair_hash(int lw, int rw,
 	i = r_id + (i << 6) + (i << 16) - i;
 #endif
 
+	if (i == 0) i = 1; /* Allow for table hash field validation. */
 	return i;
 }
 
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -140,6 +140,7 @@ static void table_alloc(count_context_t *ctxt, unsigned int shift)
 
 	if (shift == 0)
 		shift = ctxt->log2_table_size + 1; /* Double the table size */
+	lgdebug(+D_COUNT, "Connector table log2 size %u\n", shift);
 
 	/* Keep the table indefinitely (or until exiting), so that it can
 	 * be reused. This avoids a large overhead in malloc/free when
@@ -210,7 +211,6 @@ static void init_table(count_context_t *ctxt, Sentence sent)
 #endif
 
 	if (MAX_LOG2_TABLE_SIZE < shift) shift = MAX_LOG2_TABLE_SIZE;
-	lgdebug(+D_COUNT, "Initial connector table log2 size %u\n", shift);
 
 	table_alloc(ctxt, shift);
 }
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -552,19 +552,18 @@ static void generate_word_skip_vector(count_context_t *ctxt, wordvecp wv,
  * Is the range [c, w) going to yield a nonzero leftcount / rightcount?
  *
  * @param ctxt Count context.
- * @param dir Direction - 0: leftcount, 1: rightcount.
+ * @param dir Direction - 0: leftcount; 1: rightcount.
  * @param c The connector that starts the range.
- * @param w The word that ends the range.
- * @param null_count The current null_count to check.
- * @param cw The word of this connector.
+ * @param wordvec_index Word-vector index.
+ * @param null_count The current null-count to check.
  *
  * Return these values:
  * @param lnull_start First null count to check (the previous ones can be
  * skipped because the cache indicates they yield a zero count.)
  * @return Cache entry for the given range. Possible values:
- *    NULL - A nonzero count may be encountered for null_count>=lnull_start.
+ *    NULL - A nonzero count may be encountered for \c null_count>=lnull_start.
  *    Table_lrcnt_zero - A zero count would result.
- *    Cache pointer - An update for null_count>=lnull_start is needed.
+ *    Cache pointer - An update for \c null_count>=lnull_start is needed.
  */
 static Table_lrcnt *is_lrcnt(count_context_t *ctxt, int dir, Connector *c,
                              unsigned int wordvec_index,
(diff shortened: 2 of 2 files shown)


linas:
> with the stress tests of using the en dict for sentence generation

Could you provide a semi/quasi-working version for the simplest case of small, simple dictionaries? The current language learning pipeline is complicated to set up, and takes a long time to run, and needs a documentation refresh, so I'd like to get started on that.

> `count.c`

Sounds reasonable to me...

linas:
> Could you provide

If this is a WIP branch in your tree, I can work from that. (as long as its in github where I can pull it.)  Merging into master is *not* urgent.

I'm going to explore your github account now :-)

ampli:
> Could you provide a semi/quasi-working version
 
OK.

My WIP still has the following problems:
1. Max Table_connector table size of 2^31, when long sentences may need more (for speed).
2. Disjunct costs may set incorrectly (some more work is needed for setting them correctly).
3. Something seems to be wrong with the random selection of words from word categories, as they don't look so random. I still need to debug that.
4. Setting of generation mode is done by appending `-generate` to the dict argument (a hack needed due to a bug that I recently introduced and I fixed today - will shortly send a PR for that bug fix and will letter change it to use `-test=generate`).
5. Some code is not yet committed (I intended to organize it into logical commits but for now I will most probably make one big commit and properly recommit it later).

So will commit the uncommited code and will push it to my github account under branch "generate" (starting to work on that now).

linas:
For me, 3. is the highest priority, but still not urgent; it will take several weeks (or more?) to dust off and power up the learning pipeline. After that, all experiments will be on small dictionaries, for a long, long time.  (small input dictionaries .. output dictionaries may be huge. I still don't quite understand how to verify if two dictionaries are substantially the same, but differ in size ... all this requires new work.)

I don't understand 4. Isn't this a brand-new binary?

ampli:
I uploaded it to my page as "generate".

Usage example:
`echo 14 |  link-parser de-generate -v=5 -panic -test=auto-next-linkage:100 -limit=1000 -sp=0 -time=1000000`
It will use the `de` dict, randomly select 1000 linkages, randomly select words from categories, and print 100 sentences
(sorted as usual, but the costs are not handled yet).

For a production version, we will need to decide on more reasonable arguments.
I propose something like `-test=generate:14` (14 is the sentence length).

Some comments:
1. For a large number of generated sentences, e.g. 1M, the sorting takes time. An argument to disable it can be added.
2. There is a test "leave-subscripts" (not tested). I don't know if this is useful.
3. The verbosirty=5 argument shows interesting info on the amount of disjunct/connector compression.

ampli:
> I don't understand 4. Isn't this a brand-new binary?

The generation mode is actually a relatively small change to the library (currently ~400 lines). If you don't append this `-generate` you just get the ordinary LG library. It is designed so it can eventually be merged with the master brange (if desired).

ampli:
> I still don't quite understand how to verify if two dictionaries are substantially the same, but differ in size

Maybe you can check the percentage of generated sentences from one dict that has a full parse by the other dict.

linas:
FYI, I just now started a command-line tool for generation; its in my repo, branch "generate"; it does not yet do anything.

linas:
OK, so ... for generating sentences: as mentioned before, there is now a tool for generating sentences, in my repo, branch `generate`. The tool is called `link-generator` and is used as follows:
```
link-generator -l data/en -s 6 -c 1500
```
to generate up to 1500 sentences of length 6. I've run this up to length 13 and as many as 150K sentences and it seems to work fine. Then I hand-crafted a dictionary, and it fails. The dict is:
```
#define dictionary-version-number 5.9.0;
#define dictionary-locale         en_US.UTF-8;

<verb>: S- & O+;

<noun>: (D- & S+) or (D- & O-);

<determiner>: D+;

cat bird dog squirrel: <noun>;

a the: <determiner>;

saw chased: <verb>;

% Mandatory, else the generator won't work.
<UNKNOWN-WORD>:  XXXXXXXX+;
``` 
and I get:
```
$ link-generator -l ../fake-lang -s 5 -c 10 
#    
# Corpus for language: "../fake-lang"                                                         
# Sentence length: 5                                                                          
# Requested corpus size: 10
link-grammar: Info: Dictionary found at ./../fake-lang/4.0.dict                               
# Linkages found: 1                                                                           
# Linkages generated: 1                                                                       
#                                                                                             
 the squirrel chased the squirrel                                                             
# Bye. 
```
which is .. wrong. If I remove the word squirrel from the dictionary, I get
```
 the dog chased the cat
```
So .. that's weird...

linas:
I replied on #1146 where this discussion should have been heppening.

PR #1147, opened 2021-02-28
list_links(): Improve the function comment

ampli:
I clarified the explanation and fixed the mathematical notations.
I guess it may still need further clarification.

---
Regarding `n` in `list_links()` (line 764):
https://github.com/opencog/link-grammar/blob/6a4c1b63a8a01c303f534502aee97ffdba249c60/link-grammar/parse/extract-links.c#L761-L776
There is still extra overhead in multiplying two 64-bit numbers (and having a 64-bit count in general). I will send a PR to make the count 32-bits (I still need to port that from an old branch).

On the same occasion: I wonder why `set_node_overflowed()` is needed at all.
It seems it would indicate an overflow iff `(sent->num_linkages_found > PARSE_NUM_OVERFLOW)` and we already know `sent->num_linkages_found` at the time `set_node_overflowed()` is invoked.

### DIFF
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -727,7 +727,8 @@ static void issue_links_for_choice(Linkage lkg, Parse_choice *pc)
  *
  * Each "level" in the parse-set tree consists of a linked lists of
  * Parse_choice elements. Each such element is pointed to by a
- * Parse_choice element of an upper level.
+ * Parse_choice element of an upper level. Each parse_choice element
+ * contains two Parse_set elements, that are denoted below as S0 and S1.
  *
  * The algo is based on our knowledge of the exact number of paths in each
  * Parse_set element. Note that the count of the root Parse_set (used at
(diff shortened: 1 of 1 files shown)


linas:
Thanks!  Is parse-choice and parse-set described in the original LG paper? I re-read the "A Robust parser..." paper and it does NOT describe this. (It does describe null links ... which is oddly similar to some other trick you (or I??) (re-)invented for dealing with "empty word" in morphology.) I've day-dreamed about adding some similar trick for supporting "zero copula" and other "zero-word" tricks.

Re: overflow: a long long time ago, the parse counter was 32 bit and it easily overflowed whenever right and left exceeded 65536 -- I changed it to 64-bit everywhere, but was not very careful in how I did that. I was paranoid, and added more checks than needed. So you might be seeing old cruft.

ampli:
> Is parse-choice and parse-set described in the original LG paper? I re-read the "A Robust parser..." paper and it does NOT describe this.

I couldn't find any place in which they are described.

> which is oddly similar to some other trick you (or I??) (re-)invented for dealing with "empty word" in morphology.

The null-links described there are the (virtual)  links connecting words with no links as if they were optionally added to each disjunct. The optional-words that I introduced in order to replace empty-words are words that may have no links at all without be considered as null-words as if they don't exist in the sentence at all.

> I've day-dreamed about adding some similar trick for supporting "zero copula" and other "zero-word" tricks.

Optional words can be used in places you suspect a "zero-word" may be inserted. Their insertion can be directed by the dict. In the main `README.md` you wrote a suggestion about using a "z" connector prefix, but I didn't understand it.

> Re: overflow: ...
>So you might be seeing old cruft.

Eliminating it can save 1-2% CPU on long sentences.
However, instead of eliminating this parse tree pass, it can maybe used to remove count overflows by eliminating high-cost disjuncts (in decreasing order of cost) until the count is reduced below a desired one (when a special parse-option is specified). This is similar to dynamically apply the suggestion "Consider retrying the parse with the max allowed disjunct cost set lower." which is printed by the library in that case.

ampli:
> It does describe null links ... which is oddly similar to some other trick you (or I??) (re-)invented for dealing with "empty word" in morphology.

You are right - optional words (that replace the usage of empty-words) may be indeed null-linked. I edited my answer above to clarify that.

PR #1149, opened 2021-03-07
Free kept_table on thread-exit

linas:
Some users of link-grammar create hundreds of thousands of threads.
The tables leaked RAM very quickly in this case.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,7 @@ Version 5.9.0 (XXX 2021)
  * Fix very rare crash/corruption introduced in v.5.8.1 #1142
  * English dict: fix problems with "just/only".
  * English dict: work on hesitation markers.
+ * Fix multi-threading mem-leak. #1149
 
 Version 5.8.1 (8 January 2021)
  * Fix macOS/SunOS build break.
--- a/configure.ac
+++ b/configure.ac
@@ -538,7 +538,6 @@ if test "$do_aspell" = yes ; then
 	CPPFLAGS=$save_cpp_flags
 	if test  "x${ASpellFound}" = "xyes"; then
 		AC_DEFINE(HAVE_ASPELL, 1, [Define for compilation])
-		ASPELL_LIBS="${ASPELL_LIBS} -lpthread"
 		AC_SUBST(ASPELL_LIBS)
 		AC_SUBST(ASPELL_CFLAGS)
 	fi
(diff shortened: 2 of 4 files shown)


ampli:
The `pthread` related stuff should be protected by an `#ifdef` to prevent compilation break on MSVC.
One way to do that is to use [`AX_PTHREAD`](https://www.gnu.org/software/autoconf-archive/ax_pthread.html) in `configure` + `HAVE_PTHREAD`. It is further possible to use `AX_PTHREAD` to set `LIB` and `CFLAGS` etc., but I don't know how much it is important (besides for AIX, which may not be so important, so it may just be mentioned there as FIXME).
Note that `pthread` usage in `spellcheck-*.c` is already protected by `HAVE_*` that is not enabled on the MSVC build.

linas:
the tss_create stuff is a part of the C11 ISO standard -- is MSVC not C11 compliant?

ampli:
> the tss_create stuff is a part of the C11 ISO standard -- is MSVC not C11 compliant?

According to section 7.26 in the [C11 standard](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf), it is optional.

MSVC is mainly a C++ compiler, and its C11 compatibility is limited.
It doesn't support C11 threads (but it supports C++11 threads).

In any case, now that you use `AX_PTHREAD` you can just add `#if HAVE_PTHREAD`.
`

linas:
Merging.

ampli:
It seems there is some confusion regarding threads here.
This fix uses C11 threads, but checks for pthread, which is POSIX threads.
It just works on Linux because it has both.

I just tried to compile LG on OS-X (10.14, currently the latest release), and found that it has no `threads.h`. For some reason `clang` there also doesn't define `__STDC_NO_THREADS__` (like some other C compilers that don't have  `threads.h`). A similar problem may exist on MinGW and Cygwin, and in other compilers.

I also found that configure finds there (OS X) PTHREAD (so HAVE_PTHRAED==1), but this doesn't help in this regard.

I propose to check for the availability of the `threads.h` header file and then define `HAVE_THREADS_H=1`
I also propose to still check for PTHREAD and protect the pthread code in `spellcheck_*.c` with `HAVE_PTHREAD`.
If this looks fine I will send a PR.

linas:
Sigh. Sounds fine.

ampli:
Fixed in PR #1178.

PR #1152, opened 2021-03-11
bindings/java/Makefile.am: Fix comment start location

ampli:
Fix this error that happens when generation `configure`:
`bindings/java/Makefile.am:45: error: '#' comment at start of rule is unportable`

### DIFF
--- a/bindings/java/Makefile.am
+++ b/bindings/java/Makefile.am
@@ -42,7 +42,7 @@ endif
 install-data-hook:
 	${LN_SF} ${javadir}/${java_DATA} ${DESTDIR}${javadir}/linkgrammar.jar
 if HAVE_MVN
-	# Install linkgrammar into the local maven repository
+	: # Install linkgrammar into the local maven repository
 	mvn install:install-file \
 	  -Dfile=${java_DATA} \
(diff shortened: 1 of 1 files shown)


ampli:
It seems something has to be fixed with CI....

ampli:
I saw that your Makefile rule comment style uses `:` so I force-pushed a fix in that style.

linas:
Turning on CI was an accident arising from a merge of the javascript bindings. Perhaps it should be turned off again; .. or left on and fixed so it passes.  I kind-of don't care, as long as it doesn't get in the way.

>  your Makefile rule comment style

I have no clue. That's automated. I've never heard of a colon used as a makefile comment ... it doesn't even sound right, to me.

ampli:
> I have no clue. That's automated. I've never heard of a colon used as a makefile comment ... it doesn't even sound right, to me.

The makefile syntax prevents whitespace before rule comments. I first fixed that by starting the offending comment at column position 0. However, I then noted that another comment uses the `:` (empty shell command) trick in order to remain aligned with rule commands.

I also don't care much, but now it sends me a failure report on each PR...

PR #1153, opened 2021-03-11
Some commits from the `generate` branch

ampli:
Some commits from the `generate` branch are not dependent on the generation code and are a good idea anyway.
Applying them right away is a part of my effort to keep both branches as identical as possible.

### DIFF
--- a/data/en/4.0.dict
+++ b/data/en/4.0.dict
@@ -53,8 +53,8 @@
  % remain, below.
 
 % Capitalization handling (null effect for now- behave as empty words).
-1stCAP.zzz: ZZZ-;
-nonCAP.zzz: ZZZ-;
+<1stCAP>: ZZZ-;
+<nonCAP>: ZZZ-;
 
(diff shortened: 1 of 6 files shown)


linas:
> keep both branches as identical as possible.

I can periodically merge `master` into `generate`; that will make the differences clear.

The long-run plan is to merge `geneerate` into `master` .. that will happen whenever you're happy with the code.

ampli:
> I can periodically merge `master` into `generate`; that will make the differences clear.

As I said in the "Generation mode" discussion, I suspect it would create numerous extra merge commits when merged again to master. So maybe commits from master should be "squashed" instead.

Another thing that I propose there is that I will rebase my local `generate` branch on `master` whenever `master` get changed, and on completing a set of changes, **force-push** it to `opencog/link-grammar` `generate` branch. This supposes you will not make local changes at your local `generate` branch. This will allow for a much cleaner `generate` branch history.

PR #1157, opened 2021-03-14
Fix sql-backed dictionary crash

linas:
The SQL-backed dictionary never actually reads the entire dict.
It only fetches the needed parts on-demand.  Thus, new connectors
will become visible at any time. 

This can be triggered by starting the parser as `link-parser data/demo-sql` and then trying the sentences "this is a test" followed by "foo is a test". The "foo" will cause `<UNKNOWN-WORD>` to be found which then adds a new connector that hadn't been seen before.


ampli:
> then adds a new connector that hadn't been seen before

This problem was a result of my recent fix 7c252be180ca1910771883b41302ef7b1a395396 (SQL dict: Keep condesc memory over sentences). Previously the connector descriptor table got removed on sentence deletion (which causes havoc if a sentence gets parsed and then a previous sentence is deleted).

On a file dict,  adding connectors after dict read would not currently work fine, as some initialization will then be missing (with the DB dict they are explicitly for each sentence). So this `assert()` could be preserved if `&& !IS_DB_DICXT(dict)`. However, the dict handle is not available there (could be added to struct `ConDesc`).

PR #1161, opened 2021-03-15
Generate - WIP 2

ampli:
This PR fixes the word/category indication bug (still by a hack - FIXME).
It also adds a `link-generator` argument `--disjuncts`.

There is still a crash on macros, e.g. `This is a <noun-main-x>`. It happens in `string-set.c` when it adds `is.v`, but I still don't have any idea why.

### DIFF
--- a/link-grammar/prepare/build-disjuncts.c
+++ b/link-grammar/prepare/build-disjuncts.c
@@ -291,7 +291,9 @@ build_disjunct(Sentence sent, Clause * cl, const char * string,
 			*loc = n;         /* update the connector list */
 		}
 
-		if (!IS_GENERATION(sent->dict) || !isxdigit(string[0]))
+		/* XXX add_category() starts category strings by ' '.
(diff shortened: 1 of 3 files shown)


link-grammar/prepare/build-disjuncts.c
```diff
@@ -303,7 +305,7 @@ build_disjunct(Sentence sent, Clause * cl, const char * string,
 			ndis->category =
 				malloc(sizeof(ndis->category) * ndis->num_categories_alloced);
 			ndis->num_categories = 1;
-			sscanf(string, "%x", &ndis->category[0].num);
+			sscanf(string, " %x", &ndis->category[0].num);
```
linas, line 308: Using `ndis->category[0].num = strtol(string, NULL, 16);` here makes the code more readable (to me). 16 means base-16.

ampli:
You are right (it will be included in the next changes).

PR #1162, opened 2021-03-16
WIP: Fix failing GitHub Actions CI

zenflow:
After merging my branch (with emscripten port of link-parser) with the latest updates from master (PR #1150), the GitHub Actions CI workflow is failing.

This is responsible for the red "X" mark on GitHub next to every commit since the merge:
image

I triggered a CI workflow in my fork (via 84124a58f0b5226c88b74f9999178d744c01b130) to verify that my latest commit before the merge is still passing CI.
CI for that latest commit **is** passing, so something must have broken with the merging.

I pushed a commit fixing some relative file paths in my scripts which needed to be adjusted due to moving the directory from `js/` to `bindings/js/`. See d6a54e8a497bb4fe29b5222259b0272788a28738. That fixed the one fatal error but there is another, which is a bit beyond me as I'm not too knowledgeable about C/C++ or it's tools...

When compiling `link-parser` with `emcc` (emscripten):

```
error: undefined symbol: call_once
warning: To disable errors for undefined symbols use `-s ERROR_ON_UNDEFINED_SYMBOLS=0`
error: undefined symbol: tss_create
error: undefined symbol: tss_set
Error: Aborting compilation due to previous errors
shared:ERROR: '/home/runner/work/link-grammar/link-grammar/bindings/js/emsdk/node/14.15.5_64bit/bin/node /home/runner/work/link-grammar/link-grammar/bindings/js/emsdk/fastcomp/emscripten/src/compiler.js /tmp/tmpfv2spcnp.txt /home/runner/work/link-grammar/link-grammar/bindings/js/emsdk/fastcomp/emscripten/src/library_pthread_stub.js' failed (1)
Error: Process completed with exit code 1.
```

I tried the suggestion to use `ERROR_ON_UNDEFINED_SYMBOLS=0`. This got it to compile but when then when trying to actually use the CLI, it crashes:

```
$ The needs of the many outweigh the needs of the few. | link-parser
link-grammar: Error: Could not open dictionary "C.UTF/4.0.dict"
Debug: Dictionary "en/4.0.dict": Locale "en_US.UTF-8" unknown
link-grammar: Info: Dictionary 'en/4.0.dict': No locale definition - "C.UTF-8" will be used.
link-grammar: Info: Dictionary version 5.9.0, locale C.UTF-8
link-grammar: Info: Library version link-grammar-5.8.1. Enter "!help" for help.
missing function: call_once
linkparser> -1
-1
exception thrown: abort(-1). Build with -s ASSERTIONS=1 for more info.
Error: Process completed with exit code 1.
```

@linas Any idea what change is causing these "undefined symbol" errors or how to fix them?

I know that the change was introduced after the v5.8.1 release, since I actually also have the emscripten port in another separate project, [link-parser-js](https://github.com/zenflow/link-parser-js), which **pulls in** the link-grammar repo. I can easily try building it with various versions/releases/commits of the link-grammar repo. 
Here you can see the build is still working when I upgrade to link-grammar v5.8.1 but is broken (with the same "undefined symbol" errors) when I upgrade to the latest commit dcec4933e784368bde731907848ce09c81207458: https://github.com/zenflow/link-parser-js/branches
If you don't know what changes in master might have caused the build to break, I can try using `git bisect` to zero in and find those changes.


linas:
Thank you!

* I pulled your code in, with the idea of providing "one-stop shopping", to minimize the number of downloads, installs, configures that users have to make.  Well, that, and also to make it easier for users to find bindings.
* I put the code into `bindings/js` but if you have a better name, e.g. `bindings/emscripten` or something like that, we can change the name.

OK, so the `tss_create/call_once` thing: Those are some C library threading calls, first appearing in C11 (or maybe earlier, for gnu compilers, or never, for microsoft compilers.) They are provided by `libpthread.so` on Linux, so you would need to link to that to get definitions. To handle the build, the `configure.ac searches for these functions, it sets `#define HAVE_PTHREAD` if they are found, else not. The Makefile than adds `-lpthread` to the link libraries, if they are found. 

My guess is that changing `build.sh` to say 
```
emcc -O3 link-parser.bc ../../link-grammar/.libs/liblink-grammar.so -lpthread
```  
or maybe `libpthread.so` will fix the issue. Although I am kind of surprised ... the above is a brute-force fix, because it should not be needed. Normally, during compile, the correct location to `libpthread.so` is found and plugged in. You can view this by saying 
```
ldd link-grammar/.libs/liblink-grammar.so
```
and it will print the linked libraries, and where they are coming from.

linas:
(BTW link-parser itself is not threaded; it runs single-threaded. However, I have some other code that links to the libs and runs a bazillion threads doing god-knows-what, some of them parsing in parallel. Of course, a bug showed up in 5.8.1 and got fixed by the tss* stuff.)

linas:
FYI see also #1166 the best solution might be to just create a special single-threaded version for the js bindings (since the command-line tool by itself is not multi-threaded and does not need that support.)

ampli:
> the best solution might be to just create a special single-threaded version for the js bindings

Wouldn't the same problem exist with using the ndoe.js bindings?

In any case, for the JS link-parser it may be enough to define stubs for the offending functions in `link-parser.c`.

linas:
> Wouldn't the same problem exist with using the ndoe.js bindings?

I guess so...

>  it may be enough to define stubs 

in `link-parser.c`, yes I guess that would work. For the node.js bindings, I guess a similar trick would work, if we only knew where to put it ... assuming that node.js never runs multi-threaded itself; I don't that much about js and would be kind of surprised if it wasn't multi-threaded. 

We kind of need interested maintainers, or at least users for this stuff, as I'm not convinced that it's now our job to maintain these bindings. Of course, if you want to mess with it, I won't stop you. (But I think the generation and learning stuff is more important in the long run.)

ampli:
> users for this stuff,

For having users, maybe it should at least compile...

> For the node.js bindings

I now think it is better to disable `HAVE_PTHREAD`.
Options:
1. Do it if `CC == emcc`. This has a drawback in that it depends on a command naming.
Alternatively, check if CPPFLAGS contains __EMSCRIPTEN__.
2. Add an argument to `configure`:  `--without pthread`. This is a more general solution, as it can be used in other cases if needed.
3. Use (if it works - not checked yet):  `emcc -CPPFLAGS="-UHAVE_PTHREAD" ...`. It has to appear after the autoconf-generated preprocessor definitions.
4. In the LG library, use: `#if HAVE_PTHREAD && !__EMSCRIPTEN__`.

I am for (4), because it is trivial to implement. I will create a PR to test if it works fine.

PR #1166, opened 2021-03-18
Link parser.js fixes

ampli:
See #1162.
This is an experimental PR to see if CI is OK with it.

EDIT: **Please don't apply this PR yet.** (see below).

### DIFF
--- a/bindings/js/link-parser/build.sh
+++ b/bindings/js/link-parser/build.sh
@@ -16,7 +16,6 @@ emcc -O3 link-parser.bc ../../../link-grammar/.libs/liblink-grammar.so \
 	--pre-js pre.js \
 	-s WASM=1 \
 	-s ALLOW_MEMORY_GROWTH=1 \
-	-s ERROR_ON_UNDEFINED_SYMBOLS=0 \
     -o dist/link-parser.js
 rm link-parser.bc
 
--- a/configure.ac
+++ b/configure.ac
@@ -110,21 +110,28 @@ lg_tls=no
 AX_TLS(,:)
 test "$ac_cv_tls" != "none" && lg_tls=yes
 
-AX_PTHREAD
-if [ test -n "$ax_pthread_ok" ]; then
-	CC="${PTHREAD_CC:-CC}"
-	#CCX="${PTHREAD_CC:-CCX}"
-
-	dnl Currently, libtool uses --nostdlib when linking shared libraries,
-	dnl which causes the -pthread flag to be ignored. Try to work around that
-	dnl by explicitly specifying the pthread library unless it is already set.
-	if [ test -z "$PTHREAD_LIBS" ]; then
-		PTHREAD_LIBS=-lpthread
-	fi
+dnl Don't use thread functions with Emscripten.
+if test "$EMSCRIPTEN" != ''; then
+	AC_CHECK_DEFINE(__EMSCRIPTEN__, [emscripten=yes],
+						 [AC_MSG_ERROR([EMSCRIPTION environment variable found, but emcc is not used])])
 fi
+if test "$emscripten" != yes; then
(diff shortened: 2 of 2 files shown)


ampli:
The `emcc` failed with:
```
root:WARNING: USE_PTHREADS + ALLOW_MEMORY_GROWTH may run non-wasm code slowly, see https://github.com/WebAssembly/design/issues/1271
shared:ERROR: If pthreads and memory growth are enabled, WASM_MEM_MAX must be set
```

**Solution:**
Since link-parser doesn't need thread support at all, the use of `emcc` can be detected to disable the thread support altogether.
This can be done in `configure.ac` or in the library code (by checking the `__EMSCRIPTEN__` preprocessor symbol).
I will do one of these and will force-push a fix here.
**Please don't apply this PR yet.**

linas:
Yes @zenflow already tripped over this in #1162 and yes, you are right, the "best" solution is to just disable threading support in the js bindings. However, that presents a practical difficulty: we would need to compile a special sigle-threaded-only variant library for the js bindings.

ampli:
This time I got:
`config.status: error: cannot find input file: 'link-grammar/Makefile.in'`
The commit `Don't use pthread stuff for emscription` also modifies `link-grammar/Makefile.in` but this shouldn't  cause it to disappear. The rest of the changes only disabled the pthread functions.
I will need to compare the CI output with and w/o this PR.

In any case, I noted a strange problem while building on my computer:
`link-grammar/Makefile.in` has the following line:
`liblink_grammar_la_LIBADD  += ${PTHREAD_LIBS}`

If I modify it to:
`liblink_grammar_la_LIBADD  += XXX`
there is no build error. How come?

linas:
> here is no build error. How come?

recall that `make V=1` will generate verbose output. (or is it `make VERBOSE=1` ? I forget)

ampli:
> make V=1

Tried it and the XXX doesn't appear in the output. 

Also `-lpthread` never appears. Instead, `-pthread` is added by `AX_PTHREAD` (which is fine of course). 
From the `gcc` manual:
``` text
-pthread
  Define additional macros required for using the POSIX threads library.
  You should use this option consistently for both compilation and linking.
  This option is supported on GNU/Linux targets, most other Unix derivatives,
  and also on x86 Cygwin and MinGW targets.
```
Because of that, `emcc` also should be invoked with `-pthread` if we would like thread support. But instead, we should neutralize the thread support for the JS code due to its limitations (the way I did it in the library is not good enough and I intend to fix it).

linas:
> Tried it and the XXX doesn't appear in the output.

Did you remember to `autogen.sh` ? Since we turned off maintainer mode in master, this means that the makefiles are NOT regenerated when `Makefile.am` is changed!  This is perhaps annoying. Perhaps I should turn maintainer-mode back on in github, and only turn it off when making release tarballs.

... I think I will do that right now. I hope I don't forget to turn it off for the releases.

linas:
Done. I turned maintainer mode back on.

linas:
I also disabled CI workflows by renaming `.github/workflows` so you have to undo that if you want CI to run.  I got sick of the email spam it was generating.

ampli:
> ... I think I will do that right now. I hope I don't forget to turn it off for the releases.

Maybe a test for it can be added to `installcheck` or `discheck`.

ampli:
> Tried it and the XXX doesn't appear in the output.

I guess this was the problem... I will need to repeat my tests.

ampli:
I have just force-pushed a fixed patch.
It changes `configure` to not check for threads if an Enscripten build is detected.
This fixes the problem for the `link-parser` build.
I guess it fixes the problem also for the library build, but this build is bit-rotten on my computer due to the use of a newer `mpm` then used by the author.
So I think this patch can now be applied.

PR #1167, opened 2021-03-19
Generate WIP 3

ampli:
I got stuck in debugging memory leak due to not freeing disjunct categories in some cases on a sqlite3 dict.
But I have some work progress which is included here.
- Don't split macros in generation-mode.
- Get rid of `category_string` in `dict_category`.
- Disable spell-check for generation-mode in the **library** because it is not appropriate with generation.

BTW, I had a hard time creating this PR, because certain commits that are already in the repository got included again.
I guess I did something wrong but I have no clue what it could be.
I had to `pull -f` and apply my commits again.

EDIT: Force-pushed to fix a compile warning.

### DIFF
--- a/link-grammar/dict-common/dict-affix.h
+++ b/link-grammar/dict-common/dict-affix.h
@@ -84,5 +84,6 @@ typedef enum {
 
 Afdict_class * afdict_find(Dictionary, const char *, bool);
 bool is_stem(const char *);
+bool is_macro(const char *); /* Not an affix but best include file to use */
 
 #endif /* _LG_DICT_AFFIX_H_ */
--- a/link-grammar/dict-common/dict-common.c
+++ b/link-grammar/dict-common/dict-common.c
@@ -35,7 +35,7 @@
 #define STEM_MARK '='
 
 /* ======================================================================== */
-/* Affix type finding */
+/* Identifying dictionary word formats */
 
 /**
  * Return TRUE if the word seems to be in stem form.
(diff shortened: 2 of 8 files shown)


linas:
I don't quite understand the implications of the change with wall-words. Walls are important: in one experiment, the system learned a grammar that was unintentionally recursive, and could generate sentences of arbitrary length. This was because I forgot to put a period at the end of the sentences in the training corpus.

In another case, I complained to you earlier about there being multiple left-walls in a sentence. This was, in a certain sense,"correct", in that the grammar allowed it. It was also "correct" in that each extra wall really did note the start of a "new sentence", a clause inside the longer sentence. (The current English dict does this with CO+ connectors, but this hints that there are other possible designs.)  I haven't made any decisions as to whether multiple left-walls are good or bad, right now it's just "interesting" and not a problem. There is no need to force special treatment... yet.

ampli:
I can easily add a `link-generator` flag `--walls` to include the walls, so we can experiment with including them.

linas:
> I can easily add a `link-generator` flag `--walls` to include the walls, 

Well, oddly enough, it is already including the left wall, at least in the dictionaries I created. So the flag would need to be `--no-walls`

ampli:
> it is already including the left wall, at least in the dictionaries I created

This is because you added all the words, including walls and macros.
I didn't add walls and macros. While walls may be useful, I didn't find anything useful by allowing macros in wildcard words.
Adding `--walls` or `--no-walls` depends on the desired default. I will add `--no-walls` but we can change it if desired.

ampli:
Can right and left wall word names be subscripted?

ampli:
> Can right and left wall word names be subscripted?

In any case, since it is physically possible, I changed the code to be wall-subscript agnostic.

linas:
I just fixed a bug (in my code) where it was being subscripted.  The walls are complicated; they're a special case, requiring special handling. I expect to have more bugs.

PR #1176, opened 2021-03-31
Generate - WIP5

ampli:
PR #1175 clashed with my changes for printing disjunct costs, so I removed them (they are not important anyway, see my comment https://github.com/opencog/link-grammar/discussions/1146#discussioncomment-537636.

Main changes here:
- Generation mode: Implement --no-walls
- sqlite3 DB: Don't include macros in the Category array
- Generation mode: Implement finding unused disjuncts

There is still no API to get the unused disjunct info, since I'm waiting for input on my proposal at https://github.com/opencog/link-grammar/discussions/1146#discussioncomment-549964, which I repeat here for convenience:
``` c
typedef struct Disjunct_struct * Disjunct; // Opaque pointer
Disjunct * linkage_dictionary_unused_disjuncts(const Linkage);
const char * disjunct_expression(Disjunct *);
const int * disjunct_categories(Disjunct *); // NULL terminated array
```
Using the categories array returned by `disjunct_categories()`  it is possible to find the list of words (this seems more useful than directly providing the list of words).

### DIFF
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -182,6 +182,7 @@ struct Sentence_s
 	/* Generation mode - finding unused dictionary disjuncts.
 	 * This stuff is here and not in Dictionary because it depends on the
 	 * current dialect and cost. */
+	bool *disjunct_used;        /* Used dict disjuncts in memblock below. */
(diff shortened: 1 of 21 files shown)


linas:
Thanks! Continuing in #1146

PR #1181, opened 2021-04-06
Add Vala Bindings File

kmwallio:
Add Vala bindings file along with an example.


linas:
Thanks!  
* I notice that nothing is installed; surely that's not right? You want to copy some file to some location, somewhere. ...Edit the `Makefile.am` as needed ... If you're not used to those, let me know, they can be cryptic.
* Please add a note to `ChangeLog` in the root dir.

kmwallio:
> Edit the Makefile.am as needed  
> Please add a note to ChangeLog in the root dir.
Will do

PR #1183, opened 2021-04-10
Config

ampli:
Fix configuration stuff.
Main fixes:
- Fix Autoconf obsolete usage and stuff.
- Add sqlite3 library detection when it doesn't have a `.pc` file (like in macOS).
- Update regarding needed `configure` argument on current Mingw-w64.
- Fix flags assignment to LDFLAGS, CFLAGS, and CPPFLAGS.
- Remove redundant libraries in executable linking.

(Tested on all the supported systems but BSD ones. I intend to test it on BSD systems after I apply BSD patches that I have collected from BSD LG packages.) 

I also validated that adding `LDFLAGS"-Wl,--as-needed"` as `configure` arguments prevents overlinking and doesn't cause underlinking. However, I didn't add it automatically, as making a good Autoconf check for the correct working of "--as-needed" is complex, and anyway this idea is supposed to be known to package managers.


mingw/README-MinGW64.md
```diff
@@ -30,12 +33,13 @@ First install `mingw-w64-x86_64-toolchain`. Also install the rest of the
 prerequisite tools from the list in the main
 [README](/README.md#building-from-the-github-repository).
 
-NOTE: You must also install **mingw-w64-x86_64-pkg-config** .
+NOTE (may not needed with recent installations):
+You must also install **mingw-w64-x86_64-pkg-config** .
 
 You may find that the system is extremely slow. In that case, consult the
 Web for how to make tweaks that considerably speed it up. In addition, to
 avoid I/O trashing, don't use a too high `make` parallelism (maybe even
```
linas, line 41: thrashing, not trashing. I'll fix this.

PR #1189, opened 2021-04-21
`generate` branch bug fixes and cleanup

ampli:
- Fix crashes.
- Fix a possible memory leak.
- Generation mode: Disable count overflow notifications (no need anymore to set verbosity=0).
- Fix typos.

@linas, Please test it in `learn` to validate that I didn't damage any functionality with these fixes.
After that, it can be merged with `master`.

After that, I would like a performance tuning PR and a cleanup PR.


ampli:
I just added `configure` support to check for `argp`. If not found, a configuration warning is issued and `link-generator` is not getting compile. On BSD systems (including macOS) there is a need to install the `argp-standalone library. I added a note on that in `README.md`.

When I tried it on macOS I found that `limit.h` (needed for `INT_MAX`)  is missing (on Linux I guess that it is included via another include file).

PR #1191, opened 2021-04-23
Support >1G entries in Table_connector 

ampli:
Because `table_size` got enlarged to 64 bit, the table element size got increased.
The following commit reduces it to its original size:
`Table_connector: Reduce memory for the count field`

Set the maximum table size to 2^34 (16G).

With null_count=0, I think the bigger table size with the bundled batches is 2^30.
With null_count>0 it can be much larger than 2^30 (but I have no idea by how much).

This patch needs also to be applied to the `generate` branch by merging master into it (there is one easy conflict).

---
The problem with big table sizes (say over 2^24, the original limit), is that getting into page swapping more than negates the benefit of a bigger table. It is better to have a table load of 90% (or even more) than swapping.
I once proposed to use the `memory` parse option for a kind of soft memory limit. If not 0, the table size would not be enlarged over some minimal size if the program takes more memory than the `memory` option indicates. This can be easily implemented for systems that have `procfs` (Linux and *BSD). On systems w/o `procfs` it can just limit the table size - maybe
this can even be the implementations for all the systems - i.e. that `memory` would only control `Table_connector` size, since it is the main memory consumer and it can be adjusted.

### DIFF
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -285,11 +285,11 @@ static inline uint32_t string_hash(const char *s)
 /**
  * Hash function for the classic parser linkage memoization.
  */
-static inline unsigned int pair_hash(int lw, int rw,
-                                     int l_id, const int r_id,
-                                     unsigned int null_count)
+static inline size_t pair_hash(int lw, int rw,
+                               int l_id, const int r_id,
+                               unsigned int null_count)
 {
-	unsigned int i;
+	size_t i;
 
 #if 0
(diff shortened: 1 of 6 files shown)


ampli:
I forgot the case when `sizeof(size_t)==4`. In that case, the table size cannot reach more than 2^31, but there is not enough address space to support even that and it should be limited to 2^25.
I will send a fix.

ampli:
I have just pushed a commit for 32-bit systems.

PR #1194, opened 2021-04-25
Idiom dictionary loading speedup

ampli:
Use a better algo for inserting idioms into the dictionary.
The original algo inserted each word of an idiom with a numbered subscript and did hard work to find the number to use (by fetching all the previous such words and iterating them).

Instead, insert the idiom word into the dictionary (as `word.I`) only for the first time it is encountered. On subsequent identical idiom words that are encountered,  just OR their expression with the existing dictionary expression for that same idiom words.

When tested with `entities.b` (see issue #1192), the speedup is ~280x.  The memory saving is much more modest but still significant.

There are some more things to discuss on idioms., to be done here or in a new issue.

### DIFF
--- a/link-grammar/dict-common/idiom.c
+++ b/link-grammar/dict-common/idiom.c
@@ -112,16 +112,9 @@ static long max_postfix_found(Dict_node * d)
  *
  * Allocates string space and returns a pointer to it.
  * In this string is placed the idiomized name of the given string s.
- * This is the same as s, but with a postfix of ".Ix", where x is an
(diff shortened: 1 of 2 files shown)


linas:
The speedup is incredible. 

> There are some more things to discuss on idioms., to be done here or in a new issue.

As you wish.

ampli:
It is not possible to exactly say how much this implementation is faster in general since the original algo was > O(n^2)
The new one is maybe (I'm not sure) O(n log n).

PR #1196, opened 2021-04-26
General cleanup

ampli:
Main fixes: typos and code/comment rot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1380,13 +1380,13 @@ to the phrase, place a wall, and then step to the next part of the
 sentence.
 
 Caution: watch out for garden-path sentences:
-```text
+``` text
   The horse raced past the barn fell.
   The old man the boat.
   The cotton clothing is made of grows in Mississippi.
-  The current parser parses these perfectly; a viterbi parser could
-  trip on these.
 ```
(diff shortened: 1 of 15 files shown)


ampli:
Some obsolete parts of the main `README.md` need cleanup (or total removal) of these sections because they are obsolete now:

1. This has already been fixed.
`Fixes the word-count problem during spell-guessing. So, for example, if the mis-spelled word "dont" shows up in the input, it could be issued as one word ("done") or two ("do n't") and the current suffix-stripping/word-issuing algo cannot deal with this correctly. By contrast, this should not be an issue for the Viterbi algo, as it could explore both states at once.`
2. Implemented:
`Perhaps a simpler alternative would be to have labeled costs (a
cost vector), so that different dialects assign different costs to
various links.  A dialect would be specified during the parse,
thus causing the costs for that dialect to be employed during
parse ranking.`

Also, several URLs are now stale, and most of them are not found anymore on the Web (e.g. for word-grammar).

linas:
I removed paragraph 1. I added that 2. needs a practical tutorial.

I did not attempt to update URL's

Issue #1195, opened 2021-04-26
make check fails on Linux Mint 20.1 - gcc 9.3.0


plankatron:
Making check in link-grammar
make[1]: Entering directory '/home/fleeky/link-grammar-5.9.0/link-grammar'
make[2]: Entering directory '/home/fleeky/link-grammar-5.9.0/link-grammar'
make[2]: Nothing to be done for 'check-am'.
make[2]: Leaving directory '/home/fleeky/link-grammar-5.9.0/link-grammar'
make[1]: Leaving directory '/home/fleeky/link-grammar-5.9.0/link-grammar'
Making check in bindings
make[1]: Entering directory '/home/fleeky/link-grammar-5.9.0/bindings'
Making check in ocaml
make[2]: Entering directory '/home/fleeky/link-grammar-5.9.0/bindings/ocaml'
make[2]: Nothing to be done for 'check'.
make[2]: Leaving directory '/home/fleeky/link-grammar-5.9.0/bindings/ocaml'
make[2]: Entering directory '/home/fleeky/link-grammar-5.9.0/bindings'
make[2]: Nothing to be done for 'check-am'.
make[2]: Leaving directory '/home/fleeky/link-grammar-5.9.0/bindings'
make[1]: Leaving directory '/home/fleeky/link-grammar-5.9.0/bindings'
Making check in link-parser
make[1]: Entering directory '/home/fleeky/link-grammar-5.9.0/link-parser'
make[1]: Nothing to be done for 'check'.
make[1]: Leaving directory '/home/fleeky/link-grammar-5.9.0/link-parser'
Making check in tests
make[1]: Entering directory '/home/fleeky/link-grammar-5.9.0/tests'
make  dict-reopen multi-dict multi-thread mem-leak 
make[2]: Entering directory '/home/fleeky/link-grammar-5.9.0/tests'
  CXX      dict-reopen.o
  CXXLD    dict-reopen
/usr/bin/ld: /home/fleeky/link-grammar-5.9.0/link-grammar/.libs/liblink-grammar.so: undefined reference to `call_once'
/usr/bin/ld: /home/fleeky/link-grammar-5.9.0/link-grammar/.libs/liblink-grammar.so: undefined reference to `tss_create'
/usr/bin/ld: /home/fleeky/link-grammar-5.9.0/link-grammar/.libs/liblink-grammar.so: undefined reference to `tss_set'
collect2: error: ld returned 1 exit status
[...]

PR #1197, opened 2021-04-26
Use $PTHREAD_CFLAGS) on the relevant source file compilation and the library linking

ampli:
Regarding issue #1195, these are my proposed changes.

They don't set the global CFLAGS, but instead, use $(PTHREAD_CFLAGS) on the `tests` sources that use pthread, on all the library sources, and on the library linking.

I validated it by inspecting `make V=1` and ensuring that `-pthread` appears on these source file compilation and the library linking. It doesn't appear in the `link-parser` sources compilation and the python wrapper compilation (because it is not set in the global `CFLAGS` as I removed this line from `configure`).

Please test.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -111,8 +111,6 @@ AX_TLS(,:)
 test "$ac_cv_tls" != "none" && lg_tls=yes
 
 AX_PTHREAD
-CFLAGS="${CFLAGS} ${PTHREAD_CFLAGS}"
-LIBS="${LIBS} ${PTHREAD_LIBS}"
 
 dnl Check if we can use C11 threads functions
 AC_CHECK_HEADERS_ONCE([threads.h])
--- a/link-grammar/Makefile.am
+++ b/link-grammar/Makefile.am
@@ -34,7 +34,7 @@ DEFS = @DEFS@ -DVERSION=\"@VERSION@\" -DDICTIONARY_DIR=\"$(pkgdatadir_realpath)\
 
 # $(top_builddir) to pick up autogened link-grammar/link-features.h
 AM_CPPFLAGS = -I$(top_builddir) -I$(top_srcdir)
-AM_CFLAGS = $(WARN_CFLAGS) $(HUNSPELL_CFLAGS) $(REGEX_CFLAGS) $(SQLITE3_CFLAGS)
+AM_CFLAGS = $(WARN_CFLAGS) $(HUNSPELL_CFLAGS) $(REGEX_CFLAGS) $(SQLITE3_CFLAGS) $(PTHREAD_CFLAGS)
 
 post-process/pp_lexer.lo: AM_CPPFLAGS += -I$(top_srcdir)/link-grammar/post-process
 MAINTAINERCLEANFILES = $(top_srcdir)/link-grammar/post-process/pp_lexer.c
@@ -43,7 +43,7 @@ lib_LTLIBRARIES = liblink-grammar.la
 
 liblink_grammar_la_LDFLAGS = -version-info @VERSION_INFO@ \
 	-export-dynamic -no-undefined \
-	$(LINK_CFLAGS)
+	$(PTHREAD_CFLAGS)
 
 liblink_grammar_la_LIBADD  =  ${REGEX_LIBS}
 
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -16,9 +16,9 @@ check_PROGRAMS = dict-reopen multi-dict multi-thread mem-leak
 
 if HAVE_JAVA
 check_PROGRAMS += multi-java
-AM_CPPFLAGS += $(JAVA_CPPFLAGS)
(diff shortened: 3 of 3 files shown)


linas (approved):
Looks right to me. I'll test shortly and push version 5.9.1 in a few hours.

linas:
This is insufficient to fix issue #1195  -- My testing procedure is this:
```
sudo apt remove libsqlite3-dev libaspell-dev
./autogen.sh --no-configure
cd build
rm -r link-parser link-grammar tests
../configure --disable-python-bindings --disable-java-bindings --disable-editline
make -j
ldd -r ./link-grammar/.libs/liblink-grammar.so.5.9.0
```
I still get
```
undefined symbol: tss_set	(./link-grammar/.libs/liblink-grammar.so.5.9.0)
undefined symbol: tss_create	(./link-grammar/.libs/liblink-grammar.so.5.9.0)
undefined symbol: call_once	(./link-grammar/.libs/liblink-grammar.so.5.9.0)
```
and `make check` fails.

ampli:
I will do it exactly as you did and will investigate what is missing.

linas:
This fixes it:
```
--- a/link-grammar/Makefile.am
+++ b/link-grammar/Makefile.am
@@ -74,6 +74,9 @@ endif WITH_SAT_SOLVER
 # Needed for tss_create and tss_set
 liblink_grammar_la_LIBADD  += ${PTHREAD_LIBS}
 
+# Brute-force fix for bug #1195
+liblink_grammar_la_LIBADD  += -lpthread
+
 # Math libraries are needed for floorf, etc.
 liblink_grammar_la_LIBADD  += -lm
```

I spent a few hours screwing around with `AX_PTHREAD` and other suggestive things in `/usr/share/aclocal` and had no luck. It appears that `${PTHREAD_LIBS}` is always empty, It also seems that linking with the `-pthread` flag does nothing discernible; it does not magically find the needed library.

I could just push the above brute-force fix, and issue 5.6.1 and cross my fingers that it doesn't wreck Apple or BSD or Windows....

linas:
https://stackoverflow.com/questions/2127797/significance-of-pthread-flag-when-compiling

linas:
Uh, hang on .. I failed to `git pull` your fixes ...  That's ... unexpected so ignore comments above, let me try again.

linas:
... I now pulled your fixes .. and it still doesn't work. I'm going to bed.

linas:
I can confirm the stackoverflow suggestion above does NOT end up using `-lpthread`. Look at `gcc -dumpspecs | less` and back up by one line to see what it applies to.

ampli:
This is what I found:

 The `gcc` manual says:
`       -pthread
           Link with the POSIX threads library.  This option is supported on GNU/Linux targets, most other Unix derivatives, and also on x86 Cygwin and MinGW targets.  On some targets this option also sets flags for the preprocessor, so it should be
           used consistently for both compilation and linking.`
So it looks like that adding `-pthred` should cause it to link with `-lpthread`.

However, the link command line contains -pthread (even twice...) and it definitely doesn't link with `-lpthread` (I tried with a single `-lpthread` just to be sure).

The link command is:
`g++ -fPIC -DPIC -shared -nostdlib ... -pthread -Wl,-soname -Wl,liblink-grammar.so.5 -o .libs/liblink-grammar.so.5.9.0`
When I tried to remove the `-nostdlib` argument (plush removing the `crtX.o` stuff) the LG library got linked with the pthread library!

From googling `libtool pthread`,  it turned out this is an old and common problem with apparently no good solution, which may be considered as a bug in `gcc` (or by others as a bug in `libtool`).
See for example https://gcc.gnu.org/bugzilla/show_bug.cgi?id=25460:
`Because -nostdlib currently has priority over -pthread, libtool can't build shared C++ libraries using libpthread. A libtool maintainer told me to file a bug with GCC. Here's my original email: ...`
(He talks about c++, but I validated that this happens also when using `gcc` instead of `g++`.)

From reading various items in the said google search results, it looks like people are solving this problem by adding `-lpthread` when linking (`-pthread` is still needed when compiling to get `gcc` pthread support). But maybe there is a better solution that I haven't located.

In a hope of better portability, I propose this solution:
1. Use `PTHREAD_LIBS` on linking (it should have been used anyway). On Linux it is empty.
2. In `configure.ac`, if `PTHREAD_LIBS` is empty, assign to it `-lpthread`.

I can send a PR that implements it (of course after I test it.) But only later today.

BTW, what I still didn't find is why the linking of `link-parser` gets the `-pthread` flag (which causes overlinking).

linas:
> I propose this solution:

Yes, that is the right solution.

FYI, the future is in flux: 
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55394   -- removes dependency on pthread, just 6 months ago ...
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99341 -- above is reverted  2 months ago.

This is why I think your proposed fix is the right one.

linas:
One pill makes you larger, and the other pill makes you small: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66146

PR #1198, opened 2021-04-28
Work around a problem with "-nostdlib -pthread"

ampli:
Currently, `libtool `uses `--nostdlib` when linking shared libraries, which causes the `-pthread` flag to be ignored. Try to work around that by explicitly specifying the pthread library unless it is already set.

Tested with:
`configure --disable-java-bindings --disable-python-bindings --disable-editline --disable-aspell --disable-hunspell --with-regexlib=c`
Also, the sqlite3 library detection had been commented out in `configure.ac` for this test.
The dependecies were inspected with:
1. `ldd -v ...`
2. `lddtree -a ...` (from the `pax-utils` package)
BTW, `--with-regexlib=c` was needed because else it used `pcre` which is linked with the `pthread` library.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -111,6 +111,18 @@ AX_TLS(,:)
 test "$ac_cv_tls" != "none" && lg_tls=yes
 
 AX_PTHREAD
+if [ test -n "$ax_pthread_ok" ]; then
+	CC="${PTHREAD_CC:-CC}"
+	#CCX="${PTHREAD_CC:-CCX}"
+
+	dnl Currently, libtool uses --nostdlib when linking shared libraries,
+	dnl which causes the -pthread flag to be ignored. Try to work around that
+	dnl by explicitly specifying the pthread library unless it is already set.
+	if [ test -z "$PTHREAD_LIBS" ]; then
+		PTHREAD_LIBS=-lpthread
+		LIBS="$PTHREAD_LIBS $LIBS"
+	fi
+fi
 
 dnl Check if we can use C11 threads functions
 AC_CHECK_HEADERS_ONCE([threads.h])
@@ -1010,7 +1022,7 @@ $PACKAGE-$VERSION  build configuration settings
 
 	prefix:                         ${prefix}
 	datadir:                        $(eval "echo ${datadir}")
-	C compiler:                     ${CC} ${CPPFLAGS} ${CFLAGS}
+	C compiler:                     ${CC} ${CPPFLAGS} ${CFLAGS} ${PTHREAD_CFLAGS}
 	C++ compiler:                   ${CXX} ${CPPFLAGS} ${CXXFLAGS}
 	Regex library:                  ${REGEX_LIBS}
 	TLS:                            ${lg_tls}


configure.ac
```diff
@@ -111,6 +111,18 @@ AX_TLS(,:)
...
+if [ test -n "$ax_pthread_ok" ]; then
+	CC="${PTHREAD_CC:-CC}"
+	#CCX="${PTHREAD_CC:-CCX}"
+
+	dnl Currently, libtool uses --nostdlib when linking shared libraries,
+	dnl which causes the -pthread flag to be ignored. Try to work around that
+	dnl by explicitly specifying the pthread library unless it is already set.
+	if [ test -z "$PTHREAD_LIBS" ]; then
+		PTHREAD_LIBS=-lpthread
+		LIBS="$PTHREAD_LIBS $LIBS"
```
linas, line 123: I'll merge this, but then remove line 123 because, I believe, only liblinkgrammar needs to be linked this way. This would match an earlier change you'd make

linas:
OK, I merged, removed that line. Can you test again? I'd like to push out version 5.9.1 shortly.  (Time zone differences means, I expect you to be sleeping right now, but what do I know?)

ampli:
It looks fine. Here it is after this patch (using `lddtree liblink-grammar.so`):
``` text
liblink-grammar.so => /tmp/link-grammar/master/link-grammar/.libs/liblink-grammar.so (interpreter => none)
    libprofiler.so.0 => /lib64/libprofiler.so.0
        libunwind.so.8 => /lib64/libunwind.so.8
    libpcre2-8.so.0 => /lib64/libpcre2-8.so.0
    libhunspell-1.7.so.0 => /lib64/libhunspell-1.7.so.0
        libtinfo.so.6 => /lib64/libtinfo.so.6
    libsqlite3.so.0 => /lib64/libsqlite3.so.0
        libdl.so.2 => /lib64/libdl.so.2
        libz.so.1 => /lib64/libz.so.1
    libpthread.so.0 => /lib64/libpthread.so.0
    libstdc++.so.6 => /lib64/libstdc++.so.6
    libm.so.6 => /lib64/libm.so.6
    libc.so.6 => /lib64/libc.so.6
    ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2
    libgcc_s.so.1 => /lib64/libgcc_s.so.1
```
Here it is in 5.9.0:
``` text
liblink-grammar.so => /tmp/link-grammar/test2/link-grammar/.libs/liblink-grammar.so (interpreter => none)
    libprofiler.so.0 => /lib64/libprofiler.so.0
        libunwind.so.8 => /lib64/libunwind.so.8
    libpcre2-8.so.0 => /lib64/libpcre2-8.so.0
        libpthread.so.0 => /lib64/libpthread.so.0
    libhunspell-1.7.so.0 => /lib64/libhunspell-1.7.so.0
        libtinfo.so.6 => /lib64/libtinfo.so.6
    libsqlite3.so.0 => /lib64/libsqlite3.so.0
        libdl.so.2 => /lib64/libdl.so.2
        libz.so.1 => /lib64/libz.so.1
    libstdc++.so.6 => /lib64/libstdc++.so.6
    libm.so.6 => /lib64/libm.so.6
    libc.so.6 => /lib64/libc.so.6
    ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2
    libgcc_s.so.1 => /lib64/libgcc_s.so.1
```

The `LIBS="$PTHREAD_LIBS $LIBS"` is indeed redundant. I included it for "consistency", because this is what `AX_PTHREAD` itself does when needed (however, in that case `liblink_grammar_la_LIBADD  += ${PTHREAD_LIBS}` is not needed).

Some Comments:
1. Somehow `link-parser` gets linked with `libpthread` anyway.
This happens even if I configure with `LDFLAGS=-Wl,--as-needed` to prevent overlinking  (this flag removes most of the libraries but not `libpthread`).
2.
>  * Fix build break when SQLite3 is not installed. #1195

For the build break to occur also libpcere2-8 has to be missing.
3. `-lm` is also not needed for the library/programs.

linas:
The `-lm` was needed either by earlier versions of `resources.c` or maybe debug printing, or maybe some of the performance measurement code. I don't recall.

ampli:
It was needed due to the use of `fabs()`  (removed in 35386ceebbbda79e5e99317d27c3a174ae282269) and `fmax()` (removed in 9d6c2e99daba6f6505faf3d57d13bd22389f6248), and maybe some other floating-point functions on costs (that are not used now).

Issue #1200, opened 2021-04-28
Idioms


ampli:
In PR #1194 I said:
> There are some more things to discuss on idioms., to be done here or in a new issue.

Currently, a word with underbars is considered to be an idiom. Because I once wanted to enable debugging of idioms using `!!word*`, I modified the dictionary reading to also add the idiom dictionary strings as they are.

I add code to provide a way to define a non-idiom word with underbars, so for example `word1_word2` would get recognized only in this form (and not with whitespace instead of underbars). The idea was to allow to backslash-escape underbars (in the dict) so they would not be recognized as idiom word delimiters. But this code is incomplete, as it doesn't remove the backslashes... so it is not useful.

I have a fix for that (~20 LOCs).  It can load a very big list of words with escaped underbars in a fraction of time of loading the same words as idioms. This is not very important,  the method of backslash-escaping has limitations, and the idiom code needs a revision anyway, but I would like to send a PR for that anyway.

ampli:
Another problem that is related to idioms is their dict notation:
Connectors that start with ID are reserved for idioms. This causes a problem for automatic dict generation in which you need to skip using these letters when generating connector names. In addition, the subscript `.I` is reserved for idioms (previously only when numbered, but now only when unnumbered...).

To solve both of these problems and yet another one, I propose to use an initial underbar. The idea is to allow it in a connector base name (in addition to uppercase letters). The connectors for idiom words would then start with an underbar, e.g. `_ID...`
or maybe better `_I...`. As with variable symbols, an initial underbar will be reserved for "system use". A similar solution can be applied for idiom word subscripts, i.e. to use `._I`.

This proposal would solve two other things:
1. In language learning dictionaries, when there is a need to compose connector names, there is no natural character to serve as a separator. An underbar can serve as such a separator.
2. In the `vn` dict an underbar was used in the connector base names as a prefix separator. In order to make this dict readable, I have replaced it with the letter `U`, a thing that made them somewhat unreadable. With this proposal, the original underbars could be restored.

The only drawback of this proposal (that I can think of) is that it is not backward compatible in case someone uses a program that analyzes link names (and search for `ID`) or would try to understand old code using new link-grammar documents (unless they mention the old conventions).

ampli:
> The idea was to allow to backslash-escape underbars (in the dict) so they would not be recognized as idiom word delimiters

The dict syntax already contains a way to escape special-meaning characters: enclosing them in double quotes.
[...]

PR #1201, opened 2021-05-01
Make `_` valid in connector names 

ampli:
Add underbar as a valid character in the connector uppercase part. See issue #1200.

It turns out it is not possible to restore the original underbars in `vn` dictionary. This is because its author used a Java implementation of the link-grammar library in which underbars and numbers belong to the connector name (a.k.a the uppercase part). BTW, I wondered why this dict has only 5 sentence examples and none of them can be fully parsed. I will write on that in a separate issue.

Main fixes here:
- Accept also '_' in the connector uppercase part
- connectors.h: Add functions to check for valid connector string
-  data/vn/4.0.dict: Fix connector strings with mixed digits & uppercase
   (An old patch that validates connector subscripts made them invalid but the problem in this file was neglected.)
- Fix line numbers in the main dictionary file
  (I noticed it again in the error on invalid connectors so I fixed it at last.)
- Improve dict error messages.

---
I would like now to fix the idiom connector-name/subscript conventions to include an initial underbar:
- `_Ixxxx` for idionm links.
- `_I` for idiom subscripts.

Edit: Force-pushed to correct a bad inclusion of a test `en/4.0.dict`.

### DIFF
--- a/data/vn/4.0.dict
+++ b/data/vn/4.0.dict
@@ -12,10 +12,11 @@
 % DT = danh từ = noun
 % TVT=trước động từ=preceeding verb
 
-% Due clashes with the conventions of the library, the following
-% link characters are replaced:
+% Due to clashes with the conventions of the library, the following
+% connector characters are replaced:
+% 2_ is replaced by 2u.
 % _ is replaced by U.
-% Đ is replaced by V
+% Đ is replaced by V.
 
 % Danh từ
 bàn ghế giường áo quần sách quyển_sách báo nhà cửa
--- a/data/vn/4.0.dict
+++ b/data/vn/4.0.dict
@@ -12,11 +12,11 @@
 % DT = danh từ = noun
 % TVT=trước động từ=preceeding verb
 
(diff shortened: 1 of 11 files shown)


linas:
thanks!

linas:
Beats me. I looked at the second sentence: "tôi mua một bông hoa" and two of the four words are not in the dictionary. Given that the dictionary is of a reasonable size...

linas:
The vietnamese dict came from here: https://www.researchgate.net/publication/287444370_Parsing_complex_-_compound_sentences_with_an_extension_of_Vietnamese_link_parser_combined_with_discourse_segmenter

Maybe the authors can be contacted.

linas:
Well, if you're uncomfortable with asking, I could write them.  I think I did write them many years ago, they never responded. Can't hurt to try again. I believe that this was a masters and/or PhD thesis project, one to write a parser in java, another to create the dict. I think the students moved on, are doing something else, and lost interest. That seems to be the most probable status.

PR #1203, opened 2021-05-02
Remove optional detection of bare + subscripted word

ampli:
There was once a documented rule that it is forbidden to have in the dict both a bare word and the same word subscripted. I am not sure why, but I think that in the old times there was no function to get a bare word in such a case (the code still has inefficiencies due to that, which I would like to fix next).

In any case, this check had a bug and some words didn't obey it. I once added an optional check (at verbosity 10 and 103) to list such words and a possibility to exempt selected words (the initial words that didn't obey this rule) from the list via a dictionary directive `%suppress DUP-BASE`.

It seems there is no reason now for this rule (I couldn't find its documentation, maybe it has been removed), as it doesn't seem harmful to have this word combination (BTW, for some reason it is a rare one).

So for the sake of clearing code bloating, this check is removed here.
On the same occasion, I cleaned up `insert_dict()` by moving the optional check for duplicate idioms to a separate function, and documented the historical reason for this check.

### DIFF
--- a/data/en/4.0.dict
+++ b/data/en/4.0.dict
@@ -1447,7 +1447,6 @@ its my.p your their.p our thy.p yisser.p yousser ye'r:
 
 % Possessive version of me
 % Cost on D, DD: avoids use as determiner on "Make me coffee"
-%suppress: DUP-BASE (for me.p)
 me.p:
   DP+
   or [{AL-} & {@L+} & (D+ or DD+)];
@@ -2426,7 +2425,6 @@ ninety-seventh.ord ninety-eighth.ord ninety-ninth.ord
 % - the strength was in the order of gerE > cotD > yfhP P2 > yfhP P1
 % also remember "-->"
 
-%suppress: DUP-BASE (for a.eq)
 A.eq B.eq C.eq D.eq E.eq F.eq G.eq H.eq I.eq J.eq K.eq L.eq M.eq
 N.eq O.eq P.eq Q.eq R.eq S.eq T.eq U.eq V.eq W.eq X.eq Y.eq Z.eq
 a.eq b.eq c.eq d.eq e.eq f.eq g.eq h.eq i.eq j.eq k.eq l.eq m.eq
@@ -12848,7 +12846,6 @@ maybe.r:
 % Argumentatives (children gain-saying).
 not.intj is_too is_not is_so unh_unh: Wa-;
 
-%suppress: DUP-BASE (for seriously.ij)
 % Openers to directives, commands (Xc+ & EI+ connection to infinitives)
 % or single-word interjections, exclamations.
 % These are semantically important, so they've got to parse!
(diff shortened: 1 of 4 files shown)


linas:
I merged this because it seems reasonable, for now. 

The duplicate-word detection is a historical debugging artifact: when editing `4.0.dict`, or the word-lists, it is very easy to accidentally repeat a word twice, and thus, have two conflicting definitions for it. ... and then have some confusing parses with hard-to-find bugs. I think this is why subscripts were invented: to allow a word to appear twice, because "we really want to have it appear more than once".   For maintaining `4.0.dict` for English, this is a very useful feature, and I rely heavily on it, e.g to detect words that accidentally show up in both `words.n.4` and `words.n.10` when just one of these was desired.

However, for some of the autogenerated dictionaries, it would be nice to completely disable this checking, so that a word can be repeated many times, with the understanding that each repetition should be an "or" of the definitions.

linas:
Hmm. I just tried this patch. Getting rid of `%suppress: DUP-BASE` is a good idea. I don't know how that got in there; the dictionary should have been fixed.

However, the loss of duplicate definition detection is harmful for the English dict. Although I don't often do maintenance on it, I do rely quite heavily on duplicate word detection to make sure I did not make mistakes. ... it needs to be turned back on.

My test was adding 
```
foo: A+;
foo: W- & S+;
```
and there are no complaints. ("foo was a foo book")

ampli:
What I intended to remove was **only** the test of both an unsubscripted word and a subscripted one (the same base word), since this test was already disabled (the original test was buggy and thus never really enforced).

So my patch was buggy. Twice the same word should still be checked.
I will send a fix.

linas:
I guess I am saying that it would be nice to enable/disable this on a per-dictionary basis. ... but this is low priority.

ampli:
I found my bug. It was in a previous patch that I sent and you you have just applied - PR #1201.
>I guess I am saying that it would be nice to enable/disable this on a per-dictionary basis. ... but this is low priority.

Because I'm going to fix it anyway, it is very easy to make it on a per-dictionary basis.
E.g.:
`#define allow-duplicate-words false`

BTW, the error message which is used is:
`Ignoring word which has been multiply defined:`
Should it be "multiplied defined"?
Or maybe "previously defined "?

> Getting rid of `%suppress: DUP-BASE` is a good idea. I don't know how that got in there;

When I disabled the buggy test that disallowed the same subscripted and unsubscripted word, there were already 6 words that evaded it (due to their definition order). So I wanted only new ones to be notified. But since this check was optional, now there are many such words. And there is no problem with that, so I just removed this unneeded check.

ampli:
> found my bug. It was in a previous patch that I sent and you

I was wrong, it is in this PR (#1203). By mistake, I used `&&` instead of the original `||`.
In any case, before I submit a fix I have to know

linas:
"multiply defined" is OK. "previously defined" is also OK. So is "defined previously". LG fails to parse all three variants :-)

PR #1204, opened 2021-05-04
Changes for dictionary duplicate word detection

ampli:
1. Fix the bug introduced in PR #1203 (unintentional duplicate word check).
2. Allow duplicate words with `#define allow-duplicate-words true`.

Only the word `true` is recognized. Anything else is considered as `false`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,6 @@
 Version 5.9.2 (XXX 2021)
  * Expanded English vocabulary
+ * Support dictionary "#define allow-duplicate-words true".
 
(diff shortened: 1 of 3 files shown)


ampli:
There is currently a limitation when a word appears multiple times in the dict: It shouldn't be used at the RHS of a rule, since currently only one of the occurrences will be then used. This can be fixed (by ORing the expressions) or a warning can be issued.
But I guess this is not important since (I guess) autogenerated dictionaries don't use RHS words.

Issue #1205, opened 2021-05-04
`prune_lookup_list()` (and also `abridged_lookup_list()`)


ampli:
[`prune_lookup_list()`](https://github.com/opencog/link-grammar/blob/5ded3151ee33f880fdd12fa1d27705c539dad473/link-grammar/dict-file/read-dict.c#L573-L608) uses [`dict_match()`](https://github.com/opencog/link-grammar/blob/5ded3151ee33f880fdd12fa1d27705c539dad473/link-grammar/dict-file/read-dict.c#L548-L564) in order to remove any Dict_node that doesn't match the bare word of the search string  or is an exact match.
However, the existence of such nodes looks impossible, as [`rdictionary()`]( https://github.com/opencog/link-grammar/blob/5ded3151ee33f880fdd12fa1d27705c539dad473/link-grammar/dict-file/read-dict.c#L631-L662) create a `Dict_node` element only for search strings with matching bare words or a whole word (when `dict_order()` returns 0).

`dict_order()` can be one of 3 functions:
1. [`dict_order_bare()`](https://github.com/opencog/link-grammar/blob/5ded3151ee33f880fdd12fa1d27705c539dad473/link-grammar/dict-file/read-dict.c#L486-L507)
2.  [`dict_order_wild()`](https://github.com/opencog/link-grammar/blob/5ded3151ee33f880fdd12fa1d27705c539dad473/link-grammar/dict-file/read-dict.c#L532-L544)
3. [`dict_order_strict()`](https://github.com/opencog/link-grammar/blob/5ded3151ee33f880fdd12fa1d27705c539dad473/link-grammar/dict-file/read-dict.c#L455-L473) (in the ifdef'ed-out `strict_lookup_list()`)

All of these functions only return 0 if the string matches a bare word or the whole word.

So the question is why `prune_lookup_list()` is needed at all.
If it is indeed not needed I would like to remove it.

ampli:
When [`make_connector()`](https://github.com/opencog/link-grammar/blob/5ded3151ee33f880fdd12fa1d27705c539dad473/link-grammar/dict-file/read-dict.c#L956-L1035) finds a word at the RHS of a dict definition, it looks it up in the dict.
[...]

PR #1207, opened 2021-05-05
Dict lookup cleanup and speedup

ampli:
(See issue #1205.)

- Simplify and clean up.
- Make faster.

The dict lookup time for tokenization is now significantly faster. However, it is a negligible part of the total parsing time.
As a result, the speedup for `en/corpus-basic.batch` is only about 1%, and for the other corpus batches, it is unnoticeable.

### DIFF
--- a/link-grammar/dict-file/read-dict.c
+++ b/link-grammar/dict-file/read-dict.c
@@ -733,6 +733,7 @@ Dict_node * file_lookup_wild(Dictionary dict, const char *s)
 	return result;
 }
 
+#if 0
 /**
  * abridged_lookup_list() - return lookup list of words in the dictionary
  *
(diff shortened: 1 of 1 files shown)


link-grammar/dict-file/read-dict.c
```diff
@@ -631,10 +635,9 @@ static Dict_node *
 rdictionary_lookup(Dict_node * restrict llist,
                    Dict_node * restrict dn,
                    const char * restrict s,
-                   bool match_idiom,
+						 bool boolean_lookup,
```
linas, line 638: indentation!

ampli:
> indentation!

 To be fixed in the next cleanup.

PR #1208, opened 2021-05-06
Change idiom construction conventions

ampli:
Use `_IX` for idiom links.
Use `_I` for idiom subscripts.
Remove the reservation of `IDX` connectors and `.In`s subscripts.
Reserve subscripts that start with `_` (connectors that start with `_` are already reserved, PR #1201).

---
Comments:
1. The new connector names for idioms may be incompatible for people who pedantically analyze link and connector names in the results.
2. There is a need to change the library documentation in the places it mentions the `ID` and `.I` reservation, and instead add the initial `_` reservation.

A reference for the "Idioms" issue: #1200.

A ChangeLog suggestion:
```
 * Support underbars in connector names (the first position is reserved).
 * Connector names starting with "ID" are not reserved now.
 * ".I" subscripts are not reserved now; "._" subscripts are now reserved.
```

### DIFF
--- a/link-grammar/dict-common/idiom.c
+++ b/link-grammar/dict-common/idiom.c
@@ -78,8 +78,9 @@ static const char * build_idiom_word_name(Dictionary dict, const char * s)
 
 	size_t n = lg_strlcpy(buff, s, sizeof(buff));
 	buff[n] = SUBSCRIPT_MARK;
-	buff[n + 1] = 'I';
-	buff[n + 2] = '\0';
+	buff[n + 1] = '_';
+	buff[n + 2] = 'I';
+	buff[n + 3] = '\0';
 
 	return string_set_add(buff, dict->string_set);
 }
@@ -146,9 +147,9 @@ static const char * generate_id_connector(Dictionary dict)
 	/* i is now the number of characters of current_name to skip */
 	t = buff;
 
-	/* All idiom connector names start with the two letters "ID" */
+	/* All idiom connector names start with the two characters "_D" */
+	*t++ = '_';
 	*t++ = 'I';
-	*t++ = 'D';
 	for (; i < IDIOM_LINK_SZ; i++)
 	{
 		*t++ = dict->current_idiom[i];
@@ -314,6 +315,6 @@ bool is_idiom_word(const char * s)
 	const char *sm = strchr(s, SUBSCRIPT_MARK);
 
 	if (NULL == sm) return false;
-	if ((sm[1] == 'I') && (sm[2] == '\0')) return true;
+	if ((sm[1] == '_') && (sm[2] == 'I') && (sm[3] == '\0')) return true;
 	return false;
 }
--- a/link-grammar/dict-file/read-dict.c
+++ b/link-grammar/dict-file/read-dict.c
@@ -100,8 +100,8 @@
   used.
 
   The dictionary writer is not allowed to use connectors that begin in
-  "ID".  This is reserved for the connectors automatically
-  generated for idioms.
+  "_".  This is reserved for connectors that are automatically
+  generated (currently only for idioms).
 
   Dictionary words may be followed by a dot (period, "."), and a "subscript"
   identifying the word type. The subscript may be one or more letters or
@@ -109,6 +109,9 @@
   (mostly?) subscripts consisting of a single letter, and these serve mostly
   to identify the part-of-speech. In general, subscripts can also be used
   to distinguish different word senses.
+
+  Subscripts that start with "_" are reserved for words that are
+  automatically generated (currently only for idioms).
 */
 
 static bool link_advance(Dictionary dict);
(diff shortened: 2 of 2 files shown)


linas:
thanks

linas:
I just changed the website to call these links `_I`.  https://www.abisource.com/projects/link-grammar/dict/section-_I.html

I did not find anything that talks about `.I` as a subscript: section 4.8 of https://www.abisource.com/projects/link-grammar/dict/introduction.html#4  (scroll down) just says:  "Idioms cannot be given subscripts; if "a_la_mode.a" is included in the dictionary, this will not be accepted."  I did not test this, I assume that still holds.  So ... where are the `.I` subscripts used?

ampli:
- `_IX` - My intention was that thee X is a placeholder for capital letters.
- `.In` reservation - I also cannot find it. I guess it is not documented after all (besides in the source code).
- "Idioms cannot be given subscripts..." From some time ago they can. This is needed if you would like an idiom to appear in the RHS of a rule, but it is defined more than once in the dict. It is also needed if you would like to have a correction to an idiom with a mistake. I found one example to the last one:
`rather_then.#rather_than: [rather_than]bad-spelling;`
``rather_then.#rather_than:` is actually a subscripted idiom.

Regarding `rather_then.#rather_than`:
Interestingly, the word `then` is inserted as `then.#rather_than._I` (with a double subscript) and is handled correctly.
But just now I found a minor bug in the `!!` command, which may actually be considered a feature:
``` text
linkparser> !!then._I
Token "then._I" matches:
    then.#rather_than._I           24  disjuncts

    then._I                          1  disjuncts
```
(The `then._I` is from `then_on`.)

ampli:
@linas,
In the announcement of 5.10.0 you wrote:
> Version 5.10.0 of link-grammar is now available for download. The minor version number was bumped due to a quasi-incompatible change in the handling of idioms. They now use an underscore as the first character of the link type, and underscores are now reserved.

Actually, this patch adds the ability to use underscore in the connector name (aka "uppercase part"), in addition to `[A-Z]`.
This is useful e.g. when you would like to programmatically construct a dict, and have connector names composed from 2 or more names while making clear their name boundaries.

https://github.com/opencog/link-grammar/blob/e3fe4eaf61fce2e5ff47065b142cbbb93f826f19/ChangeLog#L11-L22

It turns out I forgot to add before line 16:
`* Add underbar to the valid characters in a connector name (aka the "uppercase part").`

**I think we should add it retroactively**...

linas:
> ability to use underscore in the connector name (aka "uppercase part"), in addition to `[A-Z]`.

Oh! I did not realize that! 

> I think we should add it retroactively...

To the ChangeLog? Sure. Most of the work will be in updating the documentation to state this clearly. And I still got to move the docs out of svn and into git, or at least, export them once, for safe-keeping. Ugh. Not enough hours in the day.

BTW, If I recall correctly, the demo Vietnamese dict used to have underscores  in it; I think I had to strip them out, but maybe they can be put back?

linas:
Yeah, here: https://github.com/opencog/link-grammar/blob/e3fe4eaf61fce2e5ff47065b142cbbb93f826f19/data/vn/4.0.dict#L24-L27

Want to make that change, or should I?

linas:
Uh, no:
```
link-grammar: Error: While parsing dictionary "vn/4.0.dict":
Invalid character in connector subscript (only lowercase letters, digits, and "*" are allowed): "_"
	 Line 96, next tokens: "TST2_1-" "or" "TST2_2-" "}" "&" 
link-grammar: Fatal error: Unable to open dictionary.
```
See branch `vn-underscore`

ampli:
Note that they used a strange connector scheme, in which numbers were a part of the connector name (not part of the connector subscript - their original dict doesn't have subscripts). But in our library, since `_` is part of the connector name, it cannot be a part of the subscript. So `TST2_1` is invalid.

ampli:
> So TST2_1 is invalid.

This was the reason I didn't convert it back, since the names should be modified anyway (e.g. by changing `2` to `B`).

linas:
Ugh.  Should we allow numbers in link names?

linas:
Anyway, most of `vn/corpus-basic.batch` is badly broken, I never figured out why, because the dict itself seems quite large and complex.

The original authors never responded to emails.  Also, thier project consisted of rewriting LG in java, so maybe they handled link names in some manner that we don't.

ampli:
> Ugh. Should we allow numbers in link names?

They are allowed in link names - in subscripts.
We shouldn't allow them in the "uppercase part" since they are already used in subscripts in several dicts (`ru`, `tr`, `fa`, `kz` and `he`), and this is also done in a meaningful way.

ampli:
> is badly broken

Indeed there is no point to touch it unless we find why it is broken. Maybe we can find this out from the document (even though it is not of this version of the dict - it uses other connector names).

In any case, maybe we should add a comment that it was written using the authors' version of LG library and it is currently broken for an unknown reason.

PR #1210, opened 2021-05-13
sqlite3 dict: Fix unintended generation mode by default

ampli:
This bug makes it impossible to parse with nulls when using an SQLite dict (because I checked the inverse condition).
It exists in 5.9.1. Also, in 5.9.1 generation mode using a sqlite3 dict gets a segfault.
However,  I think there is no need to hurry to issue a new version due to the rarity of sqlite3 dict usage.
Before issuing a new version, I would like to apply two speedup fixes, add the expression resolving API call, and make a code/comment cleanup.

In any case, after 5.9.1 got issued, one of my fixes broke generation mode. It will be fixed in the next PR (after I find out the problem...).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 Version 5.9.2 (XXX 2021)
  * Expanded English vocabulary
  * Support dictionary "#define allow-duplicate-words true".
+ * Fix parsing with nulls when using an sqlite3 dictionary.
 
 Version 5.9.1 (28 April 2021)
  * Fix build break when SQLite3 is not installed. #1195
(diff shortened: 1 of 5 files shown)


ampli:
I pushed an addition fix:
- read-sql.c: Fix memory leak in dict->category
- 
The problem happened when I created `dictionary_generation_request()` and move to it the memory allocation of `dict->category`. But `read-sql.c` also allocated memory for it. So I moved this memory allocation back to `dictionary.c`.

There is an additional problem that needs fixing: No sentences are generated using `demo-sql`.
`link-generator lg -L -l demo-sql --verbosity=200` looks fine, and I think it once worked.

I still investigate the problem, but this PR can be applied anyway.
Note that the generation mode is still broken due to another bug, to be fixed in the next PR.

linas:
thanks!

PR #1211, opened 2021-05-13
Fix sentence generation (for file dict)

ampli:
Fix yet another logical conditions bug...

This time I added a test that sentence generation can get linkages.

---
EDIT:
~Generation from demo-sql still doesn't work, and I don't know why.
It didn't work when the `generate` branch got merged.
I'm still checking for the problem.~
There was actually no problem. It just cannot generate sentences in the default length (5). Using `-s 4` works.


ampli:
I added a generation test using `demo-sql`.

PR #1212, opened 2021-05-13
A slight speedup for very short sentences

ampli:
For very short sentences (<=10 words) the `lrcnt` cache is not used.
So don't allocate/free it too.

### DIFF
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -1318,7 +1318,7 @@ int do_parse(Sentence sent, fast_matcher_t *mchxt, count_context_t *ctxt,
 /* sent_length is used only as a hint for the hash table size ... */
 count_context_t * alloc_count_context(Sentence sent, Tracon_sharing *ts)
 {
-	count_context_t *ctxt = (count_context_t *) xalloc (sizeof(count_context_t));
+	count_context_t *ctxt = malloc (sizeof(count_context_t));
 	memset(ctxt, 0, sizeof(count_context_t));
 
 	ctxt->sent = sent;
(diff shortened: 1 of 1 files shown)


linas:
FYI Related to speedups: I have this 2-year-old unapplied patch to use hugepages for `kept_table`. If I recall correctly, it might have improved performance by 1% (??) but I decided, at the time, I didn't want to pursue it due to complexity and maintenance headaches. But if you are interested, that old patch is here: https://github.com/linas/link-grammar/commit/e0ebf0a585f8271b5ef6416d56f1d56b75fadb98

linas:
... and as I think about it, it might be the case that using hugepages for everything might offer a 1% or 2% speedup.   As I now recall, this does not actually require any changes to the C code ... I think that just linking to `libhugetlb.so` is enough to turn it on.  I didn't do this because, on that particular day, I did not feel like messing with configure.ac and/ore Makefile.am to make this happen, nor did I feel like doing the testing. But it is another way of squeezing out a few more percent of performance ...

ampli:
> libhugetlb.so

This library exists in binary form only for very few systems (see https://pkgs.org).
Especially, it is not available for Fedora (my development system). However, I was able to install it from Centos 8 (in the hope it is compatible).

But when I wanted to modify LG  to use it, I encountered a slight problem:
https://github.com/opencog/link-grammar/blob/606414cc6a923a608124faafa149499f96e4d157/link-grammar/parse/count.c#L129-L145
Line 136 needs to be replaced by `unmap()` with the table size as a second argument.
However, it is not available in this function (maybe `ptr_to_table` can point to a struct that includes the table pointer and the table size, but it is not clear to me how to do that).

PR #1213, opened 2021-05-13
Code, comment ,and spelling cleanup

ampli:
Mainly code, comment, and spelling cleanup (besides some comment changes, detailed below).
- Fix indentation.
- Fix typos (according to http://sid.ethz.ch/debian/report/codespell/link-grammar-5.8.0.html).
  I skipped some contributed files.
- Add FIXME.
- Fix comment rot.

The following changes are more than cleanup:
- read-dict.c: Add dialect component description
From it and from `!help dialect`, an update to the library documentation may be constructed.
- read-dict.c: Remove "including exponent" from cost description.
Costs with exponents are not currently used in the dicts, and I would
like to replace the cost reading function with a simpler one that
doesn't look at the locale and doesn't support an exponent.
- Update the comment on -lm.
  I think it is not needed, but I cannot easily check that because it is included in `libstdc++` which I have not checked yet how to remove when not compiled with C++. `libm` is also included in `libsqlite3` (and `libstdc++` is also included in `libhunspell` and `libaspell`).


linas:
Wow!   I don't want to know how you found those typos, do I?

Issue #1217, opened 2021-05-26
Dictionary definitions for user API


ampli:
`dictionary_lookup_list()` and `dictionary_lookup_wild()` are part of the API, and words that they return, if subscripted, have `SUBSCRIPT_MARK` in them. However, this symbol is currently not part of the API.
Similarly, returned parses may include `LEFT-WORD` and `RIGHT-WORD`, which may need to be handled by the user.
These symbols are defined in `dict-defines.h`, which is not part of the user API.

I encountered this problem when I added the rest of the API functions (those that are in `dict-api.h` and `dict-structures.h`)  to the Python bindings. When I wanted to test `dictionary_lookup_list()` I found that `SUBSCRIPT_MARK` needs to be hard-coded.
I already had such a problem in `link-generator`:
https://github.com/opencog/link-grammar/blob/5cb9e68e553b6ba39f17b021cdd13b6c6105a414/link-parser/generator-utilities.c#L13-L15

Possible solutions: 
1. Continue to hard-code these constants.
2. Provide a public `.h` file with definitions for constants that are used by the API. The file `dict-api.h` is not appropriate unless we would eventually like to use another include file for non-dictionary constants. It also contains irrelevant (to the API) definitions.
3. Put the constants in the public API file that contains functions that use them, with a priority to `link-includes.h`.
4. Implement a new API call `linkgrammar_get_library_definition(const char *)`, when the argument may be strings like `left-wall`, 'max-word`, etc. But I think this is less desirable than (3).

I propose to implement (3). It is already implemented in `link-includes.h` for the `lg_error_severity` and `ConstituentDisplayStyle` enums, and also for `LG_PANIC_DISJUNCT_COST`. Similarly, `LEFT_WALL_WORD` and `RIGHT_WALL_WORD` can be moved to it. The `SUBSCRIPT_*` definitions can be moved to `dict-api.h`. It is not clear to me what to do with constants like `MAX_WORD`.

However, if (3) is implemented, there will be a need to include `link-includes.h` in files that need e.g. `LEFT_WORD

linas:
why not just make `dict-common/dict-defines.h` part of the public API? That is, just install it, and let users use it.

ampli:
> why not just make `dict-common/dict-defines.h` part of the public API? That is, just install it, and let users use it.

It currently also contains internal-only definitions:
``` c
static const double UNINITIALIZED_MAX_DISJUNCT_COST = -10000.0;
[...]

PR #1220, opened 2021-05-26
Make `dict-defines.h` a public API

ampli:
See issue #1217.

`dict-defines.h`:
- Move internal definitions to `dict-common.h`.
- Move `WILDCARD_WORD` from `tok-structures.h`.
- Perform cleanups.
- Add it to the public include files.
- Use it in `link-generator` and remove the private library-constants definitions.
- Add it to the SWIG interface (Fix #include style and use WILDCARD_WORD in `tests.py` instead of a hard-coded string.

On the same occasion:
- In the `link-generator` sources, use `#include <...>` instead of `#include "..."`.
  (I don't know why I previously encountered a problem in that.)
- In `constituents.h`, remove a duplicate `#include` of `dict-defines.h`.

(I have not exposed `MAX_SENTENCE` for now.)

### DIFF
--- a/bindings/python-examples/tests.py
+++ b/bindings/python-examples/tests.py
@@ -981,12 +981,12 @@ class YGenerationTestCase(unittest.TestCase):
     """
     def test_getting_linkages_file_dict(self):
         ParseOptions(test='generate')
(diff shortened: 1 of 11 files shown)


linas:
Looks good!

PR #1223, opened 2021-05-31
regex-morph.c: Use REG_GNU if defined

ampli:
This patch is for NetBSD.


linas:
Do you want a ChangeLog entry for this? Something like "fix for regexes in BSD" or something like that?

ampli:
I propose something like:
Fix regexes for NetBSD when using libc regexes.

Issue #1222, opened 2021-05-31
`lg_exp_stringify()` (and also `lg_exp_*()`)


ampli:
https://github.com/opencog/link-grammar/blob/bdb90344f8474157037c9bf3f6b5df570aecaefd/link-grammar/dict-common/dict-structures.h#L98-L107

The `lg_exp_stringify()` returns a pointer to a static TLS memory. I did it this way because it is used internally in many places and it was inconvenient to use a temporary variable + `free()`. However, this is not consistent with the rest of the API functions, which don't change the memory pointed by existing pointers on subsequent calls.

So I propose to change it to return newly allocated memory that needs to be freed by the API user. (For internal use, I would define a new function `exp_stringify()` returning a static TLS memory, and the API user can do the same if desired.)

There is also a question of how to free the result. The rest of the API uses special memory-freeing functions that usually just call `free()`. I guess this is for future flexibility in case a more complex freeing would be needed. But `free()` seems to always be fine because future added allocations could be done in one contiguous block. In case a special memory-freeing function is desired, I couldn't think of a good name for it... (`lg_exp_free_stringify()` is not good...).

A similar API-design problem exists with the new `lg_exp_resolve()` (which returns an expression in one memory block),
and the existing experimental functions `sentence_unused_disjuncts()` and `disjunct_expression()` (`free()` can be used for all of them).

While writing this I noted that the `lg_exp_operand*()` functions don't use `const` on their parameter and returned result. I cannot think of a reason for that, and maybe this is unintentional. I propose to fix that.

To sum up:
1. Proposal: Change `lg_exp_stringify()` to return new memory.
2. Question: Should we define special memory-freeing functions for all the API functions (listed above) that return new memory that has to be freed?
3. Proposal: Use `const` for `lg_exp_operand*()` parameter and result.

linas:
I merged #1224 without thinking very much about it.  Perhaps its just late at night, and I don't want to think about it :-/ 

const for the static inline is fine; since they are inlined, any imagined future changes won't be binary compatible.  I won't sweat bullets because this is not glibc.

grepping through my code base, I see no external users of stringify -- I thought there was one, but I don't see it.

PR #1224, opened 2021-06-05
`lg_exp_stringify()` fixes

ampli:
Bug fix:
- `print_expression_tag_start()`: Fix printing extra `[` when not printing tags
- 
API change:
- `lg_exp_stringify()`: Return a new memory
- `lg_exp_operand_*()`: Use const for parameter and result

The API changes are according to my proposals in issue #1222. I didn't implement a special `free()`.

`clinkgrammar.lg_exp_stringify()` now leaks memory in Python. This is to be fixed in the next PR (needs a rebase on `master` after applying this PR).

Force-pushed:
-  Change `lg_exp_stringify()` to return `char *` instead of `const char *`, as usual for functions that don't own the resulting memory.
- Change  `exp_stringify()` (the function for internal use, that uses TLS memory for its result) to call an internal function that `lg_exp_stringify()` calls (instead of calling `lg_exp_stringify()`, which has an external linkage).

### DIFF
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -541,7 +541,7 @@ void sentence_delete(Sentence sent)
 	free_linkages(sent);
 	post_process_free(sent->postprocessor);
 	post_process_free(sent->constituent_pp);
-	lg_exp_stringify(NULL);
+	exp_stringify(NULL);
 	free(sent->disjunct_used);
 
 	global_rand_state = sent->rand_state;
--- a/link-grammar/dict-common/dict-structures.h
+++ b/link-grammar/dict-common/dict-structures.h
@@ -105,7 +105,7 @@ static inline const Exp* lg_exp_operand_next(const Exp* exp)
                                             { return exp->operand_next; }
 link_public_api(const char *)
 	lg_exp_get_string(const Exp*);
-link_public_api(const char *)
+link_public_api(char *)
 	lg_exp_stringify(const Exp *); /* To be freed by the caller. */
 
 /**
--- a/link-grammar/dict-common/dict-structures.h
+++ b/link-grammar/dict-common/dict-structures.h
@@ -99,12 +99,14 @@ static inline Exp_type lg_exp_get_type(const Exp* exp) { return exp->type; }
 static inline char lg_exp_get_dir(const Exp* exp) { return exp->dir; }
 static inline bool lg_exp_get_multi(const Exp* exp) { return exp->multi; }
 static inline double lg_exp_get_cost(const Exp* exp) { return exp->cost; }
-static inline Exp* lg_exp_operand_first(Exp* exp) { return exp->operand_first; }
-static inline Exp* lg_exp_operand_next(Exp* exp) { return exp->operand_next; }
+static inline const Exp* lg_exp_operand_first(const Exp* exp)
+                                             { return exp->operand_first; }
+static inline const Exp* lg_exp_operand_next(const Exp* exp)
+                                            { return exp->operand_next; }
 link_public_api(const char *)
 	lg_exp_get_string(const Exp*);
 link_public_api(const char *)
-	lg_exp_stringify(const Exp *); /* Result is in static variable. */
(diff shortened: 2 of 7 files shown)


linas:
I don;'t think you need to rebase for a second PR.  That is, the following works fine:
```
git checkout master
git checkout -b branch-a
hack
git commit
git push origin branch-a
# go to github and create pull req for branch-a
git checkout -b branch-more   # This is a continuation of the work on branch-a, e.g. the python work.
hack
git commit
git push origin branch-b
# goto github and create pull req for branch-b. This will look "just fine".
# oops found an error in branch-a
git checkout branch-a
hack
git commit
git push origin branch-a # this will auto-update the pull req. No force-push required.
# the above won;'t change the pull-req for branch-b, but everything will be fine
# after a and b are merged. 
```
You can bounce around between these branches, and more, adding more commits all the time, and it will all work fine, without any need at all for any rebases or any force-pushes,  And if, for some reason, you need to have the latest fix from `branch-a` before you can commit something to branch be, well, that's easy, too:
```
git checkout branch-b
git merge branch-a
hack
git commit
git push
``` 
Still no rebase or force-push required, and everything will look just fine in the github pull reqs; they won't get garbled.

ampli:
> I don't think you need to rebase for a second PR
> [...]
> Still no rebase or force-push required

Some comments:

When I fix a code that is not yet merged to `master`, I just continue my regular workflow of removing the bad code altogether by rewriting the commits. The forced-push was needed because I didn't add a commit but changed an existing one. But in that case, I forgot to release beforehand... (see the `fixup!`).

> git merge branch-a

This is similar to a rebase, just leaving a merge record. 


Regarding working on feature branches, my workflow is totally different:

I use worktrees (now using `git worktree` but previous to the worktree integration into `git` I used an external script for that). This means these branches are always checked out (I rarely use `git checkout` for branches). I currently have over **100**(!) such branches, which I consider as "active development". I keep the inactive branches (several hundred) using `git tag -a` for "future reference").

The worktree branches get out of date every time I send a PR... After a PR is applied, I return to the branches from which I intend to prepare the next PRs and rebase them so I can continue to work on an up-to-date code. I could rebase in advance but I would like to be sure I do it with an actual repository code.

There is an additional workflow I regularly use (almost for each PR):

I break branches into several PRs and/or cherry-pick commits for the candidate PR. Then I rebase the original branches so I don't send commits that I already sent (GitHub is not clever when this happens). For example, now I have a branch named `exp-resolve` that I created for adding the `lg_exp_resolve()` API call. But I wanted to make a test for it and to that end, I needed to bind an API call that returns an expression. So I decided to bind `dictionary_lookup_list()`. But I realize I can just bind all the rest of the API at the same time. And then I needed to test the API calls that I need for testing `lg_exp_resolve()`... and meanwhile I found and fixed bugs in the code that I touched.

So I finally got a branch that implements several different things, among them binding the rest of the API calls and implementing a new API `lg_exp_resolve()` with its tests. I already extracted out of it commits that implement fixes/enhancements and sent them separately. Now I intend to break it (by cherry-picks) into a pure `lg_exp_resolve()` implementation and a branch to add the rest of the API.

**Of course, I would appreciate much any idea how to manage working on >100 branches and keep them up to date with less frequent rebasing (or merging).**

(BTW, in old branches that I try to revive ( rebasing becomes too complicated. For example, I have a big difficulty in reviving my expression-simplification branch. I got a buggy result, and it is not easy to debug.)

linas:
OK, well, you've got a sophisticated system going, then. 

I have lots of old, stale branches, but effectively all of them are abandoned, and will probably never be revived. Whatever idea I was pursuing in them, it became clear it would not work. Thus, I don't ever need to keep them working and up-to-date.

I explained my workflow mostly because you said that you won't push your next request until the last one is merged. For me, that is never a problem: I can create many pull reqs, each depending on the earlier ones, and they all work out just fine. I think I've done a rebase once in my life. Merges more often, but the need is infrequent.

PR #1229, opened 2021-06-08
Re-enable GitHub Actions CI

zenflow:
Reverts "yet another attempt to disable testing" f0d854ef356a32ed2ea43748df805a6136c7a37c.

After #1166, the emscripten js build seems to be working, so we could re-enable the "bindings-js.yml" CI workflow to help keep it that way.

You can see the results of the workflow in my fork of the repo: https://github.com/zenflow/link-grammar/runs/2769757309
I think the workflow should also run in this repo once I open this PR.

image

---

Regarding the [Travis CI workflow](https://github.com/opencog/link-grammar/blob/136afc5eb3e1df13303c6b23e08ea0364ea80db2/.travis-xxx-disable-failing.yml) that was also disabled:
That workflow checks the native build (`make && make check && sudo make install && sudo ldconfig && make installcheck`), which the "bindings-js.yml" workflow doesn't do.
I think it would be convenient to have both workflows running on a single CI service.
I feel like GitHub Actions is the better choice (over Travis CI) because it's built right into the platform we're already using, and it leaves little to be desired.
If you want I will be happy to convert the Travis CI workflow into a GitHub Actions workflow.

/cc @linas @ampli


linas:
Thanks!  I have no strong feelings about travis or github actions or other things of this sort, so I'm evaluating this as a request "by popular demand".

Issue #1236, opened 2021-06-21
Bug: verbosity>0 & batch mode


linas:
Something changed recently: when running `cat corpus-basic.batch | link-parser > logfile` the current code will print `No complete linkages found.` while the old code did not. 

In `link-parser.c` line 972 I see the print statement. One easy fix would be to move that line down a few lines. The other is to make sure that `verbosity==0` for batch mode. @ampli this is your specialty...

ampli:
Maybe the bug is due to this change: d237dd73b3fcdf18836daed2e8e3dd3ee389bed8
I will look at it.

ampli:
> One easy fix would be to move that line down a few lines.

This was the fix I used.

PR #1239, opened 2021-06-25
link-parser.c: Don't print "No complete linkages found" in batch mode

ampli:
See issue #1236.
This bug got introduced in the panic mode timeout fixes.


linas:
thanks

Issue #1247, opened 2021-09-07
python test fails on little-endian 32bit archs: realloc(): invalid next size


jonassmedegaard:
Building link-grammar 5.10.0 on Debian fails on architectures armel, armhf, i386, mipsel, hurd-i386, kfreebsd-i386, and x32 - i.e. all little-endian 32-bit architectures if I am not mistaken.
Overview here: https://buildd.debian.org/status/package.php?p=link-grammar (but please ignore kfreebsd-amd64 failing for a different Debian-specific reason).
Excerpt from build log of armhf build:
```
Running by: /usr/bin/python3
Running ./tests.py in: /<<PKGBUILDDIR>>/bindings/python-examples
PYTHONPATH=./../python:../python:../python/.libs
srcdir=.
LINK_GRAMMAR_DATA=./../../data
Compiled with: gcc __VERSION__="10.3.0"  
OS: linux-gnueabihf __unix__ 
Standards: __STDC_VERSION__=201112L _POSIX_C_SOURCE=200809L 
Configuration (source code):
	CPPFLAGS=-Wdate-time -D_FORTIFY_SOURCE=2
	CFLAGS=-D_DEFAULT_SOURCE -std=c11 -D_BSD_SOURCE -D_SVID_SOURCE -D_GNU_SOURCE -D_ISOC11_SOURCE -g -ffile-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -O3 -fvisibility=hidden
Configuration (features):
	DICTIONARY_DIR=/usr/share/link-grammar
[...]

PR #1250, opened 2021-09-16
A additional fix for issue  #1247 ++

ampli:
- Fix bad initial size of the "category" Disjunct field. Now it seems trivial but I had to install a vm for **armv7l** and use gdb...
- On the same occasion: Fix non-portable union field order in Disjunct_struct (not a problem for this issue).


linas:
> Fix bad initial size of the "category" 

Of course! That should have obvious!

PR #1257, opened 2021-11-13
Cleanup++

ampli:
This PR contains typo fixes.
However, while checking for typos (this time using the `codespell` command), I found the following sentence in the `corpus-fixes` batch that is not getting parsed:
`He tried but failed to acheive his goal.`
From its context, I guess this is not an intended typo. So I fixed it.
But since it demonstrates a problem in the current tokenization algo (issue #404) I also added two examples to `corpus-fixes`.
I added a FIXME in the commit because I think I know how to fix it with negligible overhead (no reparsing).


linas:
I've been wondering how to do `codespell` for a long time!

PR #1258, opened 2021-11-13
Support using C++ regex

ampli:
It can be configured using `--with-regexlib=c++`. It is not the default because it is currently significantly slower than the other regex libraries.
Benchmark:
en/corpus-basic.batch +2.7%
en/corpus-fixes.batch  +2.2%
ru/corpus-basic.batch  +51% (!)

Note that if  `--with-regexlib=c++` is used then the `regex-morph.c` compilation is done using the C compiler.
This is done to support using CFLAGS that discard unused libraries from the linkage, in that case, the C++ library (such flags can be added in principle to `configure.ac` when it finds that they are supported). Configuring with `--with-regexlib=c++` triggers the use of the C++ compiler on `regex-morph.c` (only).

I rewrote the `regex-morph.c` code because it is currently unreadable due to the numerous `#ifdef`s, and adding the C++ regex support made that worse. As always, this has the drawback of replacing a well-tested code with a new one.

For MSVC C++ regex is set as default because it is painful to add an external POSIX regex-compatible library.  On `en` it is not significantly slower. However, it is 17% slower on `ru`. But users that need speed should use it on Linux anyway (it is faster there in any case).

Other regex packages, implement start-up optimizations that support fast failure before invoking the regex engine.
For example, if the last character before `$` mismatches, the subject string surely mismatches. Implementing a few such shortcuts may significantly speed up the use of C++ regexp in LG, especially for `ru` (due to  8648 `MORPH-END` regexes).


link-grammar/dict-common/regex-morph.c
```diff
@@ -48,84 +54,199 @@ typedef struct {
...
 			{
-				prt_regerror("Failed to compile regex", rn, rc ,erroffset);
 				rn->re = NULL;
-				return rc;
+				return false;
 			}
 
 			/* Check that the regex name is defined in the dictionary. */
-			if ((NULL != dict) && !dict_has_word(dict, rn->name))
+			if ((dict != NULL) && !dict_has_word(dict, rn->name))
```
linas, line 240: The reason for writing `NULL != dict` instead of `dict != NULL` is to help avoid "undetectable single-byte errors". In this case, if you accidentally delete the `!` the first will not compile; the second will compile but produce incorrect results. 

These kinds of errors are very rare an unusual, but I did watch one guy spend a week debugging something with a single-byte error.  Then again, he was a bit of a, umm. ahh. Actually, I think he was stoned all the time.

linas:
This was the right way to refactor this.

Issue #1050, opened 2019-11-17
Sentence: bird flu was observed in which countries?


ampli:
In an unrelated search I encountered [page 358 of "Intelligent Information and Database Systems: 8th Asian Conference ..., Part 2"](https://books.google.co.il/books?id=DtCzCwAAQBAJ&pg=PA358&lpg=PA358&dq=%22link-grammar%22+graph&source=bl&ots=aJRAMVvpoX&sig=ACfU3U0DuHIY7CZPdi18Gkvi2HKcmhdJzQ&hl=en&sa=X&ved=2ahUKEwipi9fc8u_lAhWfRxUIHZOmAV04ChDoATADegQIBxAB#v=onepage&q=%22link-grammar%22%20graph&f=false).

This conference was in 2016, but according to their benchmark time it seems they used the original CMU version (a common thing), but the problem is the same:
``` test
    +--------------------------------Xp--------------------------------+
    +--------------->WV--------------->+                               |
    +------>Wd------+                  |                               |
    |        +--AN--+---Ss--+----Pvf---+---MVp--+-------Jp------+      |
    |        |      |       |          |        |               |      |
LEFT-WALL bird.n flu.n-u was.v-d observed.v-d in.r [which] countries.n ?
```
(They turned out using another parser.)

On the other hand, `in which countries?` does parse:
```
Found 1 linkage (1 had no P.P. violations)
	Unique linkage, cost vector = (UNUSED=0 DIS= 2.00 LEN=4)

    +-------------Xp-------------+
    |       +------Jp-----+      |
    +-->Wj--+-JQ-+---Dmc--+      |
    |       |    |        |      |
LEFT-WALL in.r which countries.n ?
```
Here `in.r` uses its disjunct `Wj- & JQ+ & J+` to attach to `which countries`, so as a test I tried adding the disjunct `MVp- & JQ+ & J+`.

It didn't work and the question is why.
Fixing this as needed may also be interesting.

linas:
```
Get a more detailed help on a variable as in "!help var".
linkparser> !bad
Display of bad linkages turned on.
linkparser> bird flu was observed in which countries ?
Found 2 linkages (0 had no P.P. violations)
	Linkage 1 (bad), cost vector = (UNUSED=0 DIS= 0.20 LEN=12)
"Misuse of preposition13"

    +-------------------------------Xp-------------------------------+
    +--------------->WV--------------->+                             |
    +------>Wd------+                  |        +------Jp-----+      |
[...]

PR #1259, opened 2021-11-14
SAT crash and doc fix

ampli:
- Fix a crash caused by the generation code.
- Update the current status of the SAT-parser and how it can be sped up.

BTW, I found another bug in the SAT-parser but I still don't know how to fix it:
```
"On second thought," he said, "this is what I'm going to do".
+++++ error 89
```
```
The only thing that has prevented a collapse of the dollar so far is that it is the currency reserve of the world.
+++++ error 130
```
With `-short=16` these sentences get parsed, but not with `-short=20`.

### DIFF
--- a/link-grammar/prepare/build-disjuncts.c
+++ b/link-grammar/prepare/build-disjuncts.c
@@ -314,7 +314,8 @@ build_disjunct(Sentence sent, Clause * cl, const char * string,
 			ndis->num_categories = 1;
 			ndis->category[0].num = strtol(string, NULL, 16);
 			ndis->category[1].num = 0; /* API array terminator */
-			assert((ndis->category[0].num > 0) && (ndis->category[0].num < 64*1024),
+			assert(sat_solver || ((ndis->category[0].num > 0) &&
+			       (ndis->category[0].num < 64*1024)),
 			       "Insane category %u", ndis->category[0].num);
 			ndis->category[0].cost = cl->cost;
 		}
--- a/link-grammar/prepare/build-disjuncts.c
+++ b/link-grammar/prepare/build-disjuncts.c
@@ -293,9 +293,14 @@ build_disjunct(Sentence sent, Clause * cl, const char * string,
 			*loc = n;         /* update the connector list */
 		}
 
+		bool sat_solver = false;
+#if USE_SAT_SOLVER
+		sat_solver = opts->use_sat_solver;
+#endif /* USE_SAT_SOLVER */
+
 		/* XXX add_category() starts category strings by ' '.
 		 * FIXME Replace it by a better indication. */
-		if (!IS_GENERATION(sent->dict) || (' ' != string[0]))
+		if (sat_solver || (!IS_GENERATION(sent->dict) || (' ' != string[0])))
 		{
 			ndis->word_string = string;
 			ndis->cost = cl->cost;
--- a/bindings/python-examples/parses-demo-sql.txt
+++ b/bindings/python-examples/parses-demo-sql.txt
@@ -19,6 +19,9 @@ OLEFT-WALL this.p is.v another test.n
 O
 
 -max_null_count=1
+% A hack: The SAT parser cannot parse with nulls yet,
+% so change back to classic parser.
+-use_sat=False
 IThis is a a test
 O
 O    +------WV------+----Osm----+
--- a/bindings/python-examples/tests.py
+++ b/bindings/python-examples/tests.py
@@ -808,9 +808,10 @@ class GSQLDictTestCase(unittest.TestCase):
 
     def test_getting_links_sat(self):
         sat_po = ParseOptions(use_sat=True)
-        if sat_po.use_sat != True:
+        if not sat_po.use_sat:
             raise unittest.SkipTest("Library not configured with SAT parser")
         linkage_testfile(self, self.d, sat_po)
+
 
 class IWordPositionTestCase(unittest.TestCase):
     @classmethod
--- a/link-grammar/README.md
+++ b/link-grammar/README.md
@@ -173,6 +173,25 @@ slower, it is a **LOT** slower than the default parser. It is:
 * More than 100x slower on `corpus-fix-long.batch`.
 * 21x slower on Jane Austen's *Pride and Prejudice*.
 
+The SAT solver speed can be significantly increased by these changes:
+- Improve the XOR encoding. It has been tested, yielding a ~2x speedup for
+  long sentences.
+- Use connector sharing. This change alone has sped up the classic parser
+  on batch parsing of `data/en/corpus-fix-long.batch` by 15x, but has less
+  speedup potential for the SAT solver because it doesn't use connectors
+  when it solves the SAT equations.
+- Use the `power_prune()` function of the classic parser instead of its own
+  `power_prune()`. This needs a tricky addition to delete connectors in the
+  word expressions according the discarded disjuncts. It has been
+  tested.
+- Use "tracons". See `disjunct-utils.c` for what they are.
+- Use memory pools.
+- Improve and add hashing.
+- Improve the postprocessing efficiency. For short sentences (also in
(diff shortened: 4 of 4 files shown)


ampli:
```
prepare/build-disjuncts.c: In function ‘build_disjunct’:
prepare/build-disjuncts.c:298:11: error: ‘struct Parse_Options_s’ has no member named ‘use_sat_solver’
  298 |   if (opts->use_sat_solver ||
      |           ^~
```
I don't know how to tell CI to configure with `--enable-sat-solver`.

link-grammar/prepare/build-disjuncts.c
```diff
@@ -295,7 +295,8 @@ build_disjunct(Sentence sent, Clause * cl, const char * string,
 
 		/* XXX add_category() starts category strings by ' '.
 		 * FIXME Replace it by a better indication. */
-		if (!IS_GENERATION(sent->dict) || (' ' != string[0]))
+		if (opts->use_sat_solver ||
```
linas, line 298: This needs to be wrapped with `#ifdef USE_SAT_SOLVER` -- that's why the ci tests are failing.

linas:
We may want to drop the SAT solver entirely. Right now, the only purpose it serves is to say "See? We tried it and it didn't help!" which is an interesting result, but not something that needs to live on in the code.

Here is some history, and the lessons that I learned from it.
* Long ago, LG was kind of slow, and there was a desire to make it run faster.
* Ben looked at the `&` and `or` and mistook them for logical-and and logical-or. (They are not. More on this later (A).)
* Ben hired a grad student via google summer of code, to create the SAT parser.  Grad student promptly realized that  `&` and `or` are not the conventional logical operators, and so considerable complexity ensued.  The project was not as simple as it first seemed to be.
* The resulting parser did run (a lot) faster on long sentences, and maybe 10% slower on short sentences.
* Then you showed up, and made more than a dozen, maybe two dozen performance improvements, including a major one that fixed long-sentence performance. The end result was that the regular parser runs faster than SAT all the time.

The lessons that I learned:
* The SAT algorithm seem like magical voodoo if you don't understand it properly. You can read the Davis-Putnam article on Wikipedia all you want, and you won't understand why SAT is so effective and fast.  (I do not know of a good written explanation.) For many optimization problems, it really is magical in it's effectiveness.
* But like all AI magic, it's only magic because one doesn't understand it. What it actually does is create a connected graph, which has cycles and trees in it, it finds and prunes away all the trees, and then solves the cycles by brute-force exhaustive search. When a solution is found to the cycles, the trees can be re-attached, and the solution can be rapidly ('trivially') propagated out to the tips of the trees. 
* The original link-parser algorithm was "almost the same thing". I think your work with tracons really did convert it into the "same thing".  (Not entirely sure, more on this later (B).)
* Therefore, the SAT solver code can never win, because it must always pay a performance penalty to convert the problem into it's own internal format, solve it there, and then convert back to LG format.  The conversions are pure overhead.
* Therefore, the SAT code can be removed without loss of generality.
* Moral of the story: SAT is not magic, and there are many equivalent formulations of it. 

Here's the "More about this later (A)" part:
* The `&` and `or` are not the conventional logical operators, they are instead two operators from linear logic. The LG dependencies are thus a fragment of linear logic.  This is not just LG; the 'pregroup grammar' is this too (see wikipedia) and so is the 'categorial grammar'. All of these encode the idea of a non-symmetric monoidal category. No, I do not know of any clear written explanation of this. It deserves a clear explanation, it is **very** important.
* Linear logic describes the tensor category. The tensor category describes quantum mechanics.  This is why you see voodoo statements about quantum and language.  For example,  Bob Coecke notes that quantum computers can do natural language parsing in linear time, and provides the algorithm to do this.  (However, language is just a *fragment*, not the whole thing.)
* The LG idea of connectors and disjuncts is way more general than just language; it can also be applied to sound, video and common-sense reasoning. This is **very important**
* Practical applications means that it is important to have fast algos for doing LG-style parsing in a general setting.

Here's the "More about this later (B)" part:
* The original Davis-Putnam SAT algo dates back to the 1960's. Its old. It was formulated in terms of logical-true, logical-false,  and logical-or, logical-and as the connectives. The "graph of cycles and trees" is that, plus the "theory of equality".  A vertex in that graph is a logical disjunct: literally a logical-and of a bunch of variables. An edge in that graph is an 'equals sign': that variable `v23` in disjunct `d66` must equal variable `v23` in disjunct `d78` (literally, it is the same variable. The variable can take the values of true or false.)  The algo works by pretending that these are 'different' variables, but then applies equality to force them to be the same. That's how you get the graph: vertices are disjuncts, and edges are 'links' connecting variables in the disjuncts.  
* When I say 'links' I really really  mean 'links' just like LG links.  In fact, you can implement SAT using LG! (and it would be "just as fast"!)  The connectors of the language are the `v1 ...vn` variables of the problem. They are all multi-connectors (@) and any-directional (ignore +/- for matching).  The problem to be solved is written in DNF, this gives you the dictionary of disjuncts. There is one word per disjunct. SAT solving is LG running in 'generator' mode: if it is able to create at least one sentence, then the problem is satisfiable (and the sentence is one possible solution), If no sentences can be generated, then the problem is not satisfiable.  
* SAT interfaces are nasty, hard to use. Answer set programming (ASP, see wikipedia) are much nicer, more elegant. ASP is more-or-less prolog. (if you fool with it, the Potsdam solver is recommended)
* Moral of the story: LG is a generalization of SAT. It replaces the "theory of equality" with link types. Basically, each link type is a different kind of "equals sign".
* Yes, the "theory of  equality" is a thing. Book by Bader & Nipkow gives an OK description. The wikipedia article is OK, too, I think. All SMT solvers include the theory of equality by default, some do not include anything more than that. All SMT solvers are built on top of SAT.

Here's the only gotcha: the no-links-crossing constraint. Generic SAT does not impose the no-links-crossing constraint. It would be nice to have an LG parser that didn't do this, either.  Only you know how to get from here to there, and you've frequently told me it's impossible. 

This is the only reason to keep the SAT solver code around: it provides an LG-compatible infrastructure, but it also allows me to turn off the no-links-crossing constraint, if I wish.  Hmmm... I need to keep that in mind.

linas:
Edit: the first sentence says "we may want to drop SAT" and the last sentences is effectively "we may want to keep SAT".  So that's an issue.

linas:
I will try to write up the above post into a more formal, more detailed document. However, doing so will take some time, and I have some mundane issues I've been neglecting for too long.

ampli:
> you've frequently told me it's impossible.

It is possible in principle to rewrite the expression/power pruning and `do_count()` to not use the no-links-crossing constraint. However, the tracon optimizations would not work (for both the pruning and parsing), and the need for cache memory would be much larger, things that would result in a vast slowness. If the number of linkages also explodes then very fast, it will be impossible then to find the optimal one by metric sorting.

When I implemented the tunneling idea for cross-links (issue #1050), I found that just enabling a single connector tunnel through a single link has a huge overhead (maybe this can be mostly fixed by tightening the conditions in which particular tunneling are allowed).

Do you have a "complex enough" toy problem that can be formulated by LG and should not have a no-links-crossing constraint, so I can play with it?

linas:
> Do you have a "complex enough" toy problem that can be formulated by LG

Not right now. I have to formalize what I wrote above to make sure I'm thinking about everything correctly, then I have to make some forward progress with the vision/sound stuff (its at version 0.01 so its actually started!) and then I have to think about what, exactly, the role of the existing LG code base is in that setting.

FYI, People use minecraft for all kinds of ... unexpected uses. One of these is the automatic generation of architectural spaces (Roman towns, in the demo). To generate "valid" layouts (where streets connect, etc.) they use constraints. To rapidly solve the constraints, they invented an algo that they've termed "wave function collapse".  I thought about it just long enough to convince myself that it is a viable algo for generic LG sentence generation. Here: https://marian42.de/article/wfc/  I have some vague daydreams of implementing it on the atomspace context.

Issue #1249, opened 2021-09-14
new aspell breaks python unit test


linas:
I have this:
```
$ aspell --version
@(#) International Ispell Version 3.1.20 (but really Aspell 0.60.8)
```
and get this result:
```
link-grammar: Info: Library version link-grammar-5.10.1. Enter "!help" for help.
linkparser> the seasand lakes are hot
Found 50 linkages (50 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS= 0.05 LEN=11)

    +--------------->WV-------------->+
    +----------->Wd------------+      |
    |      +--------Dmc--------+      |
    |      |        +-----A----+--Spx-+--Pa-+
    |      |        |          |      |     |
LEFT-WALL the seasonal[~].a lakes.n are.v hot.a

Press RETURN for the next linkage.
linkparser> 
	Linkage 2, cost vector = (UNUSED=0 DIS= 0.05 LEN=13)

    +------------------>WV------------------>+
    +---------->Wd-----------+------Spx------+
    |      +--Dmc-+<--SJlp<--+->SJrp->+      +--Pa-+
    |      |      |          |        |      |     |
LEFT-WALL the seas[&].n and[&].j-n lakes.n are.v hot.a
```
The python unit test `parses-pos-spell-en.txt` is expecting the second parse and fails when the first one arrives.

ampli:
I also use this exact `aspell` version in my system, and it also returns "seasoned" first:
``` text
$ aspell -a                                                                                                                                                                                     LG (master $% u=)
@(#) International Ispell Version 3.1.20 (but really Aspell 0.60.8)
seasand
& seasand 34 0: seasoned, sea sand, sea-sand, seas and, seas-and, seasons, season, Zealand, peasant, sealant, seasonal, Sand, sand, season's, send, Seaside, sassed, seaside, Susan, stand, reascend, reasoned, seesawed, sensed, Susana, ceased, resend, savant, secant, second, pheasant, ceasing, servant, Susan's
```
However, using the latest master (d57493eed849eccd2c7ff6220d5997b24c2c7766) I get:
``` text
link-grammar: Info: Dictionary found at ./data/en/4.0.dict
link-grammar: Info: Dictionary version 5.10.0, locale en_US.UTF-8
link-grammar: Info: Library version link-grammar-5.10.1. Enter "!help" for help.
linkparser> the seasand lakes are hot
Found 26 linkages (26 had no P.P. violations)
[...]

PR #1260, opened 2021-11-15
Fix issue #1249

ampli:
There were 3 problems:
1. `guess_mispelled_word()` (in `tokenize.c`) wrongly assumes the run-on word are always provided first.
When this is not the case, it can miss the run-on fix that we test.
Fix that by always using all the run-on guesses (if `use_spell_guess>0`) and only then `use_spell_guess` single-word guesses.
2. The test asks only for a single guess.
3. Fix that by asking allowing a big number of guesses (that is never reached).
3. There can be linkages with other guesses (that we cannot know in advance) before the linkage we need for our tests.
Fix that by skipping until we find the guess we check for.\\

The comments in `linkage_testfile()` explain why this (I hope) should work.

### DIFF
--- a/bindings/python-examples/tests.py
+++ b/bindings/python-examples/tests.py
@@ -1255,6 +1255,12 @@ class ZXDictDialectTestCase(unittest.TestCase):
 
 #############################################################################
 
+# Note on the linkage order of next(linkage):
+# tests.py uses add_eqcost_linkage_order(), which sorts linkages with
+# identical metrics according to their diagram character values. Thus the
+# order of linkages with the identical metrics doesn't depend on their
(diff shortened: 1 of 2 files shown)


ampli:
While looking and testing spell-guesses, I found other problems in the spell-guessing code that I intend to address in an additional PR.

ampli:
It was not clear to me whether to refer to the issue or PR number in the ChangeLog line.
I thought that maybe it is better to refer to the issue if there is one.
So I referred here to issue #1249 instead of PR #1260.

linas:
> refer to the issue or PR number in the ChangeLog line.
Either will do. Both should refer to each other, anyway.

PR #1261, opened 2021-11-15
Spell-guessers fixes

ampli:
1. Hunspell: Fix my old and buggy fix for a race condition in `spellcheck_create()`.
2. Aspell: By default, don't print an error if lang files are not found. Only print such an error in verbosity=4 (on a par with the same error from Hunspell - done in an old change). This fixes potential errors messages on startup and while testing when Aspell is configured.

FIXME:
1. Remove the hard-coded language names and spell-guesser dictionary names, and take them from a dict definition.
2. In (1), support using multi dicts, like `en_US+en_GB_db+en_medical`, adding regional dictionaries, etc.

### DIFF
--- a/link-grammar/tokenize/spellcheck-aspell.c
+++ b/link-grammar/tokenize/spellcheck-aspell.c
@@ -60,8 +60,7 @@ void * spellcheck_create(const char * lang)
 		aspell = (struct linkgrammar_aspell *)malloc(sizeof(struct linkgrammar_aspell));
 		if (!aspell)
 		{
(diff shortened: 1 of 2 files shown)


linas:
`en_medical` -- wow, but ... its not in the dict definition, so?

ampli:
There are `data/en/words/words-medical.*` files and these words can get misspelled too.

linas:
I agree they can get misspelled, but you wrote:  ... oh, I see, you wrote FIXME which means that this has not been done yet.  I was just complaining that the FIXME pieces seemed to be ... missing!

PR #1265, opened 2021-12-11
do_count() speedup

ampli:
- Check at the start of  `do_count()` before anything else if the end connector's `next_word` values prevent a linkage. It saves costly table lookups.
- Implement a similar idea in `pseudocount()`. Also, see the comment there.
- Fix a problem in the table debug code due to the recent 32-bit linkage count change. The histogram code compilation that got broken by the 32-bit linkage count change is to be fixed in the next PR.

This patch saves 3%-4% on long sentences, and also (surprisingly maybe), on the `pandp-union` batch.


linas:
Cool!

PR #1267, opened 2021-12-11
Fix histogram compilation break due to the 32-bit linkage-count change

ampli:
BTW
- For the histogram counter, I used a 64-bit int instead of a 32-bit one, as it seems there is no need to improve the performance of the histogram code by a few percent.
- A comment in `histogram.h` says "the code runs about 6% faster when it is disabled".
However, at the time it was written it was not feasible to test it on a long sentence batch.
It turns out that now the impact on short sentences is 1%-3%, but for `fix-long` it is ~40% and for `pandp-union` it is ~25%.
I guess this is due to cache misses on the big cache memory that is needed for parsing long sentences. (I'm now working on reducing the needed parsing cache.)

### DIFF
--- a/link-grammar/parse/histogram.h
+++ b/link-grammar/parse/histogram.h
@@ -17,17 +17,17 @@
 
 #define PARSE_NUM_OVERFLOW (1<<24)  // We always assume sizeof(int)>=4
 
+typedef int32_t count_t;
 typedef int64_t w_count_t;          // For overflow detection
 
 /*
  * Count Histogramming is currently not required for anything, and the
  * code runs about 6% faster when it is disabled.
- *
-#define PERFORM_COUNT_HISTOGRAMMING 1
  */
-#ifdef PERFORM_COUNT_HISTOGRAMMING
+#define PERFORM_COUNT_HISTOGRAMMING 0
 
-typedef int64_t count_t;
+#if PERFORM_COUNT_HISTOGRAMMING
+
 #define COUNT_FMT PRId64
 
 /**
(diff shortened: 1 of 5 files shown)


link-grammar/parse/histogram.h
```diff
@@ -42,36 +42,37 @@ typedef int64_t count_t;
...
-count_t hist_cut_total(count_t* tot, int min_total);
+static inline w_count_t hist_total(Count_bin* tot) { return tot->total; }
+w_count_t hist_cut_total(Count_bin* tot, count_t min_total);
 
-double hist_cost_cutoff(count_t*, int count);
+double hist_cost_cutoff(Count_bin*, count_t count);
 
 #else
 
 typedef int32_t count_t;
```
linas, line 71: this is duplicated above.

ampli: Only the first one is needed. I will remove it in the next cleanup.

linas:
So ... did you graph any of the histograms? What did they look like? 

When I added this histogramming code, I was hoping that the resulting graphs would provide some kind of insight into how the parser worked, or insight into the structure of language, or something.  I tried it a few times, but did not get anything interesting coming out. I don't remember what it was or why it wasn't interesting.  Somehow, I didn't learn anything.... but I've forgotten all that, and it seems I did not write up the results anywhere.

It might be worth repeating this; maybe you'll see something I didn't.  If nothing else, there should be some explanation: "histograms of what?" and "what do they look like?" and "why do they look like that?" and "what does it mean?" ... those four questions should be answered.

ampli:
> So ... did you graph any of the histograms? What did they look like?

I have never tried to look at how to actually use this code... If you have hints for that, they maybe can save me much time in figuring this out. If I just set `PERFORM_COUNT_HISTOGRAMMING` to 1 it just uses the `Count_bin` struct and runs slower, but produces nothing.

> but I've forgotten all that, and it seems I did not write up the results anywhere.

I thought the idea was to get insights into how to handle the combinatorial explosion and/or how to speed up the parsing.
(I have some ideas for both that I am investigating just now, but they are not related to the cost histogram.)

linas:
> If you have hints for that, they maybe can save me much time in figuring this out. 

Hah! No. I would have to study it carefully myself and try to remember what it was for. The meta-issue is: do we even need to have this code? Maybe it should be removed?

>  but produces nothing.

I must have hacked in some manual printfs that dumped the data to a file. I use `gnuplot` to convert dumps like that into graphs; sometimes I hand-write scripts to massage the data first. Obviously, whatever I had found did not seem to merit formalizing this stuff any further. The question is: is that still the case?

> I thought the idea was to get insights into how to handle the combinatorial explosion and/or how to speed up the parsing.

Yes, that was the idea. Much of what I do is to try to understand the statistical distribution of stuff, and that is what this was about. Exactly what can be done with that statistical distribution is ... yet another question to ponder.

The meta-issue is this: the dictionaries encode information; the parser (and now generator) crawl over that information, performing a kind of pattern matching. You've focused a lot on the code that implements the parser. I'm trying to focus on the structure of the data itself, but this is  hard, because it's ill-defined. What does it even mean to say "this data has this structure"? and even if I can answer that question, the next question is "so what?" and "what can I do with the knowledge of that structure?" ... so its all very squishy and indistinct.

link-grammar/parse/histogram.h
ampli: PR #1273.

Issue #1266, opened 2021-12-11
Porting `link-generate` to MS Windows


ampli:
> From: Dennis

> With VS I get errors in generator-utilities.c, mainly to do with [nwords] capture and cclist/cclen , and that's after I located and placed versions of argp.h and getopts.h (gcc header files)

> I do know it's a Work-In-Progress but if you have time or inclination, can you get a version of the link-generator working to compile with VS x64 ?

@linas,
I can do it by adding `argp` as a submodule under the `msvc` directory and also fixing the indicated VLA problem.
If this seems right I will send a PR.

linas:
Sure .. I don't see any problem with this. You're good with this stuff; you know what you're doing.  I'm not even sure why you're asking, other than to say "Hi!" and, well, I guess they don't say "Merry Christmas" in Israel, I assume they say "Happy Hanukah" ... these are probably inappropriate comments for a github issue, but hey... (Delete this if it makes you feel nervous. I don't generally respect "boundaries", as they are called. )

ampli:
I just wanted to be sure my solution of adding `argp` as a GitHub submodule seems reasonable (as opposed to adding the source code itself like the current `minisat`, or leaving it for the users to download/compile like previously done with regex).
BTW, here people usually just say "Happy Holliday" when they refer to Hanukah or any holiday, without explicitly mentioning which holiday.

linas:
Oh. "git submodule" ... that's different. I'm not sure I like them, it make creating tarballs hard. But you are proposing this as an MSVC thing only; are all MSVC uses going to build from git, instead of from tarballs? 

How big is argp and getopts?  If it's just 2-4 files, just copy them. It's not like you need to do a git pull to get the actual latest and newest update.  It wont have a lot of churn.

ampli:
I think it is preferable to let them build from both. But building from git is more reasonable, as the Windows port is not always well checked before a release so a working version is needed to be taken from GitHub (after a fix).

Only argp is needed (not getopt). It is ~300K. See for example https://github.com/ericonr/argp-standalone . It is about 15 C files.

In any case, it is only a small fraction of the tarball size. So the easiest way is to just include is a subdirectory of `msvc`.

linas:
Wow! I suppose it's an annoying question, but maybe it would be easier to use ordinary `getopt`, which I assume is available by default with MSVC?

linas:
Do you need the sophistication?

ampli:
[...]

PR #1269, opened 2021-12-30
Port link-generator to MS Windows

ampli:
See issue #1266.

The main change here is the porting of `link-generator` to MS Windows:
- Install GNU `getopt` under `msvc/getopt`.
- Use `getopt` instead of `argp (still using the original `argp` definitions, which are more readable and easy to change).
- Add `--help` and `-usage` support that exactly mimics those of `argp`.
- Make changes and additions for MSVC and MS Windows support.
- Use `parser-utilities.[ch]` (see below).
- Create `link-generator.bat`.

In addition, implement these changes for `link-generator`:
- Convert numeric arguments with better error checking.
- Add prompt **linkgenerator>** when reading a sentence template (`editline` is not supported yet).
- Refactor system-dependent functions and move them to `parser-utilities.[ch]`. This removes most of the system-dependent stuff from `link-parser.c`, and makes the Windows support for `link-generator` very easy to implement.
- Use library verbosity 1 as a default.

Documentation changes (@linas, please correct my wording if needed):
- Update the `link-generator` manual.
- Update the main README file (among other things, remove the obsolete Windows regex info).

Additional changes:
- link-parser: Remove the use of %n.
- Minor cleanup in `parser-utilities.[ch]`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@ Version 5.10.3 (XXX 2021)
  * Fix memory leak in the "!" link-parser command. #1256
  * Add C++ regex support. It is now the default for MSVC builds. #1258
  * Fix spell-guess for run-on words. #1249
+ * Port link-generator to MS-Windows. #1269
 
 Version 5.10.2 (16 Sept 2021)
  * Fix python install path.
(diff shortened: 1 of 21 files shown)


linas:
Looks reasonable to me!

PR #1275, opened 2022-01-22
minisat: remove `#include <fpu_control.h>`

SoapGentoo:
* `fpu_control.h` doesn't exist on musl, and doesn't do anything
  on glibc anymore for the past years. Remove it to unbreak musl
  builds.

Bug: https://bugs.gentoo.org/717404


linas:
Thanks!

PR #1283, opened 2022-02-26
Remove pointless MAX_WORD limitation

linas:
There were four places that applied a buffer size limit of `MAX_WORD` that weren't really needed.  Remove those uses.  Two usages remain: during dict printing (seems harmless) and in dict reading (seems practical).


linas:
OK, so point 2. Seems to be the important one. Will fix and merge.

linas:
Points 1. & 3.: this was driven by the Thai dicts, which change MAX_WORD from 180 to 1024.  It seemed to be a good time to review how constants were being used.

PR #1286, opened 2022-03-01
Revert to use again -Og for --enable-debug, but allow to override compilation flags

ampli:
The idea to use -Os for debugging was bad, and this hard-coded flag could not be overridden using `configure CFLAGS=-Og` because compilation flags could only be added but not override hard-coded ones.

Fix both of these problems.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -283,11 +283,6 @@ AC_ARG_ENABLE( debug,
 
 if test "x$enable_debug" = "xyes"
 then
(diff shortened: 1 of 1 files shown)


linas:
I merged this, but I don't see `-Os` anywhere in the diffs. Are you sure everything is correct, here?

ampli:
> [...] -Os anywhere in the diffs

I had strange typos in my PR description (fixed).

PR #1288, opened 2022-03-04
Update dockerfile

wannaphong:
I update `dockerfile` from trusty to focal and I add thai to list locales. The python is python3.

### DIFF
--- a/docker/docker-python/Dockerfile
+++ b/docker/docker-python/Dockerfile
@@ -6,10 +6,7 @@
 FROM linkgrammar/lgbase:latest
 MAINTAINER Linas Vepstas linasvepstas@gmail.com
 
-RUN apt-get install -y swig2.0
-RUN apt-get install -y python
-RUN apt-get install -y python-dev
-RUN apt-get install -y zlib1g-dev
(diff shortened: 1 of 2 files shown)


docker/docker-python/Dockerfile
```diff
@@ -6,10 +6,7 @@
 FROM linkgrammar/lgbase:latest
 MAINTAINER Linas Vepstas linasvepstas@gmail.com
 
-RUN apt-get install -y swig2.0
-RUN apt-get install -y python
-RUN apt-get install -y python-dev
-RUN apt-get install -y zlib1g-dev
+RUN apt-get update && apt-get install git build-essential python3-dev libpcre2-dev python-is-python3 libtre-dev wget automake locales libtool flex m4 autoconf-archive autoconf pkg-config swig libthai-dev help2man -y && rm -rf /var/lib/apt/lists/*
```
linas, line 9: Aren't most of these dependencies already installed in the base package?

wannaphong: > Aren't most of these dependencies already installed in the base package?

Yes, there are in `build-essential`.

linas: No, I meant that this is based on `FROM linkgrammar/lgbase:latest` which should already have installed ... oh that's weird. I looked, and it does not already install these packages. Huh. Unexpected. So ... nevermind.

Issue #1289, opened 2022-03-07
List of problematic words


ampli:
Many times the role of missing words can be correctly inferred by the UNKNOWN-WORD device.
However, sometimes the dict defines a word partially (missing POS), or an unknown word is misclassified by regex or gets spell-corrected.

I also have an additional list of words that are totally missing (some of them may have similar problems).

Missing POS
----
affection n
architect v
archive n
continue n
debug n
disaffection n
discarding n
disjunct n
fallback n
hash n
hashtag v
inverse v
merge n
negative n
optional n
outgoing n
padding n
positive n ("false positive" doesn't parse)
postprocessing n-u
predate n
prehistory n
preset a
pruning n
redistribution n
running n
speedup n-u
tar n
tombstone v
trim n

Misclassified by regex
----
hyperfine a (\<MC-NOUN-WORDS>.n)
iterable n (\<ADJ-WORDS>.a)
misdefine v (\<MC-NOUN-WORDS>.n)
multiline a (\<MC-NOUN-WORDS>.n)
preprovide v (\<MC-NOUN-WORDS>.n)
reenable v (\<ADJ-WORDS>.a)
untar v (\<ADJ-WORDS>.a)
versioning n (\<ING-WORDS>.g)

Missing plural noun words are misclassified by regex
----
indirections n (\<S-WORDS>.v)
(There are more.)

Spell corrected (partial list)
----
bitfield n
changelog n
overlink v
toolset n

Idioms
------
flounce out (corpus-failures.batch: And she flounces out.)
make file n

Ref: Comments in PR #1284.

linas:
Trying to fix these now.
 
negative.n  was already in the dict.
tar.n-u and padding.n-u was already in the dict as a mass noun

Do you have example failing sentences for these?

linas:
I made the following verbs transitive.  Perhaps they should be optionally transitive?
```
architect
hashtag
inverse
misdefine
overlink
reenable
tombstone
```
They're transitive, because you can say "I will architect it" but its awkward to say "I will architect"

Do you have any examples where these are intranstive?

ampli:
tar:
The word `tar.s` got added on Apr 30, 2021, and it seems it happened after I have added `tar.n` to my list.
It was mentioned as a kind of substance, and I noticed Wiktionary says it can be countable.

padding:
I don't have the sentence now, but it included the word [`paddings`](https://en.wiktionary.org/wiki/paddings).

(I will add dates and examples to my word list from now on.)

[...]

PR #1290, opened 2022-03-20
Disjunct building speedup

ampli:
This patch speeds up disjunct build for typical sentences by a few tens of percent.
When tested on the English corpus batches, the resulted speedup is a few percent.
(It also speeds up generation by a few percent, but in the case of wildcard words, there is more to do - I'm now preparing a PR for that, that will also fix the buggy unused disjunct count).

Most of the speedup here is due to this simple change:
- `build_clause()`: Implement a more efficient `Tconnector` catenation

In addition, in order to investigate the memory exhaustion problem with the `micro-disinfo` dict when more than one unknown word is used, I added a debug printout (at verbosity=5) just before and after `expression_prune()`.
The result shows the problem:

`echo 'This is a test test' | data/micro-disinfo -v=9 -debug=exprune.c`
```
link-grammar: Info: Dictionary found at ./data/micro-disinfo/dict.db
link-grammar: Info: Dictionary version 5.6.0, locale en_US.UTF-8
link-grammar: Info: Library version link-grammar-5.10.4. Enter "!help" for help.
#### Finished tokenizing (5 tokens)
expression_prune: Initial expression sizes
This[85454] is[85503] a[31417] test[522980] test[522980] 

Debug: Before expression_prune():
This(596854799544) is(483921501177) a(3397807) test(5525349327306844725) test(5525349327306844725) 
Total: 11050699735393387978 disjuncts

expression_prune: l->r pass removed 230135
This[11201] is[35627] a[12196] test[212485] test[503409] 

expression_prune: r->l pass removed 197226
This[10935] is[34499] a[11995] test[203262] test[73204] 

expression_prune: l->r pass removed 14
This[10935] is[34499] a[11995] test[203262] test[73190] 

expression_prune: r->l pass removed 0
This[10935] is[34499] a[11995] test[203262] test[73190] 

Debug: After expression_prune():
This(10276) is(79151) a(9493) test(12166744394147993390) test(4783562197) 
Total: 12166744398931654507 disjuncts

++++ Finished expression pruning                 0.14 seconds
```

Maybe implementing power pruning on expressions may solve this problem for this particular dict (for normal dicts it has no benefit).

In any case (not related to the problem of `micro-disinfo`) I would like to implement a better algo for building disjunct, inspired by `list_links()`.

### DIFF
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -96,7 +96,7 @@ struct Parse_Options_s
 #endif
 
 	/* Options governing the parser internals operation */
-	double disjunct_cost;  /* Max disjunct cost to allow */
+	float disjunct_cost;  /* Max disjunct cost to allow */
 	short min_null_count;  /* The minimum number of null links to allow */
 	short max_null_count;  /* The maximum number of null links to allow */
 	bool islands_ok;       /* If TRUE, then linkages with islands
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -141,6 +141,8 @@ struct Sentence_s
 	Pool_desc * X_node_pool;
 	Pool_desc * Disjunct_pool;
 	Pool_desc * Connector_pool;
+	Pool_desc * Clause_pool;
+	Pool_desc * Tconnector_pool;
 
 	/* Connector encoding, packing & sharing. */
 	size_t min_len_encoding;     /* Encode from this sentence length. */
--- a/link-grammar/api.c
+++ b/link-grammar/api.c
@@ -556,6 +556,16 @@ void sentence_delete(Sentence sent)
 	pool_delete(sent->wordvec_pool);
 	pool_delete(sent->Exp_pool);
 	pool_delete(sent->X_node_pool);
+
+	/* Usually the memory pools created in build_disjuncts_for_exp() are
+	 * deleted in build_sentence_disjuncts(). Delete them here inn case
+	 * build_disjuncts_for_exp() is directly called. */
+	if (sent->Clause_pool != NULL)
+	{
+		pool_delete(sent->Clause_pool);
+		pool_delete(sent->Tconnector_pool);
+	}
+
 	if (IS_DB_DICT(sent->dict))
 	{
 #if 0 /* Cannot reuse in case a previous sentence is not deleted yet. */
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -79,7 +79,7 @@ struct condesc_struct
 	lc_enc_t lc_mask;
 
 	const char *string;  /* The connector name w/o the direction mark, e.g. AB */
-	// double *cost; /* Array of cost by connector length (cost[0]: default) */
+	// float *cost;    /* Array of cost by connector length (cost[0]: default) */
 	connector_hash_t uc_num; /* uc part enumeration. */
 	uint8_t length_limit; /* If not 0, it gives the limit of the length of the
 	                       * link that can be used on this connector type. The
--- a/link-grammar/dict-common/dict-common.h
+++ b/link-grammar/dict-common/dict-common.h
@@ -103,9 +103,9 @@ struct Dictionary_s
 	const char * lang;
 	const char * version;
 	const char * locale;    /* Locale name */
-	double default_max_disjunct_cost;
 	locale_t     lctype;    /* Locale argument for the *_l() functions */
 	int          num_entries;
+	float default_max_disjunct_cost;
 	define_s     define;    /* Name-value definitions */
 
 	bool         use_unknown_word;
--- a/link-grammar/dict-common/dict-structures.h
+++ b/link-grammar/dict-common/dict-structures.h
@@ -14,6 +14,8 @@
 #ifndef _LG_DICT_STRUCTURES_H_
 #define _LG_DICT_STRUCTURES_H_
 
+#include <stdint.h>
+
 #include "link-includes.h"
 
 #ifndef SWIG
@@ -100,6 +102,7 @@ typedef struct
 #ifndef SWIG
 bool cost_eq(float cost1, float cost2);
 const char *cost_stringify(float cost);
+uint64_t count_clause(const Exp *);
 #endif /* !SWIG */
 
 /* API to access the above structure. */
--- a/link-grammar/dict-common/dict-structures.h
+++ b/link-grammar/dict-common/dict-structures.h
@@ -49,11 +49,14 @@ typedef enum { Exptag_none=0, Exptag_dialect, Exptag_macro } Exptag_type;
  * list of operands (when each of them points to the next one through
  * "operand_next"). Else "condesc" is the connector descriptor, when "dir"
  * indicates the connector direction.
+ * For CONNECTOR_type, the "pos" filed is the expression ordinal position
+ * (found when building the disjuncts, for "!!word/m").
  */
 struct Exp_struct
 {
 	Exp_type type:8;      /* One of three types: AND, OR, or connector. */
-	unsigned int unused:24;
+	unsigned int unsued:8;
+	unsigned int pos:16;  /* The position in the expression. */
 	union
 	{
 		struct /* For non-terminals. */
--- a/link-grammar/dict-common/dict-structures.h
+++ b/link-grammar/dict-common/dict-structures.h
@@ -37,7 +37,7 @@ typedef enum
 
 #ifndef SWIG
 static const int cost_max_dec_places = 3;
-static const double cost_epsilon = 1E-5;
+static const float cost_epsilon = 1E-5;
 
 #define EXPTAG_SZ 100 /* Initial size for the Exptag array. */
 typedef enum { Exptag_none=0, Exptag_dialect, Exptag_macro } Exptag_type;
@@ -95,8 +95,8 @@ typedef struct
 } Category_cost;
 
 #ifndef SWIG
-bool cost_eq(double cost1, double cost2);
-const char *cost_stringify(double cost);
+bool cost_eq(float cost1, float cost2);
+const char *cost_stringify(float cost);
 #endif /* !SWIG */
 
 /* API to access the above structure. */
--- a/link-grammar/dict-common/print-dict.c
+++ b/link-grammar/dict-common/print-dict.c
@@ -1156,7 +1156,7 @@ display_word_split_error:
  * Only one minor cheat here: we are ignoring the cost_cutoff, so
  * this potentially over-counts if the cost_cutoff is set low.
  */
-static unsigned int count_clause(Exp *e)
+static unsigned int count_clause(const Exp *e)
 {
 	unsigned int cnt = 0;
 
--- a/link-grammar/dict-common/print-dict.c
+++ b/link-grammar/dict-common/print-dict.c
@@ -30,7 +30,7 @@
 #include "utilities.h"                  // GNU_UNUSED
 /* ======================================================================== */
 
-bool cost_eq(double cost1, double cost2)
+bool cost_eq(float cost1, float cost2)
 {
 	return (fabs(cost1 - cost2) < cost_epsilon);
 }
@@ -38,7 +38,7 @@ bool cost_eq(double cost1, double cost2)
 /**
  * Convert cost to a string with at most cost_max_dec_places decimal places.
  */
-const char *cost_stringify(double cost)
+const char *cost_stringify(float cost)
 {
 	static TLS char buf[16];
 
@@ -119,7 +119,7 @@ static void print_expression_tag_end(Dictionary dict, dyn_str *e, const Exp *n,
 	}
 }
 
-static void get_expression_cost(const Exp *e, unsigned int *icost, double *dcost)
+static void get_expression_cost(const Exp *e, unsigned int *icost, float *dcost)
 {
 	if (e->cost < -cost_epsilon)
 	{
@@ -170,7 +170,7 @@ static void print_expression_parens(Dictionary dict, dyn_str *e, const Exp *n,
                                     bool need_parens, int *indent)
 {
 	unsigned int icost;
-	double dcost;
+	float dcost;
 	get_expression_cost(n, &icost, &dcost);
 	for (unsigned int i = 0; i < icost; i++) dyn_strcat(e, "[");
 	print_expression_tag_start(dict, e, n, indent);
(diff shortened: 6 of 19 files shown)


linas:
> data/micro-disinfo

Wow!  That was from a looong time ago.  I have much better, bigger, more complicated dicts now. Although: ignore the ones I posted a few months ago, they're junk. They were junk even then; they were what I had on hand at the time.

The large set of complex UNKNOWN-WORD declarations was created in a very simple way: for each word-class, there's an UNKNOWN-WORD, with all the disjuncts for that word-class. This seemed like a good idea at the time -- simple, direct. Obviously fails, due to the combinatoric explosion. There might be better ideas, although they would be more complex. 

One alternative idea might be to turn the problem upside-down: given an unknown word, assume that it can have any/every possible set of connectors on it. That is, skip the `fast_match` step completely; assume that any/every connector can connect to an unknown word.  Then do the parse, determine what the connectors would have to be, for it to be valid, and then call a dictionary-provided callback, asking "is this disjunct OK? yes/no?".

There is a related idea I want to pursue for sentence generation, to generate sentence fragments. So, today, the English dict contains the ellipsis "..." and there are five or six variants: a noun-phrase ellipsis, a verb-phrase ellipsis, etc, and so the following sentences parse:  "I gave her a ...", "... all the way to the bank." and so the ellipsis is kind-of-like a wild-card, but it is a kind-of variable-length wild-card. It is kind-of-like an unknown-word, in that almost anything can connect to it -- the current English dict defines the ellipsis with very general connector sets.

I'm playing with these ideas, because someone wrote me an email, asking to compare link-grammar to GPT. One of the standard GPT party-tricks is to "guess the next word": you type in the first few words of a sentence, and it guesses the rest.  Well, we could do something similar, if we mashed up the correct combination of sentence-generation, wild-card, ellipsis and unknown-word.  It shouldn't be that hard -- we already have almost all of the needed ingredients.  (well, OK, sure, we are maybe 5 years too late to that party, since most cell-phone chat apps already try to finish the sentence for you... but this is still a worthwhile trick to demonstrate.)

ampli:
> Then do the parse, determine what the connectors would have to be

If we limit ourselves to disjunct in the dict, wouldn't a wildcard word just do it?

> "I gave her a ...", "... all the way to the bank." and so the ellipsis is kind-of-like a wild-card, but it is a kind-of variable-length wild-card.

I once started to implement that and may return to that.

> I have much better, bigger, more complicated dicts now.

Please provide a link to one, as I use such dicts for code tests.

linas:
>>  Then do the parse, determine what the connectors would have to be
>
> If we limit ourselves to disjunct in the dict, wouldn't a wildcard word just do it?

Wildcards only work in link-generator.  We don't have wild-cards in link-parser.  Wild-cards in the parser require two things: some kind of escape sequence to say "this is a wildcard, right here",  and then, if we do it the old-fashioned way, a dictionary entry listing all possible connectors with @  and {} to make them al optional.  The second part does not require creating any new code, but it seems like it will lead to combinatoric explosion.

>> ellipsis is kind-of-like a wild-card, but it is a kind-of variable-length wild-card.

It turns out this is harder than it seems.  I am unable to implement "It seems that ..." or "it appears as if ..."  because "that" has `(C+ & CV+)` on it  and I cannot sink both of those connectors into a single word.

>> I have much better, bigger, more complicated dicts now.
>
> Please provide a link to one, as I use such dicts for code tests.

If its urgent, I could create one immediately (in about a day); in practice, I was going to actually create and measure the dicts in another month or two.  I've discovered some things that seem more urgent. 

Here's the problem: My "raw data" is a large collection of words, and the disjuncts on them, where each connector can connect only to one single word.  Each word-disjunct pair has a cost.  I can use this to define the similarity between words. I also have a good algorithm for clustering similar words together. It is fancy enough that it will also cluster together only those disjuncts that belong together, leaving the dis-similar disjuncts un-clustered. That is, it can automatically distinguish word-senses.  This works nicely, and I like what I see.

If I set the parameters tightly, then the "similar words" are collections of synonyms.  If I set them loosely, then the clusters grow larger, and become not just collections of synonyms, but collections of words that are grammatically similar, but are not actual synonyms.  Sometimes they grow too big, turning into garbage (not sure, hard to tell...). So there is this smooth landscape (literally: the tops of hills are the similar words; its very high-dimensional. But when I merge two words into one, the nearby landscape changes...). 

I would like to have both: the tight and the loose groupings. This would require two (or three, or more ...) dicts.  For each dict, I want to parse a single sentence and see how the parses differ, and how they are the same ... I'm not sure what I will learn from doing this, but it seems like a good idea.  (Well, one thing I will learn is which disjuncts actually get used, and how often.  It is possible that some will never be used...)

The problem of using sqlite3 is that each export "freezes" the data into place. This causes a data management problem. I've already got dozens (hundreds) of files, that I juggle around, and having dozens of dictionaries, all with different parameter settings, different contents ... ugh. Its a data management nightmare. 

So I'm thinking of writing a new dictionary back-end, that attaches directly to the atomspace, to directly pull out the words and disjuncts "on demand".  This avoid having to "freeze" the contents into one or more sqlite3 dicts.  I have two ways I could write the new dict backend: either to link to the C++ code directly, or to talk to it over the network (the "cogserver"). This would allow the cogserver to run on a different machine... I might start writing this new dict backend in the next few weeks, maybe. Not sure.

I'm having a hard time figuring out what to do next, what the most important thing is: there are too many good ideas, too many things worth doing. They all are complicated and they all take too much time...

linas:
The landscape ... there's this curious phenomenon happening: the larger a cluster gets, the more similar it gets to other words. Its as if there was "gravity" and its a black hole: large clusters tend to suck in more words, and get bigger and bigger. I'm not sure if this is a good thing or not. In network theory, this is called "preferential attachment", and society (people) do it all the time: large websites are popular, and popular scientific papers get cited a lot. Large corporations have lots of products and make lots of money. The result is a power-law distribution. Might be great for people, for society, but I'm not so sure when it comes to words & grammar. Hard to tell.  Anyway, I didn't design it to work like that; it just happens automatically. And so one of the problems is to figure out if some cluster has gotten too big.  Or what it even means to be "too big".  And this is why I want to have multiple dicts: some with large groupings, some with small.

ampli:
> ... if we do it the old-fashioned way, a dictionary entry listing all possible connectors with @ and {} to make them all optional. The second part does not require creating any new code, but it seems like it will lead to combinatoric explosion.

I tested that may years ago, just after I started to hack the library. When you add more connector types, it very fast gets to a combinatoric explosion. However, this way seems unuseful because most of such disjuncts are not found in real words.

> We don't have wild-cards in link-parser.

We can provide a unified operation mode, and maybe it can even be the default.
The current generator is very similar to the parser, as you can see from this:
`link-parser  -s 0 -l en --disjuncts`
```
...
LEFT-WALL this is a test RIGHT-WALL
            LEFT-WALL     0.000  hWd+ hWV+ RW+
              this.p     0.000  Wd- Ss*b+
                is.v     0.000  Ss- dWV- O*t+
                    a     0.000  Ds**c+
              test.n     0.000  Ds**c- Os-
           RIGHT-WALL     0.000  RW-

LEFT-WALL this is a test RIGHT-WALL
            LEFT-WALL     0.000  hWd+ hWV+ RW+
              this.p     0.000  Wd- Ss*b+
                is.v     0.000  Ss- dWV- O*m+
                    a     0.000  Ds**c+
              test.n     0.000  Ds**c- Os-
           RIGHT-WALL     0.000  RW-

LEFT-WALL this is a test RIGHT-WALL
            LEFT-WALL     2.000  hWd+ RW+
              this.p     0.000  Wd- Ss*b+
                is.v     1.000  Ss- O*t+
                    a     0.000  Ds**c+
              test.n     0.000  Ds**c- Os-
           RIGHT-WALL     0.000  RW-

LEFT-WALL this is a test RIGHT-WALL
            LEFT-WALL     2.000  hWd+ RW+
              this.p     0.000  Wd- Ss*b+
                is.v     1.000  Ss- O*m+
                    a     0.000  Ds**c+
              test.n     0.000  Ds**c- Os-
           RIGHT-WALL     0.000  RW-

# Bye.
```
It can be most useful by supplying a user API to put constraints on wild words.
In my current active branch of generator improvements, I called such constraints a "prescription".
(The current spelling mode is a kind of such "prescription".)
Such words may become optional by marking them with the Sentence::word::optional flag.
We can also have a flag to insert a word (constrained by a certain "prescription") only if needed, using your "phantom word" algo.

Each wild word can also have a mode in which it represents a variable number of words, by internally adding a variable number of trailing optional words (i.e. only consecutive words at its end may become optional).

Some optimizations can be made for faster handling of wild words and "prescripted" words. Some of them I would like to include in my next PRs.  Also, the parsing code still has a large room for speedup.

> If its urgent, I could create one immediately (in about a day);

It is not urgent.
One of the things I learned from the current auto-generated dictionaries I have is that their expressions cause a lot of work in build-disjuncts. I have an old branch with an expression simplified that flattens nested AND/OR expressions and this greatly improves the speed of disjunct generation. However, the code uses the old Exp struct, and when I converted it to the new Exp I introduced subtle bugs that I still need to debug. But in any case, the latest improvements in `build_disjuncts()` may make it less sensitive to complex expressions, and anyway I would like to rewrite it to be very much more efficient.

> but collections of words that are grammatically similar, but are not actual synonyms.

This seems very good for constructing an LG dict, as their semantical difference may not be relevant (after all, thoughts can be green...).

>Each word-disjunct pair has a cost. I can use this to define the similarity between words.

...
> So I'm thinking of writing a new dictionary back-end, that attaches directly to the atomspace, to directly pull out the words and disjuncts "on demand".

If words just may have more disjuncts in each dict when is less tight, it seems you can select the disjuncts tightness by just specifying the cost. However, the current disjunct building algo doesn't strictly select disjuncts by their cost (and can emit very costly disjuncts relatively to max-cost).

> I'm having a hard time figuring out what to do next, what the most important thing is: there are too many good ideas, too many things worth doing. They all are complicated and they all take too much time...

I also think about language learning, but my work just now concerns parser and generator features and speed improvements.
The cross-link addition is mostly implemented but needs a discussion on tightening the criteria for cross-linking, as its current specification prevents good pruning and causes an unneeded slowness. I also started to work on phantom worse according to your idea. In addition, I tested a new idea of examing the final parse to decide which parses to display. For example, a word may currently be incorrectly classified by regex (I included such examples in issue #1289) so a sentence may have no parses. If more than one classifying method is used in parallel (e.g. such words also get spell-corrected), parses with spell correction for a certain word may be extracted only if there is no parse with its regex classification (and so on for alternative tokenizations that are now not tried). For further speed up of parsing and generating, I'm examing a new cache in `form_match_list()` and also parallelism of `do_count()`.

ampli:
I edited my post above to include the generator input "This is a test".

linas:
>> We don't have wild-cards in link-parser.
>
> We can provide a unified operation mode, and maybe it can even be the default.

Yes, that would be a good idea.  One major change is needed, though: for each wild-card, we need to print a list of the words that could fit into that location (with the given linkage).

> It can be most useful by supplying a user API to put constraints on wild words.

Yes. The most obviously useful API would be one where the user supplies a list of words the sentence must contain. For example, "Ben " "eat" "pizza", and the goal is to generate sentences that contain those words.  There are several problems here.  One is that all combinations need to be explored; here, there are six possible word orders to explore, including "pizza eats Ben".  The second is the need for handling choice:  so the actual input is (Ben & (eat or eats or eaten or ate) & pizza). A third constraint might be on word order: generate all sentences where the word "Ben" comes before the word "pizza".  

> One of the things I learned from the current auto-generated dictionaries.

These dictionaries are created in the simplest, most obvious way, given my data. I can certainly create them differently, if that would make things easier.

>  flattens nested AND/OR expressions

Huh? The sqlite dictionaries are completely flat. The dictionary entries are just long lists of disjuncts.  All my data is naturally flat, and I do not have any code to unflatten them. But ... that could be an interesting idea. For the more distant future.

linas:
> this is needed if you would like to provide #-style corrections in the dict when one token should, e.g., split, or have an additional token (like `all.#all_of`, `its.#it-is` or even `there.#they're`, currently commented out).

These are commented out, because they result in a dictionary read error.  Yes, what you describe is a great idea, as it does seem solve several problems at once. To rephrase what you imply: 
* Replacing the one word `all` by the two words `all of` solves the phantom-word problem in an elegant way. That is, the places where phantom words can appear are very limited. They cannot appear just anywhere, and they cannot be just anything, and so being able to make lists of these seems to "solve" the phantom-word problem in a simple, easy way.
* It would solve the ellipsis problem, by replacing one ellipsis by two.  I just now pushed a new dict that will parse: "It became clear that ... ..."  The first ellipsis stands in for some unknown subject, and the second for some unknown verb.
* Hebrew: ... 

> In order to finish this implementation, we need to decide on a dict syntax for that.
> I used: `as.#same-as: [[the,same.a,as.e]0.05]colloquial;`, i.e used `[]` for a different purpose 
> (identified by the comma that it includes, when a real comma is to be backslash-escaped) 
> and commas inside macros are skipped.

Sounds perfect to me.

ampli:
> > flattens nested AND/OR expressions
> 
> Huh? The sqlite dictionaries are completely flat. The dictionary entries are just long lists of disjuncts. All my data is naturally flat, and I do not have any code to unflatten them. But ... that could be an interesting idea. For the more distant future.

For example, word `yes.8` from `micro-disinfo`:
`[TOK+ or TQM+]-2.810 or [(TOI- or TOR-) & (TOK+ & (TOG+ or TOH+))]-8.244 or [TOF- & (TPG- & TOL+)]-8.583 or [(TPD- or (TPL- or (TOF- or (TQG- or TOU-)))) & TOL+]-8.631 or ...`

Note these sequences:
`[TOF- & (TPG- & TOL+)]` can be converted to: `[TOF- & TPG- & TOL+]`.
`(TPD- or (TPL- or (TOF- or (TQG- or TOU-))))`  can be converted to: `(TPD- or TPL- or TOF- or TQG- or TOU-)`.

I guess `read-sql.c` can be fixed to create long `&` and `OR` expressions. But the idea of simplifying the expressions after reading them from the dict seemed more general, especially for file dicts, when the expressions can be simplified just after reading them (with the drawback of using more memory since currently the memory for macros is shared).

linas:
> For example, word `yes.8` from `micro-disinfo`:

Oh. Hah. Yes. That's from a much older algo that I do not use any more.  At that time, I was able to cluster words, but was not able to cluster connectors, and so the word-around was to expand a cluster into a list of individual words, with a distinct link type for each word.  Or something like that. I don't remember. I don't do it that way any more.

I don't do it that way any more because 
* There was a bit of a combinatorial explosion in the link types, in the connectors, in the expressions ...
* It didn't seem to be "self-consistent", in that going from word to connector wasn't the same as moving from connector to word, and so "generalizations" seemed wrong.  A lot of the blather about sheaves is about getting this to be self-consistent in the appropriate way. 

An example of the new-style format is here:  https://linas.org/datasets/dict-en.tar.bz2 -- caution: although this is in the new-style, I suspect this is a very low quality dataset. I never actually measured it, but a few quick checks showed it couldn't parse any easy demo sentences I tried.  I don't know why. Probably because its from an incomplete/failed run. Or something ...

Issue #1296, opened 2022-05-23
error loading link-parser for Thai


Bancherd-DeLong:
I received this error while trying to run "link-parser th"

> link-grammar: Info: Dictionary found at ../data/th/4.0.dict
link-grammar: Error: Failed to compile regex: "THAI-NUMBERS: ^[๐-๙,.]*[๐-๙]$": Invalid collation character (code 3)
)Fatal error: Unable to open dictionary.

A few weeks ago, I was able to run "link-parser th" without problems on a different setup/machine.

Thank you for your assistance.

ampli:
This is a problem in the regex implementation of `libc`.
If you have built the library by yourself, try to install the `pcre2` package and then build it again.

Alternatively, do the following changes in `data/th/4.0.regex`:
1. Replace all occurrences of `๐-๙` with `๐๑๒๓๔๕๖๗๘๙` (using `[:digit:]` instead doesn't work).
2. Replace all occurrences of `ก-ฮ` with `[:alpha:]` (this fixes a similar problem in `THAI-PART-NUMBER`).

(For fixing this in the distribution, instead of `[:alpha:]` the patch should list all the Thai alphabetical characters,  since c++ regex is one of the configuration options (default for Windows), and it doesn't support Unicode in `[:alpha:]`. A better option may be adding support for defining/redefining character classes in `4.0.regex` - this looks like simple addition.)

linas:
> In order that all the regex libraries would work fine with Unicode ranges, 

I don't mind. The right person to talk to is @wannaphong and @kaamanita -- Note also, that those guys seem to still be using their own branch of link-grammar, and that they have not realized that I've integrated all of their work into mainline. So ... hi   @wannaphong and @kaamanita  -- please try mainline, and please comment on this bug report!

Bancherd-DeLong:
I know both of them personally.  Will forward this note to them.
Meanwhile, I might make the changes to my local copy.  Thank you.

wannaphong:
> > In order that all the regex libraries would work fine with Unicode ranges,
> 
> I don't mind. The right person to talk to is @wannaphong and @kaamanita -- Note also, that those guys seem to still be using their own branch of link-grammar, and that they have not realized that I've integrated all of their work into mainline. So ... hi @wannaphong and @kaamanita -- please try mainline, and please comment on this bug report!

Hello. Now, I use mainline but I don't see the error.

```
[...]

PR #1297, opened 2022-06-04
4.0.regex: Auto-expand non-ASCII ranges in character classes

ampli:
See issue #1296.
This patch solves the problem of unhandled non-ASCII ranges by expanding them before compiling the regexes.
Checked for the configurable regex libraries.
Can be disabled by adding the dummy definition:
`NO_EXPAND: //`

On the same occasion:
- Improved the error messages.
- Remove the commented-out definition for MAX_REGEX_LENGTH, as increasing its size have no efficiency implications.
- read-regex.h: Add link-includes.h to prevent IDE errors
- Add a test for the 2 problematic `th` regexes. More sentences can be added there if desired.
- Fix the syntax for matching `]` in `HYPHENATED-WORD`, which causes a problem in C++ `<regex>`.

BTW, `ಠ-ಠ` triggers `HYPHENATED-WORD` in PCRE2 but `EMOTICON` in C++ (this seems to be the only difference in the distribution corpora between the regex packages).

Regarding my idea to support:
[:thai-alpha:] : /[ก-ฮ]/
[:thai-digits:] : /[๐-๙]/
This support is not included in this patch, because I found that the direct expansion of the ranges, as done here, is a simpler solution.

But I actually wrote this code and the WIP is in [online here](https://github.com/ampli/link-grammar/tree/regex-class) (it needs a cleanup).
You can see the relevant definitions in [`data/th/4.0.regex`](https://github.com/ampli/link-grammar/blob/regex-class/data/th/4.0.regex). For now, I don't think we need it.

Update: 
In order not to produce unexpected errors in non-fatal cases, I turned all the errors (but the UTF8 bad encoding) to warnings in debug verbosity.  It also seems a good idea to add a runtime regex range test and expand ranges only if it fails (not like the case of issue #1296, it usually doesn't cause a fatal error but just doesn't work as expected).
- expand_character_ranges(): Convert most errors to debug warnings.

### DIFF
--- a/bindings/python-examples/Makefile.am
+++ b/bindings/python-examples/Makefile.am
@@ -25,6 +25,7 @@ EXTRA_DIST =               \
    parses-dialect-en.txt   \
    parses-en.txt           \
    parses-lt.txt           \
+   parses-th.txt           \
    parses-pos-en.txt       \
    parses-pos-he.txt       \
    parses-pos-ru.txt       \
--- a/bindings/python-examples/parses-th.txt
+++ b/bindings/python-examples/parses-th.txt
@@ -1,0 +1,29 @@
+% This file contains test sentences, and the expected parses and
+% constituents output, for the Thai dictionary
+
+% Test that the character ranges in THAI-PART-NUMBER and
+% THAI-NUMBERS actually work.
+
+-display_morphology = True
+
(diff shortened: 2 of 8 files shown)


linas:
Looks good to me. Thanks!

PR #1298, opened 2022-06-12
Parsing acceleration by match list caching

ampli:
Benchmark results:
```  text
1.002x    data/en/corpus-basic.batch
1.000x    data/en/corpus-fixes.batch
1.007x    data/ru/corpus-basic.batch
1.185x    data/en/corpus-fix-long.batch
1.140x    data/en/corpus-failures.batch
1.060x    pandp-union.txt
```
This PR accelerates the `fix-long` corpus batch by ~18%, and the `pandp-union` one by ~6%. It is especially effective for generation - (`link-generator --language=en --length=7` is accelerated ~10x).
Currently, the acceleration is only for parsing w/o nulls.

How it works:
The match lists disjuncts and parse-counts are cached after they are created for the first time. The trick is that only the match list disjuncts that yield a positive count are cached (along with their leftcount/rightcount). The caching of the match lists saves much processing in `form_match_list()` and in the main loop of `do_count()` which doesn't need anymore to iterate over disjuncts that don't yield a count. The cached leftcount/rightcount, which is stored in directly-addressed arrays, saves many branches and `do_count()` calls, and especially `Table_connector` lookups which usually cause CPU cache misses (by average more than one lookup per leftcount/rightcount, due to `multi` connectors).

Notes:
1. The reason that the short-length corpora batches are not accelerated is that the parsing-related functions take negligible time for them - for the `fixes` corpus batch the functions in the files `count.c`, `fast-match.c` and `extract-links.c` take together about 2 (two) percent!

2. I have more parsing acceleration branches, specific generation accelerations and memory usage reductions, and also pruning ideas, that I would like to release before releasing feature PRs.

3. Now that the parsing functions for long sentences take about 25% of the CPU, other speed improvements would become significant (e.g. in tokenization, postprocessing, memory handling, etc.).

### DIFF
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -68,11 +68,8 @@ static inline bool is_connector_subscript_char(unsigned char c)
 }
 /* End of connector string character validation. */
 
-/* The size of the following struct on a 64-bit machine is 32 bytes.
- * It should be kept at this size. If needed, uc_length
- * and uc_start can be eliminate or moved out. Also, there not enough
- * space here to implement cost per connector length - an index to a
- * cost table should be used instead.*/
+/* Note: If more byte-size fields are needed, to save space
+ * uc_length and uc_start may be moved to struct hdesc. */
 struct condesc_struct
 {
 	lc_enc_t lc_letters;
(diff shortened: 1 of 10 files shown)


ampli:
I have just pushed additional commits to address a case of match lists that are bigger than the number of elements in a memory pool block. This doesn't happen even in `en` generation mode (3.4M disjuncts/word), but it seems better to prevent a crash in such an unlikely case.

linas:
Thanks!

PR #1301, opened 2022-06-20
Read dictionaries from the AtomSpace

linas:
Basic implementation. Works. Let's call it "version 0.7" 
Does not yet support word-classes.


linas:
... and its implemeneted incorrectly. Ooops!

PR #1308, opened 2022-06-25
Assorted bug fixes and improvments

ampli:
- `pool_reuse()`: Zero-out the first block if `zero_out==true` 
   It is a potentially severe bug, but no need for an immediate new release (see the detailed commit description).

- Convert `exit()` calls to  a library failure handler hook.
  I left `xalloc()` / `exalloc()` (obsolete but still used) with `exit()`. No memory conditions should be handled in
a new memory management.

- `build_clause()`: Remove shortcut for overcost temporary clauses.
   This seems to have a potential for problems but somehow currently it doesn't have a negative affects (see the detailed commit description).

- `build_disjunct()`: Fix crash on `!!test//m` while using SAT.

- lg_library_failure_hook: Rename assert_failure_trap and export..
  (I implemented this hook long ago, but it is currently inaccessible to the library user.)

BTW, while enabling SAT for the check of this fix, I found that 2 sentences from the basic corpus don't parse when running the batch, but are parsed just fine manually. Maybe this is due to misinitializations. However, I am not going to check that for now.

### DIFF
--- a/link-grammar/dict-common/file-utils.c
+++ b/link-grammar/dict-common/file-utils.c
@@ -530,11 +530,9 @@ char *get_file_contents(const char * dict_name)
 
 			if (err)
 			{
-				char errbuf[64];
-
-				lg_strerror(errno, errbuf, sizeof(errbuf));
+				prt_error("Error: %s: Read error (%s)\n", dict_name,
+				          syserror_msg(errno));
 				fclose(fp);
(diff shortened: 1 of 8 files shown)


linas:
Thanks! It's not worth spending any time debugging the sat solver.

I'm going to start a new discussion on unrelated topics; I'm not sure if you will be automatically notified, so I mention it now.

PR #1312, opened 2022-06-30
Fixes for "amy" and more

ampli:
1. Fix the bit-rot that broke "amy" (see issue #1310).
2. Add tests for "any" and "amy".
3. Fix a memleak in `anysplit.c`.
4. Add `gen-parses-file.py` to generate test diagrams in the needed format and canonical order.

P.S.
https://github.com/opencog/link-grammar/blob/a0e4d727d5e9346207a169e7d36522d8afacc1a8/data/amy/4.0.affix#L19-L22
https://github.com/opencog/link-grammar/blob/a0e4d727d5e9346207a169e7d36522d8afacc1a8/link-grammar/tokenize/anysplit.c#L614-L620

It is possible to fix that, but I will need to know how exactly to mark the extra splits.
Should all the middle ones be marked as `=x=`, or maybe as `x.=`?

### DIFF
--- a/bindings/python-examples/Makefile.am
+++ b/bindings/python-examples/Makefile.am
@@ -27,6 +27,8 @@ EXTRA_DIST =               \
    parses-en.txt           \
    parses-lt.txt           \
    parses-th.txt           \
+   parses-amy.txt          \
+   parses-any.txt          \
    parses-pos-en.txt       \
    parses-pos-he.txt       \
    parses-pos-ru.txt       \
--- a/bindings/python-examples/Makefile.am
+++ b/bindings/python-examples/Makefile.am
@@ -20,6 +20,7 @@ EXTRA_DIST =               \
    LICENSE                 \
    README.md               \
    example.py              \
+   gen-parses-file.py      \
    sentence-check.py       \
    parses-demo-sql.txt     \
    parses-dialect-en.txt   \
--- a/bindings/python-examples/gen-parses-file.py
+++ b/bindings/python-examples/gen-parses-file.py
@@ -1,0 +1,155 @@
+#!/usr/bin/env python3
+"""
+Generate linkage diagrams in the order and format used in the
+parses-xx.txt files of tests.py.
+Note: For the correct order, linkage_limit here and in test.py
+should be enough to produce all the sentence linkages.
+
+TODO: Add constituents.
+"""
+
+import sys
+import re
+import argparse
+
+from linkgrammar import (Sentence, ParseOptions, Dictionary,
+                         LG_Error, LG_TimerExhausted, Clinkgrammar as clg)
+
+import lg_testutils  # Found in the same directory of this test script
+
+def nsuffix(q):
+    return '' if q == 1 else 's'
+
+class Formatter(argparse.HelpFormatter):
+    """ Display the "lang" argument as a first one, as in link-parser. """
+    def _format_usage(self, usage, actions, groups, prefix):
+        usage_message = super(Formatter, self)._format_usage(usage, actions, groups, prefix)
+        return re.sub(r'(usage: \S+) (.*) \[lang]', r'\1 [lang] \2', str(usage_message))
+
+#-----------------------------------------------------------------------------#
+
+
+print("Version:", clg.linkgrammar_get_version())
+
+# Decorate Sentence.parse with eqcost_sorted_parse.
+lg_testutils.add_eqcost_linkage_order(Sentence)
+
+is_stdin_atty = sys.stdin.isatty()
+
+PROMPT = "sentence: " if is_stdin_atty else ""
+
+args = argparse.ArgumentParser(formatter_class=Formatter)
+args.add_argument('lang', nargs='?', default='en',
+                  help="language or dictionary location")
+args.add_argument("-v", "--verbosity", type=int, default=0,
+                  choices=range(0, 199), metavar='[0-199]',
+                  help="1: Basic verbosity; 2-4: Trace; >5: Debug")
+args.add_argument("-d", "--debug", type=ascii, default="",
+                  help="comma-separated list of files and/or functions to debug")
+args.add_argument("-p", "--position", action="store_true",
+                  help="show word sentence position")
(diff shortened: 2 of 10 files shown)


ampli:
Added commits:

- In the amy & any dicts, define locale C just to prevent a dict warning.  Also, convert the version definition to `#define`.
- Non-related to the other commits here: Fix a definition error in `ar/4.0.dict`.

BTW, in any & any on Windows (Windows build fixes are in the next PR), the tokens of non-English characters are wrongly classified as JUNK. I think (not validated) that this is due to the use of C++ regex, as it needs setting the locale for C++. I will try to add a fix in my locale branch (in which I try to fix other locale problems, but it is still not ready for PR). In any case, I think it will still not work fine in multi-lingual sentences under Windows.

linas:
> I will need to know how exactly to mark the extra splits.
> Should all the middle ones be marked as =x=, or maybe as x.=?

I don't know. Just right now, off the top of my head, either form sees OK.  Perhaps `=x=` is best ... 

> in any & amy on Windows 

I expect exactly zero users on Windows. As mentioned before, I use `any` for uniform random sampling; it is wrapped in code, and no human eyes ever see the output as it passes through the pipeline(s).  FYI, the wrappers are here:  https://github.com/opencog/lg-atomese/tree/master/opencog/nlp/lg-parse although they seem to be a bit .. old. They will need refurbishment. So it goes.  As far as I know, no one has tried to compile opencog on windows in over a decade. It does seem to run on Mac (bugs get reported and fixed). I've gotten contradictory reports as to whether it runs on android (there appear to be problems with guile!? Except guile works!?). 

After your fix, the following works great with zany amy: 这是一个测验  and هذا اختبار لجملة أطول and ይህ የረዥም ዓረፍተ ነገር ፈተና ነው። so I'm happy.  But the thai variant นี่คือการทดสอบประโยคที่ยาวขึ้น  does come up `[!JUNK]` on the first linkage, and it never uses the LL link type, whereas the prior three do use the LL link type.  On very rare occasions, the = sign prints weird:
```
linkparser> 
	Linkage 30, cost vector = (UNUSED=0 DIS= 0.00 LEN=1)

            +--------------ANY-------------+
    +--ANY--+              +------ANY------+
    |       |              |               |
LEFT-WALL นี่=[?] คือการทดสอบประโยคท[?].= =ี่ยาวขึ้น[?]
```
... although the above is not how it prints in my terminal window. Heh. I'm guessing that amy chose to insert a break where breaks are not allowed. The `=ี่` seems to be some kind of character modifier and I think there must never be a break there.

The sanskrit variant इदं दीर्घतरवाक्यस्य परीक्षा अस्त gives a combinatorial explosion, which is unexpected:
```
linkparser> इदं दीर्घतरवाक्यस्य परीक्षा अस्ति
link-grammar: Warning: Combinatorial explosion! nulls=0 cnt=2147483647
Consider retrying the parse with the max allowed disjunct cost set lower.
At the command line, use !cost-max
Found 2147483647 linkages (301 of 301 random linkages had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS= 0.00 LEN=1)

    +-----------ANY----------+
    |         +------ANY-----+-------ANY------+-----ANY----+
    |         |              |                |            |
LEFT-WALL इदं[!JUNK] दीर्घतरवाक्यस्य[!JUNK] परीक्षा[!JUNK] अस्ति[!JUNK]

linkparser> 
	Linkage 2, cost vector = (UNUSED=0 DIS= 0.00 LEN=2)

    +---------ANY---------+----------ANY---------+
    |         +----ANY----+----ANY----+          +-----ANY----+
    |         |           |           |          |            |
LEFT-WALL इदं[!JUNK] दीर्घतरवा[?].= =क्यस्य[?] परीक्षा[!JUNK] अस्ति[!JUNK]
```

and again, the LL link is never used. Surprising.

linas:
Amy works great on the following 
* Uigher بۇ ئۇزۇنراق جۈملىنىڭ سىنىقى
* Hebrew זהו מבחן של משפט ארוך יותר
* Armenian: սա ավելի երկար նախադասության փորձություն է
* Greek αυτό είναι ένα τεστ μιας μεγαλύτερης πρότασης
* Korean 이것은 더 긴 문장의 테스트입니다
* Kursigh ئەمە تاقیکردنەوەی ڕستەیەکی درێژترە  although this has a surprising large count

Fails on Khemer: នេះគឺជាការសាកល្បងនៃប្រយោគវែងជាងនេះ។
The above gives only one parse total
Also Lao gives only on parse total: ນີ້ແມ່ນການທົດສອບຂອງປະໂຫຍກທີ່ຍາວກວ່າ

* Anything that uses the sanskrit or related writing system has a count overflow, e.g. 
* Bhojpuri ई एगो लमहर वाक्य के परीक्षा ह  and others.
* Bengali এটি একটি দীর্ঘ বাক্য পরীক্ষা
* Dvihevi އީ ދިގު ޖުމްލައެއްގެ އިމްތިޙާނެކެވެ
* Kurdish ئەمە تاقیکردنەوەی ڕستەیەکی درێژترە
* Malayalam ഇത് ദൈർഘ്യമേറിയ വാക്യത്തിന്റെ പരീക്ഷണമാണ്
* Manipuri ꯃꯁꯤ ꯍꯦꯟꯅꯥ ꯁꯥꯡꯂꯕꯥ ꯋꯥꯍꯩ ꯑꯃꯒꯤ ꯇꯦꯁ꯭ꯠ ꯑꯃꯅꯤ꯫
* Burmese ဤသည်မှာ ပိုရှည်သောဝါကျတစ်ခု၏ စမ်းသပ်မှုတစ်ခုဖြစ်သည်။
* Odia ଏହା ଏକ ଲମ୍ବା ବାକ୍ୟର ପରୀକ୍ଷା |
* Tamil இது ஒரு நீண்ட வாக்கியத்தின் சோதனை

I'm guessing that the failing ones have something to do with utf32 while the working ones might be utf16 .. or something like that.

linas:
Opened #1315 so that this does not fall off the radar.

PR #1313, opened 2022-07-02
Windows fixes

ampli:
1. Fix build break due to the change of  `strltod_l()` to `strltof_l()` in 5.10.4.
    (It means that in 5.10.4 and 5.10.5 building on Windows was broken.)
2. Fix build break due to recent changes (lg_library_failure_hook, dict-ram, dict-atomese).
3. Fix some Windows warnings that it is a good idea to fix anyway.

@linas,
As I cannot afford to compile the library under Windows and macOS too often, if you tell me before a new release I will try to do a build check there in a timely manner.

### DIFF
--- a/bindings/swig/link_grammar.i
+++ b/bindings/swig/link_grammar.i
@@ -22,6 +22,8 @@
 %ignore parse_options_get_max_memory(Parse_Options opts);
 // End of ignored API calls.
 
+%ignore lg_library_failure_hook;     /* Not supported. */
+
 %nodefaultdtor lg_errinfo;
 
 #define link_public_api(x) x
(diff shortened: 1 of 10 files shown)


linas:
> It means that in 5.10.4 and 5.10.5 building on Windows was broken.

Yikes!  I guess that means I should create 5.11.0 as soon as "reasonable".

> As I cannot afford to compile the library under Windows and macOS too often, if you tell me before a new release I will try to do a build check there in a timely manner.

Oh, OK!  The next release will be 5.11.0 and I can release it in a few days, a few weeks, or a month. ... any preferences?

PR #1314, opened 2022-07-03
Find zlib using `pkg-config`

SoapGentoo:
* `AX_CHECK_ZLIB` injects wrong `-L` paths such as `-L/usr/lib`, which in turn
  breaks on multilib systems:

      ld.lld: error: dict-common/.libs/dialect.o is incompatible with elf32-i386
      ld.lld: error: dict-common/.libs/dict-common.o is incompatible with elf32-i386

  To reproduce, use LLD when linking:

      LDFLAGS="-fuse-ld=lld" ./configure --enable-sat-solver

* This also subtly break cross-compilation, since the linker might use libraries
  outside the sysroot.

* By using `pkg-config`, all of these issues are avoided.

Bug: https://bugs.gentoo.org/729908

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -66,7 +66,7 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])], [AC_SUBST([AM_DEFAULT_VERB
 
 dnl "Undefined AX_ macro" is checked in autogen.sh, so if this message
 dnl is changed here it should be changed there too.
-m4_pattern_forbid([^AX_(COMPILER_VENDOR|TLS|AX_CHECK_ZLIB|PKG_SWIG)\b],
+m4_pattern_forbid([^AX_(COMPILER_VENDOR|TLS|PKG_SWIG)\b],
 	[Undefined AX_ macro. Please install the "autoconf-archive" package.])
 m4_pattern_forbid([^PKG_],
 	[Undefined PKG_ macro. Please install the "pkg-config"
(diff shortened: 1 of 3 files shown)


linas:
Thank you! Any comments on the urgency of this issue? We were going to publish a version 5.11.0 in the next week or two, but can do it even sooner...

SoapGentoo:
Oh this isn't urgent at all, but it would be nice to have it in the next release, because we've been carrying this patch for ages (and it's the last patch we've had for link-grammar).

In the next two weeks is absolutely fine!

SoapGentoo:
@linas any news on a new release?

ampli:
@linas,
I originally wanted to send PRs for various affix-file related issues and also for `amy` issues.
For the affix fixes, I still need to complete the efficiency patch.
For the `amy` issues, there is a the >3 parts split (with =middle= morphemes).
There is also the new duplicate-linkages problem that happens on splits with the original word as an alternative (and the old problem of a huge number of "unsane` linkages due to alternatives (that I'm now investigate how to fix).

If we would like to issue a new release ASAP, we can postpone the fixes of the above issues.
In that case, what I still need to do for this release is:
1. Send  a Windows/Mac fixes PR.
2. Send a cleanup PR.

linas:
> various affix-file related issues

Do any of these issues affect English? 

> issue a new release ASAP,

We're overdue for a release. So, yes, soon ...

ampli:
> Do any of these issues affect English?

No (beside maybe a slight speedup).

> We're overdue for a release. So, yes, soon ...

No problem.

ampli:
I have just submitted a cleanup patch for the single-precision floating point change I once started.
Tomorrow I intend to submit the last PR before the new release - a general cleanup.

ampli:
I have just sent my last PR for this release.
There is still a need to fix the affix file for `lt` before the new release.

linas:
I will be unable to create a release in the next two weeks, possibly longer. Technical reasons (out of town; dead computer, login credentials...).

PR #1319, opened 2022-07-07
Fixes for Python

ampli:
- `configure.ac`: Bypass a version check bug in `AX_PYTHON_DEVEL`

`AX_PYTHON_DEVEL` on my system (Fedora 36) has a bug in checking the python version (it makes a simple string comparison). So it thinks that Python 3.10 is less then 3.4.

It was supposed to get fixed by now, but it is not. This guess this may happen in other distributions.
So I just added a bypass of this problem.

- `link_grammar.i`: `Fix PyEval_CallObject()` deprecation.

I fixed it using a version check, so old Python versions would not get affected.


ampli:
I somehow pressed enter while out of the writing box, and triggered the misfeature of sending the PR.
So I will now update the main comment, please read it directly on GitHub.

Issue #1315, opened 2022-07-03
AMY language difficulties


linas:
This recaps posts made on pull req #1312 -- basically AMY works great with some character sets, but not others.

Amy works great on the following 
* Uigher بۇ ئۇزۇنراق جۈملىنىڭ سىنىقى
* Hebrew זהו מבחן של משפט ארוך יותר
* Armenian: սա ավելի երկար նախադասության փորձություն է
* Greek αυτό είναι ένα τεστ μιας μεγαλύτερης πρότασης
* Korean 이것은 더 긴 문장의 테스트입니다
* Georgian ეს უფრო გრძელი წინადადების გამოცდაა
* Chinese 这是一个测验 
* Arabic هذا اختبار لجملة أطول
* Amharic ይህ የረዥም ዓረፍተ ነገር ፈተና ነው።

Fails giving one parse only:
* Khmer: នេះគឺជាការសាកល្បងនៃប្រយោគវែងជាងនេះ។
* Lao: ນີ້ແມ່ນການທົດສອບຂອງປະໂຫຍກທີ່ຍາວກວ່າ

Anything that uses the sanskrit or related writing system has a count overflow, e.g. 
* Bhojpuri ई एगो लमहर वाक्य के परीक्षा ह  and others.
* Bengali এটি একটি দীর্ঘ বাক্য পরীক্ষা
* Dvihevi އީ ދިގު ޖުމްލައެއްގެ އިމްތިޙާނެކެވެ
* Gujarati: આ લાંબા વાક્યની કસોટી છે
* Kurdish ئەمە تاقیکردنەوەی ڕستەیەکی درێژترە
* Malayalam ഇത് ദൈർഘ്യമേറിയ വാക്യത്തിന്റെ പരീക്ഷണമാണ്
* Manipuri ꯃꯁꯤ ꯍꯦꯟꯅꯥ ꯁꯥꯡꯂꯕꯥ ꯋꯥꯍꯩ ꯑꯃꯒꯤ ꯇꯦꯁ꯭ꯠ ꯑꯃꯅꯤ꯫
* Burmese ဤသည်မှာ ပိုရှည်သောဝါကျတစ်ခု၏ စမ်းသပ်မှုတစ်ခုဖြစ်သည်။
* Odia ଏହା ଏକ ଲମ୍ବା ବାକ୍ୟର ପରୀକ୍ଷା |
* Tamil இது ஒரு நீண்ட வாக்கியத்தின் சோதனை

[...]

PR #1320, opened 2022-07-07
Fix amy for scripts that use graphems

ampli:


linas:
These appears to fix #1315

linas:
Thanks!  This was interesting.

Issue #1325, opened 2022-07-12
Potentially undetected count overflow in do_count()


ampli:
While running tests on "amy" with ASN/UBSAN, I got this:
```
... amy/link-grammar/parse/count.c:1390:7: runtime error: signed integer overflow: 2147483647 * 5569944920 cannot be represented in type 'long int'
... amy/link-grammar/parse/count.c:1390:7: runtime error: signed integer overflow: -7518702831433443131 + -6485378441171344729 cannot be represented in type 'long int'
```
This just happens to never occur with the other languages and the various corpus-* files, but theoretically, it could happen due to the current code.

All the calculations are done in signed 64-bit (the storage is in signed 32-bit due to a recent change, but this has no implications here).
It is **signed** 64-bit due to historical reasons, but it can be a good idea to keep it signed because overflows could then be detected by the compiler's UBSAN checks.

The problem arises from the clipping value. It is `INT_MAX`, which is `2^31-1`.
Observe the following code:
https://github.com/opencog/link-grammar/blob/c1b15b8d937c1da7bfdf7f82ef931888f50b1a38/link-grammar/parse/count.c#L1371-L1381
Each of the 4 do_count() calls may return INT_MAX, and hence `leftcount` (and similarly `rightcount`) can be up to `4*(2^31-1)` = `2^33-4`.

Then we have this code:
https://github.com/opencog/link-grammar/blob/c1b15b8d937c1da7bfdf7f82ef931888f50b1a38/link-grammar/parse/count.c#L1390
We may get here up to (`total` + `(2^33-4)` __*__ `(2^31-1)` )  which may be >  **`2^63`** .

And we also have this:
https://github.com/opencog/link-grammar/blob/c1b15b8d937c1da7bfdf7f82ef931888f50b1a38/link-grammar/parse/count.c#L1428
So we may get here (`total` + `(2^33-4)` __*__ `(2^33-4)`) = which is much more than `2^63` (and `total` here may already be near or > **`2^63`**.

Possible solutions:
1. Do nothing, as this has empirically proved not to be a problem with "real" dicts.
2. Clamp `leftcount` and `rightcount` before using them in the multiplications.
3. In `parse_count_clamp()`, clamp to `2^29-1`.

(2) and (3) will add a slight overhead, but by analyzing the overflow possibilities I found some places in which the efficiency can be improved:
[...]

PR #1327, opened 2022-07-14
A fix for undetected count overflows

ampli:
See issue #1325.

I added comments both at a function comment block and inside the function.
naturally, they overlap.

@linas,
As always, please feel free to modify my wording, or propose modified wording and I will resend the PR.

While inspecting the current code, I noted that there is a comment-rot in the function comments of `do_parse()` as `null_count` is no longer an explicit parameter. I left it as it is because the intention is still clear.
(I also noted a yet more speedup potential, that I will inspect after I send the "amy" and tokenizer fixes.)

### DIFF
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -1354,7 +1354,7 @@ static Count_bin do_count(
 
 #define CACHE_COUNT(c, how_to_count, do_count) \
 	{ \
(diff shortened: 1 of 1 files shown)


linas:
I merged, and then edited the file for grammar, and pushed that. I didn't review the code to see if it actually does what you say it does.

PR #1329, opened 2022-07-21
Affix class cleanup

ampli:
My new "amy" modifications are based on 2 other PRs, and it may be a good idea not to combine them all into a big PR.
This is the first one.
After it is applied,  I will make sure there are no conflicts and send the second one.


ampli:
Please don't apply yet. I just found an old bug in the handling of LPUNC.
It seems the problem happened because I once removed the punctuation subscripts of the affix classes (besides UNITS) in `en/4.0.affix`, and then a code rot has happened so it now cannot handle subscript marks on LPUNC anymore (and they exist in some other languages).
I will fix that.

linas:
OK.

ampli:
More in issue #1330.

ampli:
I added fixes according to issue #1330.
Now it may be applied.

**--->** The strippable affixes for `lt` still need attention.
The strippable affixes for `any`, `ady`, and `amy` will be handled in an upcoming PR.
A fix for strippable subscripted affixes will appear in the next PR.
I'm also preparing a PR for CAPSTART.
Finally, I would like to send a cleanup PR for assorted things.

ampli:
> A fix for strippable subscripted affixes will appear in the next PR.

 It turned out it is already included in this PR:
- `LPUNC`,`MPUNCT`: Fix subscript mishandling

ampli:
Sorry again, please don't apply yet!
This patch causes a significant slowness, and I also found a few parse differences (in `en`) that I still need to find their reason.
I need to find out which changes in `tokenize.c` cause the slowness and the parse differences.

ampli:
> I need to find out which changes in `tokenize.c` cause the slowness and the parse differences.

This was a bug in `is_afdict_punc()`: My fixed comparison was more buggy than the original.
This causes both slowness and inadequate tokenization.


**I fixed it and made extensive checks so now finally it may be applied!**

ampli:
Now the affix file may have subscripted affixes again. They are slightly more efficient because their dict lookup is done with their subscript (I guess the possible speedup is small because the pruning will most probably remove the extra disjuncts that may be fetched by an unsubscripted affix).

linas:
OK, thanks

Issue #1333, opened 2022-07-31
Strippable affix class regexes


ampli:
I finished implementing and testinmg it, and here are the examples I used:
``` text
% TODO: this list should be expanded with other "typical"(?) junk
% that is commonly (?) in broken texts.
-- ‒ – — ― "(" ")" "[" "]" ... ";" ±: MPUNC+;
% Split on comma's, but be careful with numbers:
% "The enzyme has a weight of 125,000 to 130,000"
% Also split on colons, but be careful not to mess up time
% expressions: "The train arrives at 13:42"
"/(?<!\d)[,:]|[,:](?!\d)/": MPUNC+;
```
In `corpus-fixes.batch`:
``` text
% Test tokenization by affix regexes.
% Sentence that should not be affected.
The enzyme has a weight of 125,000 to 130,000
The train arrives at 13:42
% Sentences that use punctuation without a trailing whitespace.
We used the same colors (red,blue,yellow).
The price of this item:$100
```
LPUNC and RPUNC also support regexes, and I tested with them `/^[[:punct:]]/`  and `/[[:punct:]]$/`  (respectively) in `amy`.

However, there is a problem: It is supported only when configured with PCRE2, and when configured with C++ the lookbehind regex compilation fails (not supported by C++). POSIX regexes (C library and TRE) also fail. (This is not really a problem for `amy` etc. since we don't need to support other regex libraries there.)

Possible solutions:
1. Distribute it with commented-out affix regexes and that's all.
2. Use autoconf to enable PCRE2 regexes if configure with PCRE2.
3. Add configuration file support for '#if SOMETHING' when SOMETHING is `HAVE_POCRE2_H`.
4.Only support PCRE2 on POSIX systems. (BTW, it is now easy for me to add PCRE2 support on MS-Windows too.)
5. Add support for regex library specification (easy to implement):
`"/(?<!\d)[,:]|[,:](?!\d)/PCRE2"` (or even flag "e" for "extended").

I am for (5) and otherwise for (2) or (1)..

ampli:
> I am for (5) and otherwise for (2) or (1).

I found a better solution, that all the regex libraries support:
Instead of lookahead/lookbehind, use a capture group for the matching part.
e.g, instead of:
`"/(?<!\d)[,:]|[,:](?!\d)/"`
use a POSIX regex:
`"/(?<!\d)([,:]|[,:])(?!\d)/"`
(note the inner parentheses.)

I will change the code to support this too.

[...]

PR #1337, opened 2022-08-12
More amy/anysplit modifications 

ampli:
This patch makes `anysplit.c` potentially grapheme-aware. It works only with the PCRE2 library.
I added a new definition to the affix file: `atomic-unit` that defines a sequence that should not get split. Initially, I defined it as `\X`. But noted that for some reasons that I don't know (I even have no idea if this is due to a bug or feature in PCRE2 or Unicode) sometimes a word starts with mark characters (that may be rendered badly as they don't have a base character to modify). So
I changed it to `\X\pM*". (A better name than `atomic-unit` may be needed to prevent confusion with the atomase...).

I also simplified the regexes in `amy/4.0.regex`. I have left there the non-PCRE2 regexes,  commented out.
In the affix file (through `any/affix-punc`) I added `[[:punct:]]` regexes for RPUNC/LPUNC/MPUNC, to strop off all types punctuations (of course this splits numbers and times too). I don't know how much this is a good idea, but of course, this is optional and can be modified. I have left the multi-character punctuations (e.g. `...`).

Due to the multi-character punctuation, I added the following, which is actually a bug fix to my previous affix-related modifications:
- `afdict_init()`: Validate affixes w/`dictionary_word_is_known()`.

Note that my new `ANY_PUNCT` accepts subscripted punctuation, and if they are used, this allows  to know from which
side they got strip. However, I now see that due to my syntax decision for affix regexes, this way cannot be used with them.
So maybe it will be an improvement to change their syntax to `/regex/\1/` after all (and then `/regex/\1.y/` could be used to add a subscript, and, for example `/regex/\1/a` can specify to split as an alternative (instead of a replacement like now).

amy
Refs: Issues #1334, #1333, #1315; PRs #1334, #1329, #1321.


linas:
Thank you!

PR #1338, opened 2022-08-13
Finish  changes for any / ady / amy

ampli:
- Change `any` and `ady` according to the recent `amy` changes.
- `amy`: Update comments
- Add a fix for commit 56247cb71 (`afdict_init()`: Validate affixes w/`dictionary_word_is_known()`):
-- Read `4.0.regex` before `4.0.affix`

@linas, please make some checks to see that `ady`/`amy`/`any` work as expected.

Regarding the fix for `lt/4.0.affix` that is still needed, I guess that maybe `!` can be added to `"."` in the dict, and the rest should be just null linked there (I can do it unless you have a better idea).

(I still need to send a few PRs regarding `4.0.affix`, an PR for REGPARTS>3, then a cleanup PR, and then fixes for Windows & macOS.)

(Edited to fix a  typo.)

### DIFF
--- a/data/ady/4.0.affix
+++ b/data/ady/4.0.affix
@@ -19,6 +19,13 @@
 
 % Anysplit parameters
 
+% A PCRE2 regex defining a character sequence that shouldn't get split.
+% The LG library must be configured with PCRE2 in order to use it. If not,
+% or if this definition is missing, a single utf8 codpoint is used  as a
+% byte sequence that should not get split.
+%#define atomic-unit "\X";       % split at grapheme boundaries.
+#define atomic-unit "\X\pM*";   % ... but include trailing mark codepoints.
+
 % Number of alternatives to issue for a word. Two values: minimum and maximum.
 % If the word has more possibilities to split than the minimum, but less
 % than the maximum, then issue them without sampling. Else use pseudo-random
--- a/data/ady/4.0.regex
+++ b/data/ady/4.0.regex
@@ -5,17 +5,35 @@
  %                                                                           %
  %***************************************************************************%
 
-% Want to match apostrophes, for abbreviations (I'm I've, etc.) since
-% these cannot be auto-split with the current splitter.
-% Hyphenated words, and words with underbars in them, get split.
-ANY-WORD:  /^[[:alnum:]']+$/
-ANY-PUNCT:  /^[[:punct:]]+$/
+% The regexes here use the PCRE2 pattern syntax.
+% The LG library must be configured with PCRE2 in order to use them.
+
+% Punctuation characters are getting strip from start and end of words,
+% and words that contain punctuation are getting split at them. See the
+% "any/affix-punc" file.
+% These punctuation characters will match here. The \x03 is to match
+% subscripted punctuation that may be specified in this file.
+ANY-PUNCT: /^[[:punct:]]+(:?\x03|$)/
 
 % Simple two-part random morphology: match any string of one or more
 % letters as a stem, and the rest as a suffix.
-% We are currently using .= to denote the end of a stem.
-SIMPLE-STEM: /^[[:alnum:]']+.=/
-SIMPLE-SUFF: /=[[:alnum:]']+$/
+% We are currently using .= to denote the end of a stem (the \x03
+% is the internal representation of the dot in .=).
+SIMPLE-STEM: /^\X+\x03=$/
+SIMPLE-SUFF: /^=\X+/
+
+% Reject anything that contains punctuation, so that the tokenizer will
+% have a chance to split them off as affixes.
+ANY-WORD: /^[^[:punct:]]+$/
+
+% The regexes below don't need the PCRE2 regex library, but:
+% 1. They work fine only with ASCII.
+% 2. They are not checked to work fine with the latest changes
+%    to the other configuration files.
+% ANY-WORD:  /^[[:alnum:]]+$/
+% ANY-PUNCT:  /^[[:punct:]]+$/
+% SIMPLE-STEM: /^[[:alnum:]]+.=/
+% SIMPLE-SUFF: /=[[:alnum:]]+$/
 
 % Match anything that doesn't match the above.
 % Match anything that isn't white-space.
--- a/data/amy/4.0.regex
+++ b/data/amy/4.0.regex
@@ -28,8 +28,6 @@ MOR-SUFF: /^=\X+/
 
 % Reject anything that contains punctuation, so that the tokenizer will
 % have a chance to split them off as affixes.
-% Most of the script-dependent punctuation characters are not mentioned in
-% the "any/affix-punc" file and thus may be included in words.
 ANY-WORD: /^[^[:punct:]]+$/
 
 % For ASCII input and non-PCRE2 regex libraries you can use these instead:
--- a/data/amy/4.0.regex
+++ b/data/amy/4.0.regex
@@ -30,7 +30,10 @@ MOR-SUFF: /^=\X+/
 % have a chance to split them off as affixes.
 ANY-WORD: /^[^[:punct:]]+$/
 
-% For ASCII input and non-PCRE2 regex libraries you can use these instead:
+% The regexes below don't need the PCRE2 regex library, but:
+% 1. They work fine only with ASCII.
+% 2. They are not checked to work fine with the latest changes
+%    to the other configuration files.
 % ANY-WORD: /^[[:alnum:]]+$/
 % ANY-PUNCT: /^[[:punct:]].*$/  % The .* is to match an optional subscript.
 % MOR-PREF: /^[[:alnum:]]+=$/
--- a/data/any/4.0.regex
+++ b/data/any/4.0.regex
@@ -5,14 +5,26 @@
  %                                                                           %
  %***************************************************************************%
 
-% For Unicode scripts that use graphemes, use the corresponding regexes
-% from amy/4.0.regex.
+% The regexes here use the PCRE2 pattern syntax.
+% The LG library must be configured with PCRE2 in order to use them.
 
-% Want to match apostrophes, for abbreviations (I'm I've, etc.) since
-% these cannot be auto-split with the current splitter.
-% Hyphenated words, and words with underbars in them, get split.
-ANY-WORD:  /^[[:alnum:]]+$/
-ANY-PUNCT:  /^[[:punct:]]+$/
+% Punctuation characters are getting strip from start and end of words,
+% and words that contain punctuation are getting split at them. See the
+% "any/affix-punc" file.
+% These punctuation characters will match here. The \x03 is to match
+% subscripted punctuation that may be specified in this file.
+ANY-PUNCT: /^[[:punct:]]+(:?\x03|$)/
+
+% Reject anything that contains punctuation, so that the tokenizer will
+% have a chance to split them off as affixes.
+ANY-WORD: /^[^[:punct:]]+$/
+
+% The regexes below don't need the PCRE2 regex library, but:
+% 1. They work fine only with ASCII.
+% 2. They are not checked to work fine with the latest changes
+%    to the other configuration files.
+% ANY-WORD:  /^[[:alnum:]]+$/
+% ANY-PUNCT:  /^[[:punct:]]+$/
 
 % Match anything that doesn't match the above.
 % Match anything that isn't white-space.
--- a/link-grammar/dict-file/dictionary.c
+++ b/link-grammar/dict-file/dictionary.c
@@ -92,6 +92,32 @@ static bool return_true(Dictionary dict, const char *name)
 }
 
 /**
+ * Process the regex file.
+ * We have to compile regexs using the dictionary locale,
+ * so make a temporary locale swap. XXX FIXME: Thread safety.
+ */
+static bool load_regexes(Dictionary dict, const char *regex_name)
+{
+	if (!read_regex_file(dict, regex_name)) return false;
+
+	const char *locale = setlocale(LC_CTYPE, NULL); /* Save current locale. */
+	locale = strdupa(locale); /* setlocale() uses its own memory. */
+	setlocale(LC_CTYPE, dict->locale);
+	lgdebug(+D_DICT, "Regexs locale \"%s\"\n", setlocale(LC_CTYPE, NULL));
+
+	if (!compile_regexs(dict->regex_root, dict))
+	{
+		locale = setlocale(LC_CTYPE, locale);         /* Restore the locale. */
+		assert(NULL != locale, "Cannot restore program locale");
+		return false;
+	}
+	locale = setlocale(LC_CTYPE, locale);            /* Restore the locale. */
+	assert(NULL != locale, "Cannot restore program locale");
+
+	return true;
+}
+
+/**
  * Read dictionary entries from a wide-character string "input".
  * All other parts are read from files.
  */
@@ -207,6 +233,8 @@ dictionary_six_str(const char * lang,
 	if (!dictionary_setup_defines(dict))
 		goto failure;
 
+	if (!load_regexes(dict, regex_name)) goto failure;
+
 	dict->affix_table = dictionary_six(lang, affix_name, NULL, NULL, NULL, NULL);
 	if (dict->affix_table == NULL)
 	{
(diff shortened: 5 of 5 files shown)


linas:
On startup, these warnings:
```
$ link-parser any
link-grammar: Info: Dictionary found at ../data/any/4.0.dict
link-grammar: Warning: afdict_init: Class LPUNC in file any/4.0.affix: Token(s) not in the dictionary: "....x" "''.x"
link-grammar: Warning: afdict_init: Class RPUNC in file any/4.0.affix: Token(s) not in the dictionary: "....y" "''.y"
link-grammar: Info: Dictionary version 5.4.3, locale C
link-grammar: Info: Library version link-grammar-5.11.0. Enter "!help" for help.
```

I notice that punctuation is *always* split.  Not sure if that is good or bad ... might be bad?
```
linkparser> this is 
Found 4 linkages (4 had no P.P. violations)
linkparser> this"
Found 4 linkages (4 had no P.P. violations)
```
The second would have 5, if the unsplit version was included.  Likewise:
```
linkparser> this...
Found 183 linkages (183 had no P.P. violations)
linkparser> this is a test
Found 156 linkages (156 had no P.P. violations)
linkparser> this is a 
Found 23 linkages (23 had no P.P. violations)
linkparser> this is 
Found 4 linkages (4 had no P.P. violations)
```
Note that 156+23+4 = 183 and so the unsplit variant is never attempted.

Then some mysteries:
```
linkparser> this,,,
Found 179 linkages (179 had no P.P. violations)
linkparser> this!!!
Found 179 linkages (179 had no P.P. violations)
```
Looking at the not-in-dictionary error message more closely:
```
linkparser> this''
Found 27 linkages (27 had no P.P. violations)
linkparser> this!!
Found 23 linkages (23 had no P.P. violations)
```
so the not-in-dictionary thing finds four more combinatoric possibilities ... 
which has 4 less than before.

ampli:
>```
>$ link-parser any
>link-grammar: Info: Dictionary found at ../data/any/4.0.dict
>link-grammar: Warning: afdict_init: Class LPUNC in file any/4.0.affix: Token(s) not in the dictionary: "....x" "''.x"
>link-grammar: Warning: afdict_init: Class RPUNC in file any/4.0.affix: Token(s) not in the dictionary: "....y" "''.y"
>link-grammar: Info: Dictionary version 5.4.3, locale C
>```

It seems you have compiled it with the libc regex. However, now any/ady/amy depend on compiling with PCRE2 (I included a comment on that in the corresponding affix files).
I also would like to add to the `configure` help text and to the main README that using PCRE2 is recommended (it currently picks it up automatically if found). It is also much faster than the alternatives.

>```
>linkparser> this"
>Found 4 linkages (4 had no P.P. violations)
>```
>The second would have 5, if the unsplit version was included.

In principle, when creating an alternative, the unsplit word can be included. If you have a rule on when to do that, I can try to implement it. In a previous post, I mentioned the idea to have an optional modifier flag for regex affixes (it can be for any affix if desired) to tell it to split with an alternative unsplit. BTW, alternatives don't seem useful for e.g. Chinese, which may be written without spaces but with embedded commas (Chinese commas), etc.

> I notice that punctuation is _always_ split. Not sure if that is good or bad ... might be bad?

In English periods may be part of words, and sometimes apostrophes too.  In Hebrew, there are also a few punctuations that may (or may not) be part of words. But I don't know if in general, over all the world scripts, punctuation are mostly included or mostly not included. I can make a short research on that. Note that you may use a more complicated affix regex than just `[[:punct:]]` to tell it which punctuation not to split at all, or which ones not to split in certain conditions.

>so the unsplit variant is never attempted

Yes. Should an unsplit alternative be always included? This can produce a total "mess".

> Then some mysteries:
>```
>linkparser> this,,,
>Found 179 linkages (179 had no P.P. violations)
>```

```
link-parser any -m
linkparser> ,,,
Found 24 linkages (24 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS= 0.00 LEN=0)

    +-----ANY----+
    |            |
LEFT-WALL ,,,[!ANY-PUNCT]
...

So I get only 24 linkages. I will need to see your linkages (or how exactly to repeat your results)  to understand that.

> ```
> linkparser> this''
> Found 27 linkages (27 had no P.P. violations)
> linkparser> this!!
> Found 23 linkages (23 had no P.P. violations)
> ```

Same here.

**Comment by ampli** [comment]
> $ link-parser any
> link-grammar: Info: Dictionary found at ../data/any/4.0.dict
> link-grammar: Warning: afdict_init: Class LPUNC in file any/4.0.affix: Token(s) not in the dictionary: "....x" "''.x"
> link-grammar: Warning: afdict_init: Class RPUNC in file any/4.0.affix: Token(s) not inSpecificallyonary: "....y" "''.y"

Specificly, this happens because libc regex doesn't support the 0x03 notation. I guess that a bare ^C would work for all the regex libraries, but it may not be a good idea to put it in configuration files (and this will help `any` only, as `amy`/`ady` now need PCRE2 anyway). (Support for `\xXX` in non-PCRE2 regexes can be added to `read-regex.c`, but it seems there are better things to do meanwhile...)

**Comment by ampli** [comment]
> I notice that punctuation is always split. Not sure if that is good or bad ... might be bad?

Maybe some punctuations can be split with the unsplit word as an alternative (or not split at all). This is a per-language list, but I don't have an idea how to do that in a manner that fits with unsupervised learning.

[intra-word punctuation mark](https://www.wikidata.org/wiki/Q1383557#:~:text=special%20characters%20that%20are%20part,.%E2%80%9D%2C%20%E2%80%9CU.F.O.%E2%80%9D)

**Comment by linas** [comment]
> It seems you have compiled it with the libc regex.

Indeed. Testing on a new machine, I overlooked this part.

> I also would like to add to the configure help text and to the main README 

Yes please! You could even make it strong: make it fail, with an error message: `PCRE2 is strongly recommended. If you wish to compile without this, then say configure --disable_pcre2`.

>  In principle, when creating an alternative, the unsplit word can be included.

It should be. Examples from English are abbreviations, where the dot must not be split: ft. in. lbs. Inc. Co. Corp. Dr. Mr. Mrs. St. Ave. Ct. Ln. Rd. ... Hmm. Oddly, as I write this, the use of a period feels distinctly old-fashioned. It feels like people have mostly stopped using periods in abbreviations ...

>  Should an unsplit alternative be always included? This can produce a total "mess".

Why?

----
I get the below (with c++ regex) which is same as you:
```
linkparser> ,,,
Found 24 linkages (24 had no P.P. violations)
```
However:
```
linkparser> ...
Found 1 linkage (1 had no P.P. violations)
```

> alternatives don't seem useful for e.g. Chinese, which may be written without spaces but with embedded commas (Chinese commas), 
> [intra-word punctuation mark](https://www.wikidata.org/wiki/Q1383557#:~:text=special%20characters%20that%20are%20part,.%E2%80%9D%2C%20%E2%80%9CU.F.O.%E2%80%9D)

I plan to focus on learning tokenization over the next 6-12 months. So the primary requirement is to be able to use `any` to obtain a uniform random sampling over the space of parse-trees. However, all of my projects seem to take years if not decades, now, so ... who knows. 

Part of this work will be to split words into individual characters. This now seems unavoidable for the Semitic languages, and also for the "synthetic" languages (Lithuanian, Finnish...) and so the task would seem to be "can one learn a finite state transducer-type thing"?

At the same time, I'm interested in segmenting audio and visual data, which is one reason why this might take more than a few years. It has been famously noted that DaLLe is unaware of topology, whereas grammar is all about topology.  I think the technical/research path is clear, just that writing all the code, and doing all the work is just a huge task.

The visual/audio tokenization will have to go through a DSP-filter mechanism (described elsewhere), but the current plan/idea is that the resulting tokens do pass through LG.

**Comment by ampli** [comment]
> Yes please! You could even make it strong: make it fail, with an error message: `PCRE2 is strongly recommended. If you wish to compile without this, then say configure --disable_pcre2`.

No problem.

.> > Should an unsplit alternative be always included? This can produce a total "mess".
> 
> Why?

The current tokenizer and the existing `en` dict **don't handle that gracefully and produce bad parse results**. Unless fixed, an added code that always adds the unsplit word as an alternative should be restricted to `any`/`amy`/`ady`.
In addition, even if we limit this to `any`/`amy`/`ady`,  is it desired, for example, to add `word,` as an alternative to the split result `word` `,`?

Regarding the bad parse results, here are some examples:
1. The current tokenizer, by default, spells-corrects unknown words which include non-letters. For example, the parse result of
`This is a test, he said` then would include `this.p is.v a test[~].n he said.q-d .` because the alternative `test,` is spell-corrected.
2. In the case that the alternative spell corrections are not parsed, the UNKNOWN-WORD device is used (on the word which includes punctuation) and usually produces a "perfect" parse which is nonsense. I found 2 such examples in the English batches, but many more on my various tokenizer tests.

BTW:
1. Some tokenizer bad results can be fixed by deferring tokenization decisions after the parsing step (before producing the linkages). This allows more aggressive tokenization when needed without producing extra bad parses. I started to write a demo for that.
2. Maybe UNKNWON-WORD should handle only letter-only words (i.e. with no digits or punctuation/signs in them). This may be controlled by a dialect. Two dialects may be added (their names here are for example): `strict-unknown-words` for allowing letter-only words (the rest will be null-linked) and `null-unknown-words` to null-link all the unknown words.

**Comment by linas** [comment]
Ah.

The English dict doesn't split "Mr. Jones has left" because "Mr." is in the dict.  However, "This is a test, it seems" is split, because "test," is not in the dict.

For amy/any/ady it wuld be appropriate to consider that "Mr." (unsplit) is in the dict. Perhaps there should be a special, unsplit regex to handle this: 

```
$ git diff
diff --git a/data/any/4.0.dict b/data/any/4.0.dict
index 551e8d888..ec8550b64 100644
--- a/data/any/4.0.dict
+++ b/data/any/4.0.dict
@@ -15,6 +15,7 @@
 LEFT-WALL: @ANY+;
 ANY-WORD: {@ANY-} & {@ANY+};
 ANY-PUNCT: {@ANY-} & {@ANY+};
+ANY-UNSPLIT: {@ANY-} & {@ANY+};
 <UNKNOWN-WORD>: {@ANY-} & {@ANY+};
 
 JUNK: {@ANY-} & {@ANY+};
diff --git a/data/any/4.0.regex b/data/any/4.0.regex
index adf2b9cc1..d74d6374b 100644
--- a/data/any/4.0.regex
+++ b/data/any/4.0.regex
@@ -19,6 +19,8 @@ ANY-PUNCT: /^[[:punct:]]+(:?\x03|$)/
 % have a chance to split them off as affixes.
 ANY-WORD: /^[^[:punct:]]+$/
 
+ANY-UNSPLIT: /^.+$/
+
 % The regexes below don't need the PCRE2 regex library, but:
 % 1. They work fine only with ASCII.
 % 2. They are not checked to work fine with the latest changes
```

Will this work? Am I forgetting anything?

**Comment by linas** [comment]
> Some tokenizer bad results can be fixed by deferring tokenization decisions after the parsing step (before producing the linkages). This allows more aggressive tokenization when needed without producing extra bad parses. I started to write a demo for that.

Huh. That sounds interesting.  Over the next 6-12 months, I hope to maybe produce tokenizers that will take the form of "opaque algorithms", black boxes of some sort. I have not yet thought very deeply about this, yet.

**Comment by ampli** [comment]
> Will this work? Am I forgetting anything?

It seems to work, and it works even for MPUNC:
``` text
linkparser> 
	Linkage 26, cost vector = (UNUSED=0 DIS= 0.00 LEN=1)

    +--------------ANY-------------+
    +-----ANY----+                 +-------ANY-------+
    |            |                 |                 |
LEFT-WALL This[!ANY-WORD] is--a[!ANY-UNSPLIT] test[!ANY-WORD]

linkparser> 
	Linkage 27, cost vector = (UNUSED=0 DIS= 0.00 LEN=1)

                 +-------------ANY-------------+
    +-----ANY----+              +------ANY-----+-----ANY-----+-----ANY-----+
    |            |              |              |             |             |
LEFT-WALL This[!ANY-WORD] is[!ANY-WORD] --[!ANY-PUNCT] a[!ANY-WORD] test[!ANY-WORD]
```

If this is the desired way it should work, I can add your modifications to `any`/`amy`/`ady` (and also add tests to `tests.py` to prevent regressions due to tokenizer modifications).

ampli:
I found two problems:
1. For MPUNC it doesn't generate all possible split alternatives.
E.g. in the example I gave, there is no `is--` and `--a`. I don't know if this needs a fix, and if so, how to fix it.
For `this--is--a--test`, it doesn't give for example `this--is` `--`  `a--test`.
2. It repeats the exact same linkages many times. This looks like a bug.
For  `this--is--a--test`, the first 182 linkages are identical!
I will investigate that.

ampli:
> 2. It repeats the exact same linkages many times. This looks like a bug.
>    For  `this--is--a--test`, the first 182 linkages are identical!
>    I will investigate that.

It seems this is a problem in the wordgraph flatening.
I am now in the process of recalling this stuff.

ampli:
> > 2. It repeats the exact same linkages many times. This looks like a bug.
> >    For  `this--is--a--test`, the first 182 linkages are identical!
> >    I will investigate that.
> 
> It seems this is a problem in the wordgraph flattening. I am now in the process of recalling this stuff.

It turned out the flattening was fine.

The original word `this--is--a--test` and these 2 alternatives are generated and converted to a correct 2-D array:
``` text
SLOT  1                2    3    4   5   6   7  
this--is--a--test
this                   --   is   --  a   --  test
```
(Slot 0 is `LEFT-WORD`.)

Now due to a mix-and-match of alternatives, 170922 linkages are created. Most of them are rejected by `sane_linkage_morphism()`. The problem is that many of the remaining linkages are exactly identical.
Here is a linkage that repeats 182 times:

`LEFT-WALL` `this--is--a--test` `{}` `{}` `{}` `{}` `{}` `{}`
The `{}` denotes null-linked words that were marked "optional" since they belong to a different alternative than  `this--is--a--test`.
`sane_linkage_morphism()` just discards such null-linked words.
The exhaustive search implemented by `do_count()` finds all the combinations of using these null-linked words (I still need to find out why there are 182 combinations). I'm checking now if this can be solved.

linas:
I've lost track of this conversation. Are you describing issues that are created by the diff in comment https://github.com/opencog/link-grammar/pull/1338#issuecomment-1224108213 or are these other, unrelated issues? Is it safe to merge the diff in https://github.com/opencog/link-grammar/pull/1338#issuecomment-1224108213 ?

ampli:
> Is it safe to merge the diff in [#1338 (comment)](https://github.com/opencog/link-grammar/pull/1338#issuecomment-1224108213) ?

When I apply this diff I get the missing linkages (that include the unsplit word itself).
However, it creates **a bad problem** that I still don't understand:  Each linkage repeats many times!
For example, the first linkage includes the single word `this--is--a--test` **and it repeats 182 times**.
So I am not sure it is a good idea to apply it until a solution is found.

Just now I investigate this problem more deeply.
Refer to the slots table from my post above:
From a debug trace I see that the word `this--is--a--test` from the first alternative (in which it is the sole word, at slot 1) is getting linked with the word `--` of the second alternative (slot 2) with link-count=1, as adjacent words with NULL edge connectors.
I added a check of the alternative-origin of the edge NULL connectors to prevent a linkage in case of different alternatives,
but this has a problem in the case of hard-coded NULL connectors in some `do_count()` calls, so it doesn't prevent the linkage between word 1 and 2. So I'm continuing my investigation.

linas:
OK. I got confused about whether these were related. Thanks.  I'll try packaging 5.11.0 in a few days.

ampli:
> I found two problems:
> 
> 1. For MPUNC it doesn't generate all possible split alternatives.
>    E.g. in the example I gave, there is no `is--` and `--a`. I don't know if this needs a fix, and if so, how to fix it.
>    For `this--is--a--test`, it doesn't give for example `this--is` `--`  `a--test`.

I didn't try to "fix" that. Fixing it may cause an exponential explode of linkages due to producing all the possible splits as alternatives.

> 2. It repeats tct same linkages many times. This looks like a bug.
>    For  `this--is--a--test`, the first 182 linkages are identical!

I found the reason for the repeating identical linkages. It is due to the random linkage sampling hitting the same linkage many times. This is because 170922 linkages are produced (about half of them are "insane"), and due to the default `limit=1000`, a sampling is used. The sampling somehow is not uniform and hits the linkage with `this--is--a--test` 182 times.
If e.g. `limit=1000000` is used, there is no sampling and this duplicate linkages problem doesn't happen.

This means that there is a need to use a huge `linkag_limit` value to prevent such duplicate linkages.
The need for using a huge `linkage_limit` can be somewhat mitigated by two `do_count()` patches to prevent "insane" linkages. The first one handles optional words, and the second one validates that the end connectors are from the same alternative.
They may have some (slight) overhead but maybe it is a good idea to apply them nevertheless.

I propose sending a PR with your `ANY-UNSPLIT` fix plus these two patches.

PR #1340, opened 2022-08-15
Configure pcre2 by default

ampli:
(Per discussion at PR #1338.)
- `configure`: Make PCRE2 the default regex library.
This makes `--with-regexlib=pcre2` the default. Because it is specified as the required regex library, it is then an error if it is not found.
- `configure`: Add `--disable-pcre2` to ignore PCRE2.
This provides a way not to use PCRE2 w/o  explicitly specifying another library. `configure` will then pick up the fisrt library according to its hardcoded list. Alternatively, the desired other library can be specified with `--with-regexlib=...`.
- configure: Add a message if PCRE2 is not found.
This is what is shown if PCRE2 is not found:
``` text
...
checking for pcre2.h... yes
checking for PCRE2... no
## -------------------------------------------------------------------------------------------------------- ##
## PCRE2 is strongly recommended. If you wish to compile without this, then say "configure --disable-pcre2" ##
## -------------------------------------------------------------------------------------------------------- ##
configure: error: --with-regexlib=pcre2: libpcre2-8 not found
```
---

How to check this patch: Temporarily rename `libpcre2-8.pc`, then restore it and rename `pcre2.h`. `configure` will then notify the message above unless explicitly not configured with pcre2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,8 +344,10 @@ be without it altogether. The library names may be without the prefix `lib`.
 * `libz1g-dev` or `libz-devel` (currently needed for the bundled `minisat2`)<br>
 * `libedit-dev` (see [Editline](#Editline))<br>
 * `libhunspell-dev` or `libaspell-dev` (and the corresponding English dictionary).<br>
-* `libtre-dev` or `libpcre2-dev` (usually much faster than the libc REGEX
-implementation, and needed for correctness on FreeBSD and Cygwin)
+* `libtre-dev` or `libpcre2-dev` (much faster than the libc REGEX
+implementation, and needed for correctness on FreeBSD and Cygwin).<br>
+Using `libpcre2-dev` is strongly recommended. It must be used on certain
+systems (as specified in their BUILDING sections).
 
 Note: BSD-derived operating systems (including macOS) need the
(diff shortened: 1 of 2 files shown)


ampli:
From the output of `bindings/js CI / build` I see that I missed to notify the PCRE2 recommendation of its  header is not found. I will fix that.

But also the `bindings/js CI / build` setup should be changed to include the PCRE2 package (I don't know how to do that).

``` text
checking for pcre2.h... no
configure: error: --with-regexlib=pcre2: pcre2.h not found
shared:ERROR: Configure step failed with non-zero return code: 1.  Command line: ./configure --disable-editline --disable-sat-solver --disable-java-bindings --disable-python-bindings at /home/runner/work/link-grammar/link-grammar
```

ampli:
Force-pushed to notify the message also if pcre2.h is not found.

(The CI setup should still be changed to include the PCRE2 package.)

linas:
Fixing the testing issue is easy:  commit c67c2f4e1b598454018e97f429f14c89f589edf3

linas:
Well, that's weird. So, another work-around in 891f39277b38c17c80e3191563f31b6b7d25b72f

PR #1344, opened 2022-10-26
Assorted generation fixes

linas:
Several fixes in this pull req. In chronological order:
1. A minor Atomese fix
2. Genereation accidentally required `UNKNOWN-WORD` to be present in the dict.
3. Fix ordinal numbering for unused disjuncts.

FYI @ampli -- The for the last two fixes can be verified by using this dict:
```
LEFT-WALL: A+;
this: A- & B+;
is: B-;
```
and then 
```
link-generator -l ../data/foo -s2  -u
```

Apparently, `LEFT-WALL` is not in any category, which resulted in an off-by-one numbering.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,6 @@
 Version 5.11.1 (XXX 2022)
  * Fix crash when using the Atomese backend.
- * Fix generation issues.
+ * Fix generation tokenization bug when dict has no unknown word token.
 
 Version 5.11.0 (27 Sept 2022)
  * Prototype support for dictionary in the AtomSpace.
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,6 @@
 Version 5.11.1 (XXX 2022)
  * Fix crash when using the Atomese backend.
+ * Fix generation issues.
 
 Version 5.11.0 (27 Sept 2022)
  * Prototype support for dictionary in the AtomSpace.
--- a/link-grammar/dict-atomese/lookup-atomese.cc
+++ b/link-grammar/dict-atomese/lookup-atomese.cc
@@ -185,13 +185,12 @@ static size_t count_sections(Local* local, const Handle& germ)
 {
 	// Are there any Sections in the local atomspace?
 	size_t nsects = germ->getIncomingSetSizeByType(SECTION);
-	if (0 == nsects and local->stnp)
-	{
-		local->stnp->fetch_incoming_by_type(germ, SECTION);
-		local->stnp->barrier();
-	}
+	if (0 < nsects or nullptr == local->stnp) return nsects;
 
+	local->stnp->fetch_incoming_by_type(germ, SECTION);
+	local->stnp->barrier();
 	nsects = germ->getIncomingSetSizeByType(SECTION);
+
 	return nsects;
 }
 
--- a/link-grammar/disjunct-utils.c
+++ b/link-grammar/disjunct-utils.c
@@ -112,9 +112,11 @@ Disjunct ** sentence_unused_disjuncts(Sentence sent)
 
 #define D_DISJ 5                        /* Verbosity level for this file. */
 
+#ifdef USE_SAT_SOLVER
 /**
  * free_disjuncts() -- free the list of disjuncts pointed to by c
  * (does not free any strings)
+ * Almost dead code -- not used anywhere, except by the SAT solver.
  */
 void free_disjuncts(Disjunct *c)
 {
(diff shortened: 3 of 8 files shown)


ampli (commented):
It seems it is not good, since the creation of the whildcard word list should be done only  for words with disjuncts containing categories (their first disjunct is checked for that).
For example, this generation template doesn't need it: th\* is a te\*.

However, I can fix this myself in my patch for a much faster generation (WIP).

ampli:
>New, improved fix for ordinal numbering

It is not good too, since the ordinal numbering should be done:

1. Before the pruning, since the pruning removes disjuncts (and potentially different disjuncts for each word), so the numbering will not be the same as the reference one done in `create_wildcard_word_disjunct_list()` (the "reference disjuncts").
2. Only for `\*` wildcard words (but see comment 3 below), since they don't contain most of the `\*` disjuncts, and those that they contain are in a different order.

Comments:

1. Note that the first packing is done after `expression_prune()`. It is also done again after `power_prune()`.
2. The goal in the ordinal numbering marking is to mark identical disjuncts on different words, and also the "reference disjuncts", with the same ordinal number so letter we can find which are missing.
4. Original bugs: I didn't mark disjuncts of regular and `pref\*` words due to extra code needed for that marking (I differed it for a future fix). The difficulty is due to the need for the same ordinal numbers for identical disjuncts. I also didn't consider the disjunct cost. So missing disjuncts can actually be used in such words or be undetected due to different costs.
5. Regular passing mode doesn't need ordinal number marking, so it is an unneeded overhead to always do it.

linas:
Yep. Closing.

I was unable to discover any kind of ordinal numbering that would stay consistent with what is in the wild-card. As you note, pruning eliminates many disjuncts, and so they cannot be numbers after pruning.  They could be numbered before pruining, except that the `dup_hash` shares the same memory, and wipes out the numbering.

I was hoping that this would provide useful info for debugging atomese, but no I'm no longer sure. That is, I'm not sure this is worth fixing...

ampli:
> They could be numbered before pruining, except that the `dup_hash` shares the same memory, and wipes out the numbering.

I made the numbering just after the duplicate elimination, so I cannot understand why `dup_hash` is a problem.
I haven't checked yet the walls problem, but in case a wildcard word includes the walls their numbering can be skipped.
If the first disjunct may be a wall, the following test for a wildcard word is not god and should be changed:
 https://github.com/opencog/link-grammar/blob/5960b92962feda4aaf2fcf1e1957c1e2ef9b3b92/link-grammar/parse/preparation.c#L217

linas:
>> They could be numbered before pruining, except that the dup_hash shares the same memory, and wipes out the numbering.
>
> I made the numbering just after the duplicate elimination, so I cannot understand why dup_hash is a problem.

1. the generic wild-card is constructed
2. each disjunct in the wild-card is given an ordinal number.
3. A sentences containing wildcards are created.
4. pruning remove most expressions
5. The remaining disjuncts on each word have no relationship whatsoever to the previously numbered wild-card disjuncts.
6. On "possible" solution would be to copy the ordinal numbers from the wild-card to the sentence expressions. This "would work", and we "could see" those numbers, because they would survive pruning. However, duplicate elimination is done after pruning; it uses that same location in storage. So this "possible" solution is impossible.

ampli:
> 6. One "possible" solution would be to copy the ordinal numbers from the wild-card to the sentence expressions.

As the sentence expressions are "compact" and each connector in them, after building the disjuncts, may appear on potentially many disjuncts, I will need more details to understand how you finally get the unique ordinal number.

In any case, by now I found how to efficiently number disjuncts of any word (including fixed ones and walls, and also "constrained" words like `test\*`).

For wild-card words, If the number of disjuncts in them is the same as in the generic one, do it like now (or even not at all, if they got created from the disjuncts of the generic wild-card, an idea to make building disjuncts faster for wild-card words in which the expression pruning pruned nothing). Else concurrently scan both lists and locate the identical disjuncts through their `dup_hash`. For disjuncts of other words, my idea is to keep the `disjunct_dup_table` of the generic wild-card and look up their `dup_hash` there.

Another possibility is not to pre-number at all, and just keep the `disjunct_dup_table` of the generic wild-card word. To find the unused disjuncts, the dup_hash of all the disjuncts will be looked up there (recomputed if needed). This allows reusing the `dup_hash` field (4 bytes) for other purposes (I need additional 2 fields in the Disjunct struct for tokenizer hard priorities in the demo of differing tokenizing decisions to the link extraction time).

I haven't implemented yet any of these ideas, and if you have a better idea I can implement it instead.

ampli:
> This allows reusing the `dup_hash` field (4 bytes) for other purposes (I need additional 2 fields in the Disjunct struct for tokenizer hard priorities in the demo of differing tokenizing decisions to the link extraction time).

It turned out this is not relevant because `ordinal` is not needed in regular parsing mode so it can be reused anyway.

PR #1345, opened 2022-10-26
Misc reorg

linas:
Several fixes in this pull req. In chronological order:

1. A minor Atomese fix
2. Generation accidentally required `UNKNOWN-WORD` to be present in the dict
3. Simplify  wildcard disjunct setup.

This consists of the good fixes from #1344, and without the bad ones.

### DIFF
--- a/link-grammar/parse/preparation.c
+++ b/link-grammar/parse/preparation.c
@@ -200,6 +200,9 @@ void prepare_to_parse(Sentence sent, Parse_Options opts)
 {
 	size_t i;
 
+	if (IS_GENERATION(sent->dict))
+		create_wildcard_word_disjunct_list(sent, opts);
+
 	build_sentence_disjuncts(sent, opts->disjunct_cost, opts);
 	if (verbosity_level(D_PREP))
 	{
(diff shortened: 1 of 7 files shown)


link-grammar/parse/preparation.c
```diff
@@ -200,6 +200,9 @@ void prepare_to_parse(Sentence sent, Parse_Options opts)
 {
 	size_t i;
 
+	if (IS_GENERATION(sent->dict))
```
ampli, line 203: As I mentioned in the closed PR, it is better not to call `create_wildcard_word_disjunct_list()` in case the generator template don't include full wildcard words (`\*`), else a big delay will happen.

Try this in the versions before and after this (#1345) PR::

`This is a test\*`

A solution can be to check if the sentence includes at least one `\*` word before calling `create_wildcard_word_disjunct_list()`.

linas: But if we're in generator mode, then its more-or-less guaranteed that there will be a wild-card, right? Otherwise, what's the point of generation if there are no wild-cards?

ampli: In the case of this template
> `This is a test\*`

`test\*` is not a full wildcard and II even didn't try to number its disjuncts. I also didn't try to number the disjuncts of fixed words.  This is because I didn't have then an efficient algo to do that.

PR #1347, opened 2022-10-29
link-generator: Provide fix for missing disjunct costs.

linas:
Costs on disjuncts were missing, when the generator was running.

### DIFF
--- a/link-grammar/linkage/linkage.c
+++ b/link-grammar/linkage/linkage.c
@@ -903,8 +903,13 @@ float linkage_get_disjunct_cost(const Linkage linkage, WordIdx w)
 	dj = linkage->chosen_disjuncts[w];
 
 	/* dj may be null, if the word didn't participate in the parse. */
-	if (dj) return dj->cost;
-	return 0.0;
+	if (NULL == dj)
+		return 0.0;
+
+	if (dj->is_category)
+		return dj->category[0].cost;
+
+	return dj->cost;
 }
 
 const char * linkage_get_word(const Linkage linkage, WordIdx w)
--- a/link-grammar/linkage/score.c
+++ b/link-grammar/linkage/score.c
@@ -61,8 +61,9 @@ static float compute_disjunct_cost(Linkage lkg)
 	lcost =  0.0;
 	for (i = 0; i < lkg->num_words; i++)
 	{
-		if (lkg->chosen_disjuncts[i] != NULL)
-			lcost += lkg->chosen_disjuncts[i]->cost;
+		Disjunct * dj = lkg->chosen_disjuncts[i];
+		if (dj != NULL)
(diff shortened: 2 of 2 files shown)


link-grammar/linkage/linkage.c
```diff
@@ -903,8 +903,13 @@ float linkage_get_disjunct_cost(const Linkage linkage, WordIdx w)
 	dj = linkage->chosen_disjuncts[w];
 
 	/* dj may be null, if the word didn't participate in the parse. */
-	if (dj) return dj->cost;
-	return 0.0;
+	if (NULL == dj)
+		return 0.0;
+
+	if (dj->is_category)
+		return dj->category[0].cost;
```
ampli, line 910: See my comment here regarding `compute_disjunct_cost()`.

link-grammar/linkage/score.c
```diff
@@ -61,8 +61,9 @@ static float compute_disjunct_cost(Linkage lkg)
 	lcost =  0.0;
 	for (i = 0; i < lkg->num_words; i++)
 	{
-		if (lkg->chosen_disjuncts[i] != NULL)
-			lcost += lkg->chosen_disjuncts[i]->cost;
+		Disjunct * dj = lkg->chosen_disjuncts[i];
+		if (dj != NULL)
+			lcost += dj->is_category ? dj->category[0].cost : dj->cost;
```
ampli, line 66: This is not generally correct, because each disjunct typically contains many categories, each with a potentially different cost. For example, when testing `\* \* \*` on the `en` dict, there are such disjuncts in the linkages (in this printout, the category is in `[ ]`, and after it the cost):

Word 2: num_categories 2
  0: [1169]3.000
  1: [1174]0.000

Word 3: num_categories 6
  0: [1172]1.100
  1: [1173]0.000
  2: [1174]0.000
  3: [1175]0.000
  4: [1185]0.000
  5: [1194]0.000

Word 3: num_categories 6
  0: [1168]3.100
  1: [1169]0.000
  2: [1174]0.000
  3: [1183]0.000
  4: [1187]0.000
  5: [1192]0.000

So the real cost depends on the category that is selected in `link-generator` and  cannot be generally extracted in the library.  However, it can be extracted in  `link-generator`.

BTW, a disjunct can contain even hundreds of categories. However, I noted that while the first one may have any cost, the rest of the categories have always 0 costs. I will investigate if this is a bug (it seems so).

I printed the disjunct's categories an costs using this code:
``` c
if (dj->is_category)
{
	if (dj->num_categories > 0)
	{
		fprintf(stdout, "Word %zu: num_categories %u\n", i, dj->num_categories);
		for (size_t c = 0; c < dj->num_categories; c++)
		{
			fprintf(stdout, "  %zu: [%u]%s\n", c, dj->category[c].num, cost_stringify(dj->category[c].cost));
		}
	}
}
```

PR #1350, opened 2022-11-03
Configuration and word-pair support

linas:
If there aren't enough disjuncts to fully parse a sentences, individual word-pairs can be used to "plug the gaps": these appear as optional connectors on existing disjuncts.

If even that is not enough, then additional optional "ANY" connectors are added to disjuncts, thus always providing a parse, by simply guessing a random linkage.

The pull req includes configuration to enable all of this.


linas:
Part one of this patch set was in #1349

linas:
Part one of this patch set was in #1349

Issue #1353, opened 2022-11-12
Empty subword!?


linas:
I recently experienced a crash (stack trace below) and so I added the check
```
--- a/link-grammar/tokenize/wordgraph.c
+++ b/link-grammar/tokenize/wordgraph.c
@@ -28,6 +28,7 @@ Gword *gword_new(Sentence sent, const char *s)
 
        memset(gword, 0, sizeof(*gword));
        assert(NULL != s, "Null-string subword");
+       assert(0 != *s, "Empty-string subword");
        gword->subword = string_set_add(s, sent->string_set);
 
        if (NULL != sent->last_word) sent->last_word->chain_next = gword;
```

As a result of this, `tests.py` fails.  Investigating now.
The crash is this:
```
assert_failure (cond_str=cond_str@entry=0x7fffecce8be3 "temp_wend>w",
    func=func@entry=0x7fffecce9108 <__func__.10> "strip_right",
    src_location=src_location@entry=0x7fffecce7cb0 "../../link-grammar/tokenize
/tokenize.c:2096",
    fmt=fmt@entry=0x7fffecce8bc6 "Unexpected empty-string word")
    at ../../link-grammar/error.c:467
467                     DEBUG_TRAP;  /* leave stack trace in debugger */
               \

(gdb) bt
#0  assert_failure (cond_str=cond_str@entry=0x7fffecce8be3 "temp_wend>w",
    func=func@entry=0x7fffecce9108 <__func__.10> "strip_right",
    src_location=src_location@entry=0x7fffecce7cb0 "../../link-grammar/tokenize
/tokenize.c:2096",
    fmt=fmt@entry=0x7fffecce8bc6 "Unexpected empty-string word")
    at ../../link-grammar/error.c:467
#1  0x00007fffeccd0f56 in strip_right (sent=sent@entry=0x7fff38411430,
    w=0x7fff382c6f10 "", wend=wend@entry=0x7fff44ff7c78,
    stripped=stripped@entry=0x7fff44ff7cd0,
    n_stripped=n_stripped@entry=0x7fff44ff7c68, p=p@entry=2,
    rootdigit=<optimized out>, classnum=<optimized out>)
    at ../../link-grammar/tokenize/tokenize.c:2096
#2  0x00007fffeccd4ea0 in separate_word (sent=sent@entry=0x7fff38411430,
    unsplit_word=unsplit_word@entry=0x7fff383e22f0, opts=0x7fff381fce10)
[...]

PR #1354, opened 2022-11-13
Fix pcre2 regex so that it's thread safe!

linas:
I think this fixes issue #1353 .. will test

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@ Version 5.12.0 (XXX 2022)
  * Major Atomese dictionary extensions, including generation support.
  * Minor tweaks to `any` uniform random parse tree language.
  * Include U+202F NARROW NO-BREAK SPACE as a space character.
+ * Fix PCRE2 regex so that it's thread safe!
 
 Version 5.11.0 (27 Sept 2022)
  * Prototype support for dictionary in the AtomSpace.
--- a/link-grammar/dict-common/regex-morph.c
+++ b/link-grammar/dict-common/regex-morph.c
@@ -206,6 +206,7 @@ static bool reg_comp(Regex_node *rn)
 	                            options, &rc, &erroffset, NULL);
 	if (re->re_code != NULL)
 	{
+		re->re_md_key = 0;
 		int trc = tss_create(&re->re_md_key, (tss_dtor_t) pcre2_match_data_free);
 		if (thrd_success != trc)
 		{
@@ -266,6 +267,7 @@ static void reg_span(Regex_node *rn)
 static void reg_free(Regex_node *rn)
 {
 	reg_info *re = rn->re;
+	tss_delete(re->re_md_key);
 	pcre2_code_free(re->re_code);
 	free(re);
 	rn->re = NULL;
--- a/link-grammar/dict-common/regex-morph.c
+++ b/link-grammar/dict-common/regex-morph.c
@@ -57,7 +57,6 @@ LINK_END_DECLS
 #if HAVE_PCRE2_H
 typedef struct {
 	pcre2_code *re_code;
-	tss_t re_md_key;
 } reg_info;
 #endif
 
@@ -77,6 +76,8 @@ typedef struct {
 #endif
 
 #define ERRBUFFLEN 120
+
+#define MAX_CAPTURE_GROUPS 10
 
 /* Define reg_comp(), reg_match() and reg_free(), and reg_span()
  * for the selected regex implementation. */
@@ -177,19 +178,29 @@ static void reg_free(Regex_node *rn)
 	free(re);
 	rn->re = NULL;
 }
+
+static void reg_finish(void) {}
 #endif // HAVE_REGEX_H
 
 /* ========================================================================= */
 #if HAVE_PCRE2_H
 
-static pcre2_match_data* get_re_md(reg_info *rif)
+static tss_t re_md_key;
+
+static pcre2_match_data* get_re_md(void)
 {
-	pcre2_match_data* md = (pcre2_match_data*) tss_get(rif->re_md_key);
+	if (0 == re_md_key)
+	{
+		int trc = tss_create(&re_md_key, (tss_dtor_t) pcre2_match_data_free);
+		assert(thrd_success == trc, "Unable to create thread key");
+	}
+
+	pcre2_match_data* md = (pcre2_match_data*) tss_get(re_md_key);
 	if (md) return md;
 
-	md = pcre2_match_data_create_from_pattern(rif->re_code, NULL);
+	md = pcre2_match_data_create(MAX_CAPTURE_GROUPS, NULL);
 	assert(md, "Unable to allocate pcre2 match data struct");
-	tss_set(rif->re_md_key, md);
+	tss_set(re_md_key, md);
 	return md;
 }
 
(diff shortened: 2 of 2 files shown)


linas:
This is currently failing because the russian dict require more than 1024 regexes.

ampli:
In any case, allocating match data per `reg_match()` (by `malloc()` or TLS) instead of per `reg_comp()` seems to me much better!
I can send a PR.

ampli:
> ` Be sure to delete keys when not in use.`

I don't understand how it fixes the 1024 limit, as `reg_free()` is called only on dict close.

ampli:
> I can send a PR.

In my fixed implementation, I removed `re_md` from `reg_info`, added it as an argument to `reg_match()`) and allocated it on the stack in `matchspan_regex()`. (It is NULL when `reg_match()` is invoked from `match_regex()`.)

I also modified  `reg_span()` to directly assign the start/end values and removed `ovector` from `Regex_node_s`.

I still have to complete and test it, thus I will be able to send a PR only tomorrow.

linas:
I just pushed one more commit. This one uses only one key, grand-total, and a per-thread `pcre2_match_data` This fix is good enough for me.  The other kinds of regexes need to be fixed the same way. 

If you have yet something else in mind, that's fine, but I think my last patch provide a minimalist solution, in terms of code changes as well as in terms of the number of malloc's/frees.

linas:
And as I'm currently happy with this, I will merge. This is worth issuing a new release, as its a fairly ugly bug.

ampli:
> If you have yet something else in mind, that's fine, but I think my last patch provide a minimalist solution, in terms of code changes as well as in terms of the number of malloc's/frees.

My fix works not only for PCRE2 but also for the other regex implementations. It currently uses `malloc()` but can be modified to use TLS for PCRE2 and stack allocation for the other regex implementations.

**In any case:**
Please note that there is still a similar problem in `reg_span()`, that uses `ovector[2]` in `Regex_node_s` which is common for all the threads.

My proposed fix is to remove `ovector` from `Regex_node_s` and modify `reg_span()`.
In my fix I used this code:
``` c
static void reg_span(Regex_node *rn, void *re_md, int *start, int *end)
{
	int cgn = rn->capture_group;
	pcre2_match_data *md = (pcre2_match_data *)re_md;
	PCRE2_SIZE *ovector = pcre2_get_ovector_pointer(md);

	if (unlikely(cgn >= (int)pcre2_get_ovector_count(md)))
	{
		*start = *end = -1;
	}
	else
	{
		*start = (int)ovector[2*cgn];
		*end = (int)ovector[2*cgn + 1];
	}
}
```
(I used `void *` because I made similar changes in the other versions of `reg_span()`.)

linas:
Yes, please send a pull req for the reg_span work.  I hadn't noticed that, and, so far, it seems not to be causing any explicit trouble.  I'm guessing the race window is much narrower.

linas:
I mean, this isn't that hard, I can do this too!?

linas:
Ah, what the heck, I've got some time to kill; I wiil try this right now.

ampli:
The `tss` fix is still buggy.

You can see that if you run `tests.py` with ASAN:
``` text
test_specifying_parse_options (__main__.BParseOptionsTestCase) ... ok
test_that_ParseOptions_cannot_get_positional_arguments (__main__.CParseOptionsTestCase) ... ok
test_that_invalid_option_properties_cannot_be_used (__main__.CParseOptionsTestCase) ... ok
test_that_invalid_options_are_disallowed (__main__.CParseOptionsTestCase) ... ok
test_that_sentence_can_be_destroyed_when_linkages_still_exist (__main__.CParseOptionsTestCase)
If the parser is deleted before the associated swig objects ... AddressSanitizer:DEADLYSIGNAL
=================================================================
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer: nested bug in the same thread, aborting.
```

Since you have a single match-data buffer per thread, a problem may happen when more than one dict is used (like in `tests.py` and `tests/multi-thread.cc`.

Instead, I propose to use my fix, which uses the traditional solution for multi-threading (the one that is used in the LG library): Allocating buffers on the stack and passing buffers as function arguments (no malloc/free per call). This fix also doesn't interfere with Emscripten.

E.g:
``` diff
-static bool reg_match(const char *s, const Regex_node *rn)
+static bool reg_match(const char *s, const Regex_node *rn, void *re_md)
```

I can send a PR for your review.

ampli:
> AddressSanitizer: nested bug in the same thread, aborting.

> a problem may happen when more than one dict is used

I now see that this is not the problem. However, apparently is a problem with the current repository code.
I'm trying to debug it.

ampli:
I added several printf's in `get_re_md()`:
``` c
static pcre2_match_data* get_re_md(void)
{
	fprintf(stderr, "re_md_key: %u\n", re_md_key);
	if (0 == re_md_key)
	{
		int trc = tss_create(&re_md_key, (tss_dtor_t) pcre2_match_data_free);
		assert(thrd_success == trc, "Unable to create thread key");
		fprintf(stderr, "tss_create re_md_key=%u\n", re_md_key);
	}

	pcre2_match_data* md = (pcre2_match_data*) tss_get(re_md_key);
	fprintf(stderr, "tss_get re_md_key=%u md=%p\n", re_md_key, md);
	if (md) return md;

	md = pcre2_match_data_create(MAX_CAPTURE_GROUPS, NULL);
	assert(md, "Unable to allocate pcre2 match data struct");
	tss_set(re_md_key, md);
	fprintf(stderr, "tss_set re_md_key=%u md=%p\n", re_md_key, md);
	return md;
}
```
When  I run `tests.py` I get:
``` text
...re_md_key: 0
tss_create re_md_key=0
tss_get re_md_key=0 md=(nil)
tss_set re_md_key=0 md=0x611000037b80
re_md_key: 0
tss_create re_md_key=2
tss_get re_md_key=2 md=(nil)
AddressSanitizer:DEADLYSIGNAL
=================================================================
AddressSanitizer:DEADLYSIGNAL
AddressSanitizer: nested bug in the same thread, aborting.
```

I cannot understand why it creates `re_md_key=0`.

ampli:
With ASAN I get (when running `tests.py`):
``` text
...re_md_key: 0
tss_create re_md_key=0
tss_get re_md_key=0 md=(nil)
tss_set re_md_key=0 md=0x611000037b80

Program received signal SIGSEGV, Segmentation fault.
__interceptor___tls_get_addr (arg=0x7ffff79391d0) at ../../../../libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:5398
5398	  COMMON_INTERCEPTOR_GET_TLS_RANGE(&tls_begin, &tls_end);
(gdb) 
```

linas:
I now see I should have used `call_once()` to create the key; I'm guessing that ASAN is making threads race on purpose, and two different threads try to create the key.

Yes, allocing on stack is much better. For some reason, I thought the data had to survive round-trips back to the user,  but it seems that is not the case. 

Alloc on-stack seems like a good idea, these structures seem to be 160 bytes or less. The only tricky part is that `pcre2_match_data_create()` does not seem to have an on-stack variant that I can find.

linas:
I just now pushed a change to use `call_once` when creating the key. I'm curious to hear if that fixes the ASAN crash.

If you think you have an adequate on-stack solution, I'll take that patch. This solution is indeed a bit over-complicated for what is needed.

ampli:
> The only tricky part is that `pcre2_match_data_create()` does not seem to have an on-stack variant that I can find.

There is `pcre2_get_match_data_size()`. It can be used (one time per thread and then freed) to find the size of `pcre2_match_data_create()`, and this size can be allocated on the stack. Another problem is that `pcre2_match_data_create()` makes an initialization, so its returned value should be cached as a template and copied to the allocated buffer on the stack. Its size for `MAX_CAPTURE_GROUPS` is 240, so this has some overhead (but it may be relatively negligible compared to tens of regex matches that follow). It is possible to significantly reduce the copy size by keeping a template per Regex_node, with exactly the needed size (which is usually only 1 capture group).

This works, and the problem in my current code is in the c++ regex - the stack allocation of `std::cmatch` doesn't work for me due to an unknown reason (SEGFAULT) and I'm trying to debug that.  I thought that this is due to a missing initialization, but trying to declare it as "std::match re_md" on the stack gave me the same problem.

Please look at my branch `regex-mt` - maybe you will be able to see the problem.

ampli:
> I thought that this is due to a missing initialization

I found the c++ regex problem is because the buffer I allocated on the stack was not initialized. I will fix that.

ampli:
Branch `regex-mt`:
I fixed the c++ `cmatch` initialization.
Now the C++ and PCRE2 regex libraries pass my tests, but POSIX regex doesn't.

I will continue with it tomorrow.

linas:
Progress?

linas:
I put your regex-mt into #1360 and plant to merge that.

PR #1357, opened 2022-11-15
Use per-thread match descriptions

linas:
This performs the per-thread management of the match descriptions for the other two supported (but deprecated) regex libs.


linas:
@ampli I have done the shameful thing of not even trying to see if this code compiles, never mind runs or runs correctly. About this, I am going to be lazy. Therefore, I am not merging just yet. 

Would you have any comments, or care to make tests? e.g. Windows or Apple? This uses tss code which is part of the C11 standard, so I would hope it's available in Windows/Apple ... but ... those guys are always full of surprises.

linas:
Ooops! I unintetionally hit "merge" cause its late and I'm tired and it was reflexive. Dohh.

PR #1359, opened 2022-11-19
On-stack regex matching

linas:


ampli:
There is still a problem with it: It doesn't free the `pcre2_match_data` template on thread-exit.
I think I have a fix: Use a static buffer for the `pcre2_match_data` template (to be initialized on dict open).
The idea is that the initialized data is always the same, so it doesn't matter if on rare occasions a race happens.
So I would like to force-push a fix into this PR (this will take me another day to implement and test).

(There is a better way to do that, but we don't yet have the infrastructure for that.)

linas:
close; replace by #1360 ... somehow, github no longer supports manual 
resolve of conflicts!?

PR #1360, opened 2022-11-19
On-stack regex matching 

linas:
Handle regex match descriptions on stack. This is a version of 
https://github.com/ampli/link-grammar.git regex-mt
with master merged, an some minor fixes.

### DIFF
--- a/link-grammar/dict-common/regex-morph.c
+++ b/link-grammar/dict-common/regex-morph.c
@@ -10,8 +10,6 @@
 /*                                                                       */
 /*************************************************************************/
 
-#include <threads.h> /* C11 Concurrency library */
-
 /**
  * Support for the regular-expression based token matching system
  * using standard POSIX regex, PCRE2 or C++.
@@ -63,7 +61,6 @@ typedef struct {
 #if HAVE_REGEX_H
 typedef struct {
 	regex_t re_code;
-	unsigned int Nre_md;
 } reg_info;
 #endif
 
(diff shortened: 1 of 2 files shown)


link-grammar/dict-common/regex-morph.c
```diff
@@ -182,16 +193,12 @@ static bool reg_comp(Regex_node *rn)
 	                            options, &rc, &erroffset, NULL);
 	if (re->re_code != NULL)
 	{
+		call_once(&call_once_flag, alloc_match_data);
```
ampli, line 196: `#if HAVE_THREADS_H` is needed for emscripten.

ampli:
There is also a need to arrange for freeing the match data on thread exit...

link-grammar/dict-common/regex-morph.c
linas: Gahh. I'll add that right now.

linas: OK, I pushed directly to master.

BTW I reviewed the regular regex code, I see nothing wrong with it, so I don't see why it would be failing your tests.

linas:
> There is also a need to arrange for freeing the match data on thread exit...

Actually not; because there is only one copy, shared by all threads. I removed the TLS keyword you had placed there.

ampli:
> Actually not

Of course you are right. (I used TLS in order to avoid `call_once()`, but then I encountered a memory leak.)

link-grammar/dict-common/regex-morph.c
ampli: > so I don't see why it would be failing your tests.

It was my mistake - I forgot the default any/amy/ady dicts doesn't work fine with the POSIX regex library.

linas:
Anyway, thanks for doing this.  I'd like to issue a version 5.12.0 in a few days or a week.

And totally off-topic, if you want to waste time, someone aimed me at some amateur sci-fi; its a Vernor Vinge knockoff. I'm half-way through; it's well written.
https://borretti.me/fiction/eog581

PR #1363, opened 2022-12-21
regex-morph.c: add ifdef with HAVE_THREADS_H

equal-l2:
This `#include` has been added after #1178, so it is not covered in that PR.


linas:
Thanks!

PR #1368, opened 2023-01-13
pcre2 malloc/free per call

### LINKED ISSUE #1366: incorrect use of pcre2_match_data would result in severe memory leak
A `pcre2_match_data` is designed to be allocated and free when no longer used, which is why is an opaque type and the API doesn't provide a "copy" method.

The following incomplete (will need at least syntax changes to support MSVC) code, changes the way the reusing of pcre2_match_data was done, to at least ensure it was really reusable within a thread and prevents `tests/multi-thread` to explode on size as a result of leaking frames on each match when using PCRE2 >= 10.41, but still doesn't do the necessary free.

```
$ git diff
diff --git a/link-grammar/dict-common/regex-morph.c b/link-grammar/dict-common/regex-morph.c
index 4ca172f3f..1e90036c9 100644
--- a/link-grammar/dict-common/regex-morph.c
+++ b/link-grammar/dict-common/regex-morph.c
@@ -167,22 +167,9 @@ static void reg_free(Regex_node *rn)
 #if HAVE_PCRE2_H
 
 #define ALLOCTE_MATCH_DATA(var) \
-       pcre2_match_data *var = \
-       memcpy(alloca(re_md_template_size), re_md_template, re_md_template_size)
-
-static unsigned int re_md_template_size;
-static pcre2_match_data* re_md_template;
-
-#if HAVE_THREADS_H
-static once_flag call_once_flag = ONCE_FLAG_INIT;
-#endif // HAVE_THREADS_H
-static void alloc_match_data(void)
-{
-       re_md_template = pcre2_match_data_create(MAX_CAPTURE_GROUPS, NULL);
-       if (re_md_template == NULL)
-               return;
-       re_md_template_size = pcre2_get_match_data_size(re_md_template);
-}
+       static __thread pcre2_match_data *var; \
+       if (!var) \
+               var = pcre2_match_data_create(MAX_CAPTURE_GROUPS, NULL);
 
 static bool reg_comp(Regex_node *rn)
 {
@@ -196,20 +183,7 @@ static bool reg_comp(Regex_node *rn)
        re->re_code = pcre2_compile((PCRE2_SPTR)rn->pattern, PCRE2_ZERO_TERMINATED,
                                    options, &rc, &erroffset, NULL);
        if (re->re_code != NULL)
-       {
-#if HAVE_THREADS_H
-               call_once(&call_once_flag, alloc_match_data);
-#else
-               alloc_match_data();
-#endif
-               if (0 == re_md_template_size)
-               {
-                       prt_error("Error: pcre2_match_data_create() failed\n");
-                       free(re);
-                       return false;
-               }
                return true;
-       }
 
        /* We have an error. */
        PCRE2_UCHAR errbuf[ERRBUFFLEN];
```

linas:
Alternate attempt to fix #1366 -- malloc and free each call.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 Version 5.12.1 (XXX 2023)
  * Assorted enhancements and fixes for the AtomSpace dictionary. #1362
  * Fix missing HAVE_THREADS_H, affects Apple Homebrew. #1363
+ * Fix pcre2 version 10.41 memleak. #1367
 
 Version 5.12.0 (26 Nov 2022)
  * Fix crash when using the Atomese dictionary backend.
--- a/link-grammar/dict-common/regex-morph.c
+++ b/link-grammar/dict-common/regex-morph.c
@@ -87,6 +87,7 @@ typedef struct {
 
 #define ALLOCTE_MATCH_DATA(var) \
 	regmatch_t mem_##var[MAX_CAPTURE_GROUPS]; regmatch_t *var = mem_##var
+#define FRE_MATCH_DATA(var)
 
 /**
  * Compile the given regex..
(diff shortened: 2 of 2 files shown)


linas:
Not needed.

PR #1369, opened 2023-01-13
configure: improve detection of pcre2

carenas:
Under the current checks, a possibly conflicting libpcre2 devel package would need to be installed to allow for both the pcre2.h header checks to succeed.

The proposed change simplifies testing with an uninstalled, or development version of pcre2, without having to set both PCRE2 and REGEX environment variables, and still detects the system package if no override is needed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -706,14 +706,10 @@ case "$with_regexlib" in
 	'')
 		break
 		;;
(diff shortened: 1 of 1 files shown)


linas:
Looks great! Thanks!

Issue #1374, opened 2023-01-15
Emscripten build fails


ampli:
In recent CI checks, in the Emscripten configuration LT_INIT checks for `threads.h`, as can be seen from the following output (still exists in PR #1365 from Jan 13 2023):
``` text
checking for sysroot... no
checking for a working dd... /usr/bin/dd
checking how to truncate binary pipes... /usr/bin/dd bs=4096 count=1
checking for mt... mt
checking if mt is a manifest tool... no
checking for stdio.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for strings.h... yes
checking for sys/stat.h... yes
checking for sys/types.h... yes
checking for unistd.h... yes
checking for vfork.h... no
checking for threads.h... yes
checking for dlfcn.h... yes
checking for objdir... .libs
```

Note the check for `threads.h`. It defines `HAVE_THREADS_H`, which protects the threads stuff that shouldn't be used if Emscripten is configured. This happens in my development system too.

It seems this is something new, as in the CI output for the Emscription configuration in PR #1352 (5 Nov 2022) `threads.h` doesn't appear:
``` text
checking for sysroot... no
checking for a working dd... /usr/bin/dd
checking how to truncate binary pipes... /usr/bin/dd bs=4096 count=1
checking for mt... mt
checking if mt is a manifest tool... no
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking for dlfcn.h... yes
checking for objdir... .libs
```
Also note that in the older output there is the message "checking for ANSI C header files." instead of the check for `stdio.h`.
This Emscription configuration doesn't.

I thought this could be due to a recent change in `configure.ac`, that removed or reordered something that changed the behaviour of `LT_INIT`. However, I couldn't find such a change, and when I tried to configure Emscripten in my system using old branches, `LT_INIT` there also detected `threads.h`.

I also couldn't find any reference that checking for `threads.h` is something new in `LT_INIT`. So this is a mystery for me.

I can think of two ways to fix this problem:
[...]

Issue #1361, opened 2022-11-26
insane ./configure in github circleci


linas:
Github seems to be using two different sets of autoconf macros in different instances. The one on this git repo is "sane" and seems to be the same as what I've got on my development machine. The one on `linas/link-grammar` seems to be insane.  As a result, circleci fails on linas/link-grammar but works here.

Working configure output:
```
72 checking for suffix of object files... o
73 checking whether we are using the GNU C compiler... yes
74 ...
75 checking for /home/runner/work/link-grammar/link-grammar/bindings/js/emsdk/fastcomp/emscripten/emcc option to accept ISO C89... none needed
```

Failing configure output:
```
72 checking for suffix of object files... o
73 checking whether the compiler supports GNU C... yes
74 ...
75 checking for /home/runner/work/link-grammar/link-grammar/bindings/js/emsdk/fastcomp/emscripten/emcc option to enable C11 features... none needed
```

First diff is at line 73: my configure has the message "checking whether we are using the GNU C compiler" at line 4193 of the configure script.  It has "option to accept ISO C89" at line 4308

My configure script does not have the message "checking whether the compiler supports GNU C" anywhere. It also doesn't have "option to enable C11 features" anywhere.

That means .. `autogen.sh` is producing two different configures on these two instances.  WTF.

linas:
Using commit 08b21f851ce9f12736bb5ae1ecedaef45da2b99a

I'm able to verify the problem is in `AC_PROG_CC`.  The first echo is at line 3615 and the second is at line 4655 which bracket the problem area.

linas:
Failing test: 

https://github.com/linas/link-grammar/actions/runs/3551719391/jobs/5966133663

linas:
Hmm. Today, the failing configure is getting used on the main repo.

linas:
Hi Alex,

If you create a CMake infrastructure, that wold be fine, and I'd merge the pull reqs. It's used in all the other parts of OpenCog, so that's a plus.

As to Emscripten & JavaScript in general; there is no one out there that is using the javascript bindings. I think these mostly more-or-less work, and the issue is with github itself (I guess) Every now and then, Amir & I make a token effort to fix this up, but without some javascript user insisting on something, not much happens.

PR #1375, opened 2023-01-16
Fix the thread code inclusion problem in Emscripten builds

ampli:
See issue #1374 .

Remove the Emscripten check from `configure` and instead check `__EMSCRIPTEN__`  in the source code.
(For some reason this didn't work for me when I previously handled this problem.)
BTW, `AX_TLS` ` already takes care not to define `TLS`.

In addition, git-ignore the file `a.wasm` that is left after `emconfigure ./configure`.

Note: This error occurs inside an autoconf macro. I didn't try to find out why.
``` text
checking if the linker (/usr/local/src/emsdk/upstream/emscripten/emcc) is GNU ld... yes
em++: error: no input files
checking whether the /usr/local/src/emsdk/upstream/emscripten/em++ linker (/usr/local/src/emsdk/upstream/emscripten/emcc) supports shared libraries... yes
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ data/th/old_words/**
 # Visual Studio keeps adding the directory ".vs". If it got added while
 # rebasing and it is not git-ignored, it somehow breaks the rebase.
 .vs
+# a.wasm is created by "emcc" (in "emconfigure ./configure").
+# For now just ignore it.
+a.wasm
--- a/configure.ac
+++ b/configure.ac
@@ -119,37 +119,27 @@ AC_CHECK_FUNCS(prctl)
 AC_DEFINE(__STDC_FORMAT_MACROS)
 AC_DEFINE(__STDC_LIMIT_MACROS)
 
-# Don't use thread functions with Emscripten.
-# Emscripten does not like threads.
-if test "$EMSCRIPTEN" != ''; then
-	AC_CHECK_DEFINE(__EMSCRIPTEN__, [emscripten=yes],
-						 [AC_MSG_ERROR([EMSCRIPTEN environment variable found, but emcc is not used])])
-fi
+# Check for a keyword for Thread Local Storage declaration.
+# If found - define TLS to it.
+lg_tls=no
+AX_TLS(,:)
+test "$ac_cv_tls" != "none" && lg_tls=yes
 
-if test "x$emscripten" != "xyes"; then
+AX_PTHREAD
+if test -n "$ax_pthread_ok"; then
+	CC="${PTHREAD_CC:-CC}"
+	dnl #CCX="${PTHREAD_CC:-CCX}"
 
-	# Check for a keyword for Thread Local Storage declaration.
-	# If found - define TLS to it.
-	lg_tls=no
-	AX_TLS(,:)
-	test "$ac_cv_tls" != "none" && lg_tls=yes
-
-	AX_PTHREAD
-	if test -n "$ax_pthread_ok"; then
-		CC="${PTHREAD_CC:-CC}"
-		dnl #CCX="${PTHREAD_CC:-CCX}"
-
-		dnl Currently, libtool uses --nostdlib when linking shared libraries,
-		dnl which causes the -pthread flag to be ignored. Try to work around that
-		dnl by explicitly specifying the pthread library unless it is already set.
-		if test -z "$PTHREAD_LIBS"; then
-			PTHREAD_LIBS=-lpthread
-		fi
+	dnl Currently, libtool uses --nostdlib when linking shared libraries,
+	dnl which causes the -pthread flag to be ignored. Try to work around that
+	dnl by explicitly specifying the pthread library unless it is already set.
+	if test -z "$PTHREAD_LIBS"; then
+		PTHREAD_LIBS=-lpthread
 	fi
-
-	dnl Check if we can use C11 threads functions
-	AC_CHECK_HEADERS_ONCE([threads.h])
 fi
+
+dnl Check if we can use C11 threads functions
+AC_CHECK_HEADERS_ONCE([threads.h])
 
 dnl If the visibility __attribute__  is supported, define HAVE_VISIBILITY
 dnl and a variable CFLAG_VISIBILITY, to be added to CFLAGS/CXXFLAGS.
--- a/link-grammar/dict-common/regex-morph.c
+++ b/link-grammar/dict-common/regex-morph.c
@@ -10,9 +10,9 @@
 /*                                                                       */
 /*************************************************************************/
 
-#if HAVE_THREADS_H
+#if HAVE_THREADS_H && !__EMSCRIPTEN__
 #include <threads.h>
-#endif
+#endif // HAVE_THREADS_H && !__EMSCRIPTEN__
 
 /**
  * Support for the regular-expression based token matching system
@@ -177,7 +177,7 @@ static void alloc_key(void)
 
 static pcre2_match_data* alloc_match_data(void)
 {
-#if HAVE_THREADS_H
+#if HAVE_THREADS_H && !__EMSCRIPTEN__
 	call_once(&call_once_flag, alloc_key);
 
 	pcre2_match_data *pmd = (pcre2_match_data *) tss_get(re_md_key);
@@ -194,7 +194,7 @@ static pcre2_match_data* alloc_match_data(void)
 	pmd = pcre2_match_data_create(MAX_CAPTURE_GROUPS, NULL);
 	if (pmd) return pmd;
 
-#endif // HAVE_THREADS_H
+#endif // HAVE_THREADS_H && !__EMSCRIPTEN__
 	prt_error("Error: pcre2_match_data_create() failed\n");
 	return NULL;
 }
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -14,9 +14,9 @@
 
 #include <limits.h>
 #include <inttypes.h>                   // PRIu64
-#if HAVE_THREADS_H
+#if HAVE_THREADS_H && !__EMSCRIPTEN__
 #include <threads.h>
-#endif /* HAVE_THREADS_H */
+#endif /* HAVE_THREADS_H && !__EMSCRIPTEN__ */
 
 #include "link-includes.h"
 #include "api-structures.h"
@@ -124,7 +124,7 @@ struct count_context_s
 #define MAX_TABLE_SIZE(s) (s / 10) /* Low load factor, for speed */
 #define MAX_LOG2_TABLE_SIZE ((sizeof(size_t)==4) ? 25 : 34)
 
-#if HAVE_THREADS_H
+#if HAVE_THREADS_H && !__EMSCRIPTEN__
 /* Each thread will get it's own version of the `kept_table`.
  * If the program creates zillions of threads, then there will
  * be a mem-leak if this table is not released when each thread
@@ -147,7 +147,7 @@ static void make_key(void)
 {
 	tss_create(&key, free_tls_table);
 }
-#endif /* HAVE_THREADS_H */
+#endif /* HAVE_THREADS_H && !__EMSCRIPTEN__ */
 
 /**
  * Allocate memory for the connector-pair table and initialize table-size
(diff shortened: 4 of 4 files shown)


ampli:
It now fails on this:
`wasm-ld: error: 'bulk-memory' feature must be used in order to emit passive segments`
This doesn't happen on my computer (I'm using the latest Emscripten from GitHub.)

Which version is getting used in CI?

linas:
I'll wait a bit before merging this...

ampli:
> I'll wait a bit before merging this...

Strangely, `emmake make` succeeds on my computer.
I will make changes in this PR to investigate the problem further.

linas:
Well, the other issue #1361  is like this too: works n my machine, fails in circle-ci. ... and more: when I first spotted it, it seemed there were two different versions of emscripten in circle ci, one passed, one failed... Hm. Maye need to look at the ci config file more carefully?

linas:
I looked at `.github/workflows/bindings-js.yml`  I see nothing suspicious there. Don't know how to pick an install version

Try adding `--features=bulk-memory` to `bindings/js/link-parser/build.sh`  -- that might solve the error message.

ampli:
From the CI log:
>/home/runner/work/link-grammar/link-grammar/bindings/js/emsdk/upstream/emscripten/emcc.py:985: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  newargs = [a for a in newargs if a is not '']

My emcc.py, as I got from `git` clone https://github.com/emscripten-core/emsdk.git` doesn't contain such a line.
Still investigating.

ampli:
https://github.com/opencog/link-grammar/blob/43d239259148d923a5ce170f9fbd01385bfee65c/bindings/js/install_emsdk.sh#L1-L13
For some reason it installs `emsdk` version `1.38.41`, which is one of the firs versions. The latest version, as installed by `emsdk install latest`,  is `3.1.30`. I don't know why a fixed version got used in the first place.

With `3.1.30`, `emmake make` runs with no errors.
However, I do't know how to check its result.
It seems there is  some bitrot in `build.sh`:
``` text
$ ./build.sh                                                                                                                                                                                                                                          LG (em-detect $%|u+6-3)
++ dirname ./build.sh
+ pushd .
/usr/local/src/link-grammar-devel/em-detect/bindings/js/link-parser /usr/local/src/link-grammar-devel/em-detect/bindings/js/link-parser
+ '[' '!' /usr/local/src/emsdk ']'
+ rm -rf dist
+ mkdir dist
+ cp ../../../link-parser/.libs/link-parser link-parser.bc
+ emcc -O3 link-parser.bc ../../../link-grammar/.libs/liblink-grammar.a --pre-js pre.js -s WASM=1 -s ALLOW_MEMORY_GROWTH=1 -o dist/link-parser.js
wasm-ld: error: unknown file type: link-parser.bc
```
And indeed `../../../link-parser/.libs/link-parser` is not a bytecode file.

linas:
> For some reason it installs emsdk version 1.38.41, which is one of the firs versions. 

The js bindings are 10-15 years old :-/

linas:
I guess you tried this:
```
./emsdk install latest
./emsdk activate latest
```

linas:
`link-parser/.libs/link-parser` is an ascii shell script created by autotools. In the same directory there is a `link-parser.wasm`.   When I `cp link-parser.wasm link-parser.bc` I get

wasm-ld: error: link-parser.bc: not a relocatable wasm file

ampli:
> `link-parser/.libs/link-parser` is an ascii shell script created by autotools. In the same directory there is a `link-parser.wasm`. When I `cp link-parser.wasm link-parser.bc` I get
> 
> ```
> wasm-ld: error: link-parser.bc: not a relocatable wasm file
> ```

I get this error too. I try to trace how it is created, in order to find out how to create a relocatable result (in a hope it will work).

linas:
> how to create a relocatable result

https://groups.google.com/g/emscripten-discuss/c/cwg_EdFYOr8/m/xLpVqi9VAQAJ?pli=1

linas:
That seems to build. I have no idea if the result is usable or not.  However, look at the debug output here: https://github.com/linas/link-grammar/actions/runs/3952974168/jobs/6768705348  -- this is from trying to uncomment the line 
```
emcc -O3 ../../link-grammar/.libs/liblink-grammar.so -o liblink-grammar.js
```
and its .. well, like the threading wasn't suppressed after all.

linas:
Except the issue is not about threading, its about duplicate symbols.  So I guess the library should not have been linked. Whatever. I'm going to comment out that line, and just punt.  No one is using link-grammar js anyway, and given how things are, it seems unlikely that anyone ever will ...

ampli:
It seems that by now Emscripten does support Pthreads:

From https://emscripten.org/docs/api_reference/wasm_workers.html:
``` text
Emscripten supports two multithreading APIs to leverage this web feature:
POSIX Threads (Pthreads) API, and

Wasm Workers API.
```

linas:
Well, its still erroring out see #1377  I'm done with this for now. If you want to work on this and create a new/different pull req, please do; otherwise I will merge #1377 or a variant in a day or two.

ampli:
Here is how I ran it using Chrome:
1. Ignore `build.sh`.
2. Currently, the last step of link-parser/Makefile for creating link-parser is:
`
/usr/local/src/emsdk/upstream/emscripten/emcc -Wstrict-prototypes -Wmissing-prototypes -Wnested-externs -Wold-style-definition -Werror-implicit-function-declaration -Wall -Wextra -Wsign-compare -Wpointer-arith -Wwrite-strings -Wmissing-declarations -Wpacked -Wswitch-enum -Wmissing-format-attribute -Wstrict-aliasing -Winit-self -Wshadow -Wno-missing-field-initializers -Wno-unused-parameter -Wno-attributes -Wno-long-long -Winline -D_DEFAULT_SOURCE -std=c11 -D_BSD_SOURCE -D_SVID_SOURCE -D_GNU_SOURCE -D_ISOC11_SOURCE -fvisibility=hidden -g -O3 -o .libs/link-parser link_parser-link-parser.o link_parser-command-line.o link_parser-lg_readline.o link_parser-parser-utilities.o  ../link-grammar/.libs/liblink-grammar.a  -lm
`
Note the `-o .libs/link-parser` in it. For my text, I manually changed it to `-o .libs/link-parser.html`.
3. Due to Cross-Origin Requests limitations, it must be loaded through a web server.
``` text
cd link-parser/.libs
python -m http.server
```
4. In chrome:
localhost:8000/link-parser.html

When I open the JS console I see:
``` text
Uncaught RuntimeError: memory access out of bounds
    at dlfree (link-parser.wasm:0x435f7)
    at object_open (link-parser.wasm:0x7193)
    at dictionary_create_default_lang (link-parser.wasm:0x57f6)
    at main (link-parser.wasm:0xb50)
    at Module._main (link-parser.js:3779:72)
    at callMain (link-parser.js:3841:15)
    at doRun (link-parser.js:3884:23)
    at link-parser.js:3895:7
```


I also tried to run it using `node'.
1. Again ignore `build.sh`.
2. Replace `-o .libs/link-parser` by `-o .libs/link-parser.js`.
3. `cd link-parser/.libs`
4.A. `node link-parser.js` # it uses `node` version `v14.18.2` from the `emsdk` repository.
It fails:
``` text
failed to asynchronously prepare wasm: CompileError: WebAssembly.instantiate(): Compiling function #316:"form_match_list" failed: Invalid opcode (enable with --experimental-wasm-threads) @+78470
...
```
4.B. `/usr/bin/node link-parser.js` # it uses `node` version `v18.12.1`.
It fails in the same point that it failed in the browsew:
``` text
/usr/local/src/link-grammar-devel/em-detect/link-parser/.libs/link-parser.js:135
      throw ex;
      ^

RuntimeError: memory access out of bounds
    at dlfree (wasm://wasm/006f930e:wasm-function[892]:0x435f7)
    at object_open (wasm://wasm/006f930e:wasm-function[153]:0x7193)
    at dictionary_create_lang (wasm://wasm/006f930e:wasm-function[128]:0x585e)
    at main (wasm://wasm/006f930e:wasm-function[25]:0xb39)
...
```

So `link-parser` actually runs but encounters a run-time error in `object_open()`. I will try to debug it.

linas:
> python -m http.server
> localhost:8000/link-parser.html

These instructions belong in a README

ampli:
It turned out the TLS variable got defined to `_Thread_local` also under the Emscripten configuration.
I addition, I found that under Emscripten, TLS variables are not initialized by default to 0 or NULL.
So `free(pf)` in `object_open()` tried to free a random address. Fixed by defining TLS to nothing by `#if __EMSCRPTEN__` in `utilities.h`.

Also, it is possible to replace the `emcc` call in `biuld.sh` by:
`(cd ../../../; emmake make EXEEXT=.html LDFLAGS="--pre-js $PWD/bindings/js/link-parser/pre.js -s WASM=1 -s ALLOW_MEMORY_GROWTH=1")`

I still haven't found how the following call in `build.sh` is intended to work:
`echo The needs of the many outweigh the needs of the few. | link-parser`

Unless I miss something, there is not supposed to be `link-parser` through `PATH` at this point.
In addition, trying to invoke `link-parser/link-parser` gives this error:
`link-parser/link-parser: line 117: exec: ... link-parser/.libs/lt-link-parser: cannot execute: Permission denied`
But `node ... link-parser/.libs/lt-link-parser` works!
So what is the idea here?

Also note that the emdsk's bundled `node.js`, which is an old version, didn't work for me at all (opcode error). But it is in the PATH by default due to `emsdk_env.sh`. I have no idea why it is bundle, and how to cleanly overcome this problem.

linas:
> Permission denied

In the normal, non emcc world,  the `link-parser/link-parser` is a shell script that forces the real executable `link-parser/.libs/lt-link-parser` to link to the shared libraries in the build dir, instead of the shared libs in `/usr/lib`. So `file link-parser/.libs/lt-link-parser` just says its a normal ELF executable.

For the emcc case, I assume  `file link-parser/.libs/lt-link-parser` says its its a WebAssembly  module, which would be normally executable in Linux, and the ascii script gets tangled up. 

For me, `node` is `/usr/bin/node` from `dpkg -S /usr/bin/node` says its `nodejs` and `dpkg -l nodejs` says its version `16.17.0-deb-1nodesource1`

So I assume if you install nodejs, then the emsdk bundled version will be ignored.

ampli:
> So I assume if you install nodejs, then the emsdk-bundled version will be ignored.

https://github.com/opencog/link-grammar/blob/750315dfa754f3d7b8ab29bc9dce85e9401e4dcb/bindings/js/link-parser/build.sh#L1-L36
Line 7  sets the path the `emsdk`'s bundled nodejs. This is the output of `emsdk_env.sh`:
``` text
Setting up EMSDK environment (suppress these messages with EMSDK_QUIET=1)
Adding directories to PATH:
PATH += /usr/local/src/emsdk
PATH += /usr/local/src/emsdk/upstream/emscripten
PATH += /usr/local/src/emsdk/node/14.18.2_64bit/bin

Setting environment variables:
PATH = /usr/local/src/emsdk:/usr/local/src/emsdk/upstream/emscripten:/usr/local/src/emsdk/node/14.18.2_64bit/bin:/home/amir/bin:/usr/share/Modules/bin:/usr/lib64/ccache:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/var/lib/snapd/snap/bin
EMSDK = /usr/local/src/emsdk
EMSDK_NODE = /usr/local/src/emsdk/node/14.18.2_64bit/bin/node
```
Note the 3rd component in PATH:
`/usr/local/src/emsdk/node/14.18.2_64bit/bin`
This old node version doesn't work with `link-parser.wasm` (aborts on invalid opcode). I needed to edit-out the provided `emsdk/node` PATH component so my `/usr/bin/node` would be used (it is version v18.12.1).

Regarding `link-parser/link-parser`:  It exec's `lt-link-parser`, which is a JS file. If I start it with
`#!/usr/bin/env node`
and do `chmod +x` on it, then `link-parser/link-parser` works!
So my what I don't understand is why `emcc` doesn't add `#!#!/usr/bin/env node` and makes the file executable. Maybe there is a way to cause it to do it.

Regarding the test in `build.sh`:
https://github.com/opencog/link-grammar/blob/750315dfa754f3d7b8ab29bc9dce85e9401e4dcb/bindings/js/link-parser/build.sh#L22-L34
It invokes `link-parser`, but which searches it in the `PATH` , and there is normally no `.` in it. So it uses an installed one (if any). In addition, line 32 performs `popd`, so `link-parser` doesn't run inside the `dist` directory. So how it is supposed toi find the `data` directory?
In addition, I don't understand the use of `npm link`. It tries to link to a system directory, which is not supposed to already contain `link-grammar` (and in any case fails for me because I don't run it as root). I think it should be tested without npm.

ampli:
I said:
> It invokes `link-parser`, but which searches it in the `PATH` , and there is normally no `.`

I am now reading the npm docs in order to understand how all of that is supposed to work.
If needed, I will rewrite `build.sh`.

linas:
Should I merge #1377? Or should I wait?

ampli:
> Should I merge #1377? Or should I wait?

Better to wait since I'm touching the same  code.
I also try to do the things in the autoconf way.

PR #1376, opened 2023-01-17
Disable aspell by default.

linas:
Apparently, aspell leaks memory. See discussion #1373 for details.


linas:
@ampli could you review/comment on this? Am I missing something?

linas:
I'm impatient; merging

PR #1377, opened 2023-01-18
Hacks to try to get javascript build to not spew errors

linas:
After experimentation, this seems to ... build without error.  No idea if the result is usable, or not.


ampli:
I'll return to that after finishing handling the memory-reduction fixes.

linas:
I merged because I got tired of seeing the failing-build error.

PR #1378, opened 2023-01-21
Deduplicate linkages

linas:
When linkages are selected randomly, it can happen that many of them are duplicates.  When the size of the alloced linkage array is just under the number found, as many as half can be duplicates!

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@ Version 5.12.1 (XXX 2023)
  * Fix missing HAVE_THREADS_H, affects Apple Homebrew. #1363
  * Fix regex thread-safety issue. #1370
  * Disable aspell; it leaks memory. #1373
+ * Deduplicate linkages when overflow forces random selection. #1378
 
 Version 5.12.0 (26 Nov 2022)
  * Fix crash when using the Atomese dictionary backend.
--- a/link-grammar/api-structures.h
+++ b/link-grammar/api-structures.h
@@ -162,6 +162,7 @@ struct Sentence_s
 	/* Parse results */
 	int    num_linkages_found;  /* Total number before postprocessing.  This
 	                               is returned by the do_count() function */
+	bool   overflowed;          /* True, if counting overflowed. */
 	size_t num_linkages_alloced;/* Total number of linkages allocated.
 	                               the number post-processed might be fewer
 	                               because some are non-canonical */
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -786,7 +786,7 @@ static void list_random_links(Linkage lkg, unsigned int *rand_state,
 
 	if (set == NULL || set->first == NULL) return;
 
-	/* Most of the times there is only one list element. */
+	/* Most of the time, there is only one list element. */
 	if (set->first->next == NULL)
 	{
 		pc = set->first;
--- a/link-grammar/parse/parse.c
+++ b/link-grammar/parse/parse.c
@@ -64,15 +64,15 @@ static void find_unused_disjuncts(Sentence sent, extractor_t *pex)
 	}
 }
 
-static bool setup_linkages(Sentence sent, extractor_t* pex,
+static void setup_linkages(Sentence sent, extractor_t* pex,
                           fast_matcher_t* mchxt,
                           count_context_t* ctxt,
                           Parse_Options opts)
 {
-	bool overflowed = build_parse_set(pex, sent, mchxt, ctxt, sent->null_count, opts);
+	sent->overflowed = build_parse_set(pex, sent, mchxt, ctxt, sent->null_count, opts);
 	print_time(opts, "Built parse set");
 
-	if (overflowed && (1 < opts->verbosity) && !IS_GENERATION(sent->dict))
+	if (sent->overflowed && (1 < opts->verbosity) && !IS_GENERATION(sent->dict))
 	{
 		err_ctxt ec = { sent };
 		err_msgc(&ec, lg_Warn, "Count overflow.\n"
@@ -86,7 +86,7 @@ static bool setup_linkages(Sentence sent, extractor_t* pex,
 		sent->num_linkages_post_processed = 0;
 		sent->num_valid_linkages = 0;
 		sent->lnkages = NULL;
-		return overflowed;
+		return;
 	}
 
 	sent->num_linkages_alloced =
@@ -98,8 +98,6 @@ static bool setup_linkages(Sentence sent, extractor_t* pex,
 	 * XXX free_linkages() zeros sent->num_linkages_found. */
 	if (sent->lnkages) free_linkages(sent);
 	sent->lnkages = linkage_array_new(sent->num_linkages_alloced);
-
-	return overflowed;
 }
 
 /**
@@ -166,13 +164,13 @@ static bool optional_word_exists(Sentence sent)
  * linkages.
  */
 static void process_linkages(Sentence sent, extractor_t* pex,
-                             bool overflowed, Parse_Options opts)
+                             Parse_Options opts)
 {
 	if (0 == sent->num_linkages_found) return;
 	if (0 == sent->num_linkages_alloced) return; /* Avoid a later crash. */
 
 	/* Pick random linkages if we get more than what was asked for. */
-	bool pick_randomly = overflowed ||
+	bool pick_randomly = sent->overflowed ||
 	    (sent->num_linkages_found > (int) sent->num_linkages_alloced);
 
 	sent->num_valid_linkages = 0;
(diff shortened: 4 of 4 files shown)


linas:
FYI @ampli, I think we talked about this before, I am not sure where.  This was the only solution I could find; trying to avoid duplicates during random linkage generation seemed to be a hopeless task. If you don't like what I did, let me know :-)

ampli:
Some comments:

1. It seems possible to avoid duplicates in the random selection, by leaving marks in the parse set tree.
This has the advantage of producing  as close as possible number of linkages to the requested number.
However, the code is likely to be more complex (I cannot estimate its efficiency).

2. In `deduplicate_linkages()`, there is a `memmove()` call for each duplicate. In case of many duplicates, it would be more efficient to first mark the duplicates and then do `memove()` a few times as possible (or to copy to a new linkage array).

3. The implemented duplicate detection is not guaranteed to find all the duplications, since the sorting is according  the linkage cost, and similar cost linkages are left in their original order. It is possible to solve that (among other ways) by a final sort criterion according to the link array (and then an identity mark can be applied in the sort comparison function).
This has the benefit that if done on any sort (not only sorts of random selection linkages) than the lsame-cost linkages will always be in a "canonical" order.

4. The problem of duplicate linkages is made more severe due to insane (should the term be unsane?) linkages. If there are a significant portion of such linkages, the number of linkages may more easily overflow a given linkage limit.
I have a patch of adding an alternative-compatibility check in `do_count()` that reduces the number of insane linkages and thus mitigates this problem .I still need to convert it to a PR. (Eliminating absolutely all the alternative-compatibility in `do_count()` seems to be too costly.)

I will add one more comment in the code context.

ampli:
Commit 706401e3b introduces a memory leak:
``` text
==3401967==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 15872 byte(s) in 4 object(s) allocated from:
    #0 0x7f6f3c6ba097 in calloc (/lib64/libasan.so.8+0xba097)
    #1 0x7f6f391acb36 in partial_init_linkage ../../../../usr/local/src/link-grammar-devel/tests/link-grammar/linkage/freeli.c:60

Direct leak of 2016 byte(s) in 4 object(s) allocated from:
    #0 0x7f6f3c6b95b5  (/lib64/libasan.so.8+0xb95b5)
    #1 0x7f6f391d77ef in gwordlist_append ../../../../usr/local/src/link-grammar-devel/tests/link-grammar/tokenize/wordgraph.c:46

Direct leak of 1984 byte(s) in 4 object(s) allocated from:
    #0 0x7f6f3c6ba6af in __interceptor_malloc (/lib64/libasan.so.8+0xba6af)
    #1 0x7f6f391d8be8 in exalloc ../../../../usr/local/src/link-grammar-devel/tests/link-grammar/utilities.c:483

SUMMARY: AddressSanitizer: 19872 byte(s) leaked in 12 allocation(s).
```

ampli:
I said:
> 3\. always be in a "canonical" order

It turned out this will not be so if it would depend on Connector pointer comparison (like now) -  the order may be different on different runs, while the goal for a canonical order is that it will even be stable over releases. Comparing `Connector:desc` is needed for that  -it still depends on connector name stability, but adding/removing connector names from the dict (or even changing their cost) may change the order of the linkages anyway.

ampli:
The memory leak happens because the duplicate linkage is not getting freed before it is overwritten.
https://github.com/opencog/link-grammar/blob/0e392a664f289d0c900b95514cc26374256a91ef/link-grammar/parse/parse.c#L326-L328
This line should be added before line 328 (`memmove()`):
`free_linkage(lpv);`

ampli:
Another problem is that the deduplication is NOT done only for sentences for which random selection is done.
https://github.com/opencog/link-grammar/blob/0e392a664f289d0c900b95514cc26374256a91ef/link-grammar/parse/parse.c#L297-L300
This is the comparison in line 299 is not  correct at this time.
It was correct in `process_linkages()` even though it looks the same:
https://github.com/opencog/link-grammar/blob/0e392a664f289d0c900b95514cc26374256a91ef/link-grammar/parse/parse.c#L173-L176
This is because `sent->num_linkages_alloced` may get reduced at the end of `process_linkages()`:
https://github.com/opencog/link-grammar/blob/0e392a664f289d0c900b95514cc26374256a91ef/link-grammar/parse/parse.c#L271-L273

In order to prevent this confusion, I propose to use `opts->linkage_limit` in both places (there is a need to add `opts` as a second parameter to `deduplicate_linkages()`.

ampli:
I said:
>Another problem is that the deduplication is NOT done only for sentences for which random selection is done.

Here is an example sentence that demonstrates the bug mentioned above, and also an additional bug.

`Alfred Baird, formerly vice president of Beevil Corp., has been appointed as president`

It has 120 linkages (you can see that with -verbosity=2), 80 of them w/o P.P. violations (you can see both of these numbers with -verbosity=4). But due to the incorrect comparison to `sent->num_linkages_alloced`, it is passed through the duplicate elimination of `deduplicate_linkages()` and only 40 linkages are reported.

Because this sentence got duplicate elimination due to this bug, I tried to use it as a test case to see if the duplicate elimination is correct. As far as I see for now (going to sleep soon...) it is incorrect.

First, you can see a problem if you use `link-parser lg -test=min-len-encoding:N`, when `N` is either 0 or 254.
For this sentence, you will get 80 and 40 linkages, respectively.
With no connector encoding (`min-len-encoding:N` means only sentences with less than `N` tokens are connector-encoded), all the connector addresses are different. With `min-len-encoding:254`, all sentences are connector-encoded by `pack_connectors()`, which also share tracon memory. It means connector addresses are reused, and most of the connectors don't take memory space (this speeds up the parsing of long sentences by order of magnitude).
It means that connectors on different disjuncts (of the same word only) may have the same addresses. **I still don't know why this confuses `deduplicate_linkages()`.**

To see how many duplicates the parsing of this sentence actually has, I produced all its linkages and tried to find out how many are unique. I have a Perl program to do that. It says 48 are unique. This contradicts the result of  `deduplicate_linkages()`. Moreover, on the result of  `deduplicate_linkages()`  it says 15 are actually duplicates. And indeed you can see, for example, that linkages 7 and 8 are identical (use `echo "..." | lg -test=auto-next-linkage`).

I still don't know why my checks apparently contradict `deduplicate_linkages()`.

linas:
I've fixed some of the bugs; pull req coming soon. I'm still digesting some of your comments...

> It has 120 linkages (you can see that with -verbosity=2), 80 of them w/o P.P. violations 

This is misleading; none of them had PP violations; there were 80 without morphism violations. We never even report morphism violations via the API, as, properly speaking those are due to mis-computation in our code. 

For this example sentence, I'm surprised that there are any unsane morphisms; what are they?

linas:
> Alfred Baird
> -test=min-len-encoding:N

After fixing assorted bad tests & etc the deduplication code for Baird is no longer entered. But if let dedupe run anyway, I get this: 
* No flags (default config) one duplicate found, the second linkage!
* `-test=min-len-encoding:0` same
* `-test=min-len-encoding:254`  no duplicates seen

The difference between the first and second linkages is that the first uses `vice.s` and the second uses `vice.n-u`  although the link types used are the same.  To be backwards-compatible, I will need to also check for mismatched words.

ampli:
> > It has 120 linkages (you can see that with -verbosity=2), 80 of them w/o P.P. violations
> 
> This is misleading; none of them had PP violations; there were 80 without morphism violations. We never even report morphism violations via the API, as, properly speaking those are due to mis-computation in our code.
 
Of course, you are right... The debug message actually says:
`link-grammar: Info: sane_morphism(): 40 of 120 linkages had invalid morphology construction`

> For this example sentence, I'm surprised that there are any unsane morphisms; what are they?

>> `Alfred Baird, formerly vice president of Beevil Corp., has been appointed as president`

``` text
$ link-parser -v=6 -debug=flatten_wordgraph,print_sentence_word_alternatives
...
Debug: Sentence words and alternatives:
...
  word8: Beevil
   alt0: Beevil
  word9: Corp.,
   alt0: Corp. {.} ,
   alt1: Corp
  word12: has
   alt0: has
...
```
For `Corp.` there are two alternatives:
`Corp.`
`Corp` `.`
(The debug message seems strange since it uses the flattening of the alternatives into the 2D array. So in slot 9, there are `Corp` and `Corp.` and in slot 10 there is an optional word `.` (slot 11 is `,`).

linas:
> there is an optional word `.`

I admit I no longer remember how morphology works...

ampli:
> > there is an optional word `.`

Optional words solve the problem of alternatives that have different numbers of tokens. Originally an "empty word" with ZZZ connectors was being used for balancing. The optional word trick is totally equivalent, but it has no need for adding/removing connectors. It seems it can be used for inserting words that may or may not be needed for parsing, but for that, an implementation of "hard priorities" is needed so the optional word will be eliminated from the parsing if there is a parse without it.

linas:
#1396 fixes most of the bugs. I'll do a distinct pull req for the sort issue.

linas:
#1397 fixes another issue, and improves deduplication by doing the compare during sorting. Many, many more duplicates are found!

linas:
I think I have addressed all of the above issues in pull req #1397.

Issue #1379, opened 2023-01-21
multi-threaded dictionary crash


linas:
Multi-threaded update of the dictionary is not thread-safe. This was always the case. We knew this all along. This didn't matter, when the dictionaries were just file-backed. But now that they can be dynamic objects, it is time to fix this.

I get this:
```
#3  0x00007ffff7e612fc in malloc_printerr (
    str=str@entry=0x7ffff7f85670 "double free or corruption (out)")
    at malloc.c:5347
#4  0x00007ffff7e62fa0 in _int_free (av=0x7ffff7fb8b80 <main_arena>, 
    p=0x7fff300aee10, have_lock=<optimized out>) at malloc.c:4314
#5  0x00007fffecb8cd95 in condesc_grow (ct=0x7fff4c00cfc8)
    at ../../link-grammar/connectors.c:539
#6  condesc_add (ct=ct@entry=0x7fff4c00cfc8, constring=0x7fff30091510 "MDB")
    at ../../link-grammar/connectors.c:558
#7  0x00007fffecba6d2d in make_connector_node (
    dict=dict@entry=0x7fff4c00cea0, mp=<optimized out>, 
    linktype=0x7fff3a7fac70 "MDB", dir=<optimized out>, 
    multi=multi@entry=false) at ../../link-grammar/dict-ram/dict-ram.c:530
```

linas:
Above stack trace fixed in #1380 Now I get
```
Thread 103 "cogserv:eval" received signal SIGSEGV, Segmentation fau
#0  pool_alloc_vec (mp=0x7fff4c0155d0, vecsize=vecsize@entry=1)
    at ../../link-grammar/memory-pool.c:189
#1  0x00007fffecbaad75 in pool_alloc (mp=<optimized out>)
    at ../../link-grammar/memory-pool.h:116
#2  Exp_create (mp=<optimized out>)
    at ../../link-grammar/dict-ram/dict-ram.c:442
#3  make_connector_node (dict=dict@entry=0x7fff4c00ca60, mp=<optimized out>, 
    linktype=0x7fff22ffbc70 "DPL", dir=<optimized out>, 
    multi=multi@entry=false) at ../../link-grammar/dict-ram/dict-ram.c:527
```

linas:
Above fixed in #1381 ... Next up,the string-set:
```
#0  find_place (ss=0x7fff50001040, h=940135772, str=0x7fff6f7f4c70 "GLKDB")
[...]

PR #1380, opened 2023-01-21
Add mutex to protect the condesc table.

linas:
The multi-threaded AtomSpace dict will hit this, when multiple threads are trying to insert words.


linas:
This should fix #1379

ampli:
Why not to do it only for atomese dicts?

PR #1381, opened 2023-01-21
Provide a thread-safe Exp pool for the dictionary.

linas:
In the Atomese dictionary backend, multiple threads update the same dict.

### DIFF
--- a/link-grammar/dict-ram/dict-ram.c
+++ b/link-grammar/dict-ram/dict-ram.c
@@ -439,7 +439,7 @@ Dict_node * strict_lookup_list(const Dictionary dict, const char *s)
  */
 Exp *Exp_create(Pool_desc *mp)
 {
-	Exp *e = pool_alloc(mp);
+	Exp *e = pool_alloc_concurrent(mp);
 	e->tag_type = Exptag_none;
 	e->operand_next = NULL;
 	e->cost = 0.0;
@@ -453,7 +453,7 @@ Exp *Exp_create(Pool_desc *mp)
  */
 Exp *Exp_create_dup(Pool_desc *mp, Exp *old_e)
 {
-	Exp *new_e = pool_alloc(mp);
+	Exp *new_e = pool_alloc_concurrent(mp);
 
 	*new_e = *old_e;
 
--- a/link-grammar/memory-pool.c
+++ b/link-grammar/memory-pool.c
@@ -84,6 +84,10 @@ Pool_desc *pool_new(const char *func, const char *name,
 	mp->curr_elements = 0;
 	mp->num_elements = num_elements;
 
+#if HAVE_THREADS_H
+	mtx_init(&mp->mutex, mtx_plain);
+#endif
+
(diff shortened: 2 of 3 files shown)


ampli:
Is it needed in general, or only for atomase dynamic dicts?

linas:
Fix for stack trace in #1379

linas:
> Is it needed in general, or only for atomese dynamic dicts?

Good question. I think it is needed in general. It has not been hit before because:
* I guess no one else is doing multi-theeaded parsing?
* The `multi-thread.c` unit test doesn't hit this, because it is parsing the same sentences over and over, and thus the Exp pool never grows, and thus no contention.
* My earlier heavy-parallel parsing always used the `any` language, which also doesn't hit this.

We could/should create a new test-case to test this.  I'll try to do that now, as I wait for datasets to load.

I imagine maybe this has a significant performance hit? I haven't measured...

ampli:
For file dicts that are created in different threads, I just cannot understand how there may be a contention problem in memory pools or string_sets, since all of there memory locations are on the stack or result from `malloc()`  (no static data).

linas:
> For file dicts

Oh right. I was imagining that `Exp_create()` is called during parsing, but of course, it isn't. 

The lock could be moved higher up, if `make_and_node()` and related took `dict` as an argument, instead of ` dict->Exp_pool` ... except that `sat-solver/sat-encoder.cpp` calls `Exp_create(_sent->Exp_pool)`

ampli:
> if `make_and_node()` and related took `dict` as an argument, instead of ` dict->Exp_pool`

In file dicts, hat is the problem of using `dict->Exp_pool` as argument?
There can be 2 cases:
1. The dict is created and then the threads are launched.  In them, `make_and_node()` is not invoked any more.
2. The threads are launched and create the dicts. Then the dict memory is private anyway.

In any case, what is the different between providing `dict->Exp_pool` as argument or alternatively providing `dict` and refer `dict->Exp_pool` inside `make_and_node()` ?

ampli:
> thus the Exp pool never grows, and thus no contention.

As far as I can see, no dict memory is getting allocated or written after `dictionary_create_from_file()` returns.

linas:
> As far as I can see
Look at `sat_encoder`

linas:
OK... I continue to have issues, and I am realizing that maybe low-level locks are just the wrong place to put this stuff.  I might actually revert the last 3 or 4 commits ...

ampli:
> > As far as I can see
> 
> Look at `sat_encoder`

https://github.com/opencog/link-grammar/blob/9060de9f0a7d9a571d321659b8148ab6e76239bc/link-grammar/sat-solver/sat-encoder.cpp#L496-L498
https://github.com/opencog/link-grammar/blob/9060de9f0a7d9a571d321659b8148ab6e76239bc/link-grammar/sat-solver/sat-encoder.cpp#L1780-L1782
https://github.com/opencog/link-grammar/blob/9060de9f0a7d9a571d321659b8148ab6e76239bc/link-grammar/sat-solver/util.cpp#L60-L64

This is a per-sentence memory.  No dict memory is written.

PR #1382, opened 2023-01-21
make_expression(): Fix dialect setting on Connectors

ampli:
The reason of the problem and how it is fixed are described in the commit message.
See also the discussion starting at https://github.com/opencog/link-grammar/discussions/1310#discussioncomment-4743306.


linas:
Thanks!

PR #1383, opened 2023-01-21
Provide a thread-safe string-set for the dictionary.

linas:


linas:
Fixes stack trace in #1379

PR #1384, opened 2023-01-22
Protect condesc table during sorting.

linas:
Put lock around a large critical section


linas:
Another fix for #1379

PR #1396, opened 2023-01-24
Fix multiple deduplication bugs

linas:
This fixes multiple issues reported for #1378, including:
* Bad identification of region to copy
* Failure to compare dictionary words
* Inefficient copying

### DIFF
--- a/link-grammar/linkage/linkage.h
+++ b/link-grammar/linkage/linkage.h
@@ -67,6 +67,7 @@ struct Linkage_s
 	Linkage_info    lifo;         /* Parse_set index and cost information */
 	bool            is_sent_long; /* num_words >= twopass_length */
 
+	bool            dupe;         /* Duplicate marker */
 	PP_domains *    pp_domains;   /* PP domain info, one for each link */
 
 	Sentence        sent;         /* Used for common linkage data */
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -863,7 +863,7 @@ static Count_bin pseudocount(count_context_t * ctxt,
                            unsigned int null_count)
 {
 	/* This check is not necessary for correctness, but it saves CPU time.
-	 * If a cross link would result, immediatly return 0. Note that there is
+	 * If a cross link would result, immediately return 0. Note that there is
 	 * no need to check here if the nearest_word fields are in the range
 	 * [lw, rw] due to the way start_word/end_word are computed, and due to
 	 * nearest_word checks in form_match_list(). */
--- a/link-grammar/parse/fast-match.c
+++ b/link-grammar/parse/fast-match.c
@@ -632,7 +632,7 @@ form_match_list(fast_matcher_t *ctxt, int w,
 		{
 			(*cmx)->match_left = false;
 		}
-		mr_end = NULL; /* Prevent a gcc "mnay be uninitialkized" warning */
+		mr_end = NULL; /* Prevent a gcc "may be uninitialized" warning */
 	}
 
 	/* Construct the list of things that could match the left. */
--- a/link-grammar/parse/parse.c
+++ b/link-grammar/parse/parse.c
@@ -337,6 +337,16 @@ static void deduplicate_linkages(Sentence sent, int linkage_limit)
 		uint32_t wi;
 		for (wi=0; wi<lpv->num_words; wi++)
 		{
+			// Parses with non-zero null count will have null words,
+			// i.e. word without chosen_disjuncts. Avoid a null-pointer
+			// deref in this case.
+			if (NULL == lpv->chosen_disjuncts[wi])
+			{
+				// If one is null, both should be null. (I think this
+				// will always be true, but I'm not sure.)
+				if (NULL == lnx->chosen_disjuncts[wi]) continue;
+				break;
+			}
 			if (lpv->chosen_disjuncts[wi]->word_string !=
 			    lnx->chosen_disjuncts[wi]->word_string) break;
 		}
--- a/link-grammar/parse/parse.c
+++ b/link-grammar/parse/parse.c
@@ -348,23 +348,43 @@ static void deduplicate_linkages(Sentence sent, int linkage_limit)
 		tgt--;
 	}
 
+	// None found!
+	if (0 == num_dupes) return;
+
 	// Phase two: Sweep
 	tgt = 0;
-	for (uint32_t isrc=1; isrc<nl; isrc++)
+	uint32_t blkstart = 0;
+	uint32_t blklen = 1; // Initial block, already skipped
+	for (uint32_t i=1; i<nl; i++)
 	{
-		Linkage lnx = &sent->lnkages[isrc];
-		if (lnx->dupe)
+		Linkage lnx = &sent->lnkages[i];
+		if (false == lnx->dupe) { blklen++; continue; }
+		free_linkage(lnx);
+
+		// If there's a block of good linkages to copy, then copy.
+		if (0 < blklen)
 		{
-			free_linkage(lnx);
-			continue;
+			// Skip initial block; it is already in place.
+			if (0 < tgt)
+			{
+				Linkage ltgt = &sent->lnkages[tgt];
+				Linkage lsrc = &sent->lnkages[blkstart];
+				memmove(ltgt, lsrc, blklen * sizeof(struct Linkage_s));
+			}
+			tgt += blklen;
+			blklen = 0;
 		}
-		tgt++;
-		if (tgt == isrc) continue; // Nothing to do, yet.
 
-		Linkage lpv = &sent->lnkages[tgt];
+		// The next good linkage comes after this bad one.
+		blkstart = i+1;
+	}
 
-		// Move one linkage.
-		memmove(lpv, lnx, sizeof(struct Linkage_s));
+	// Copy the final block.
+	if (0 < blklen && 0 < tgt)
+	{
+		Linkage ltgt = &sent->lnkages[tgt];
+		Linkage lsrc = &sent->lnkages[blkstart];
+		memmove(ltgt, lsrc, blklen * sizeof(struct Linkage_s));
 	}
 
 	// Adjust the totals.
--- a/link-grammar/parse/parse.c
+++ b/link-grammar/parse/parse.c
@@ -304,13 +304,15 @@ static void deduplicate_linkages(Sentence sent)
 
 	// Phase one: Mark
 	uint32_t num_dupes = 0;
+	uint32_t tgt = 0;
 	for (uint32_t i=1; i<nl; i++)
 	{
-		Linkage lpv = &sent->lnkages[i-1];
+		Linkage lpv = &sent->lnkages[tgt];
 		Linkage lnx = &sent->lnkages[i];
 
 		// Mark
 		lnx->dupe = false;
+		tgt ++;
 
 		// Rule out obvious mismatches.
 		if (lpv->num_links != lnx->num_links) continue;
@@ -331,24 +333,23 @@ static void deduplicate_linkages(Sentence sent)
 		// If we are here, then lpv and lnx are the same linkage.
 		lnx->dupe = true;
 		num_dupes ++;
-
-		i--; // Do not advance i; there may be more!
+		tgt--;
 	}
 
 	// Phase two: Sweep
-	uint32_t off = 0;
-	for (uint32_t i=1; i<nl; i++)
+	tgt = 0;
+	for (uint32_t isrc=1; isrc<nl; isrc++)
 	{
-		Linkage lnx = &sent->lnkages[i];
+		Linkage lnx = &sent->lnkages[isrc];
 		if (lnx->dupe)
 		{
 			// Free stuff
 			continue;
 		}
-		off++;
-		if (off == i) continue; // Nothing to do, yet.
+		tgt++;
+		if (tgt == isrc) continue; // Nothing to do, yet.
 
-		Linkage lpv = &sent->lnkages[off];
+		Linkage lpv = &sent->lnkages[tgt];
 
 		// Move one linkage.
 		memmove(lpv, lnx, sizeof(struct Linkage_s));
--- a/link-grammar/parse/parse.c
+++ b/link-grammar/parse/parse.c
@@ -171,7 +171,7 @@ static void process_linkages(Sentence sent, extractor_t* pex,
 
 	/* Pick random linkages if we get more than what was asked for. */
 	bool pick_randomly = sent->overflowed ||
-	    (sent->num_linkages_found > (int) sent->num_linkages_alloced);
+	    (sent->num_linkages_found > (int) opts->linkage_limit);
 
 	sent->num_valid_linkages = 0;
 	size_t N_invalid_morphism = 0;
@@ -292,11 +292,10 @@ static void process_linkages(Sentence sent, extractor_t* pex,
  * This can be done in a single pass, after the linkages have been
  * sorted. We only need to check nearest neighbors.
  */
-static void deduplicate_linkages(Sentence sent)
+static void deduplicate_linkages(Sentence sent, int linkage_limit)
 {
 	/* No need for deduplication, if random selection wasn't done. */
-	if (!sent->overflowed &&
-	    (sent->num_linkages_found <= (int) sent->num_linkages_alloced))
+	if (!sent->overflowed && (sent->num_linkages_found <= linkage_limit))
 		return;
 
 	uint32_t nl = sent->num_linkages_alloced;
@@ -372,7 +371,7 @@ static void sort_linkages(Sentence sent, Parse_Options opts)
 	      sizeof(struct Linkage_s),
 	      (int (*)(const void *, const void *))opts->cost_model.compare_fn);
 
-	deduplicate_linkages(sent);
+	deduplicate_linkages(sent, opts->linkage_limit);
 	print_time(opts, "Sorted all linkages");
 }
 
--- a/link-grammar/parse/parse.c
+++ b/link-grammar/parse/parse.c
@@ -329,6 +329,19 @@ static void deduplicate_linkages(Sentence sent, int linkage_limit)
 		}
 		if (li != lpv->num_links) continue;
 
+		// Compare words. The chosen_disjuncts->word_string is the
+		// dictionary word. It can happen that two different dictionary
+		// words can have the same disjunct, and thus result in the same
+		// linkage. For backwards compat, we will report these as being
+		// different, as printing will reveal the differences in words.
+		uint32_t wi;
+		for (wi=0; wi<lpv->num_words; wi++)
+		{
+			if (lpv->chosen_disjuncts[wi]->word_string !=
+			    lnx->chosen_disjuncts[wi]->word_string) break;
+		}
+		if (wi != lpv->num_words) continue;
+
 		// If we are here, then lpv and lnx are the same linkage.
 		lnx->dupe = true;
 		num_dupes ++;
--- a/link-grammar/parse/parse.c
+++ b/link-grammar/parse/parse.c
@@ -302,10 +302,15 @@ static void deduplicate_linkages(Sentence sent)
 	uint32_t nl = sent->num_linkages_alloced;
 	if (2 > nl) return;
 
+	// Phase one: Mark
+	uint32_t num_dupes = 0;
 	for (uint32_t i=1; i<nl; i++)
 	{
 		Linkage lpv = &sent->lnkages[i-1];
 		Linkage lnx = &sent->lnkages[i];
+
+		// Mark
+		lnx->dupe = false;
 
 		// Rule out obvious mismatches.
 		if (lpv->num_links != lnx->num_links) continue;
@@ -324,14 +329,35 @@ static void deduplicate_linkages(Sentence sent)
 		if (li != lpv->num_links) continue;
 
 		// If we are here, then lpv and lnx are the same linkage.
-		// Shuffle down all linkages.
-		memmove(lpv, lnx, (nl-i)* sizeof(struct Linkage_s));
-		nl--;
-		sent->num_linkages_alloced --;
-		sent->num_valid_linkages --;
-		sent->num_linkages_post_processed --;
+		lnx->dupe = true;
+		num_dupes ++;
+
 		i--; // Do not advance i; there may be more!
 	}
+
+	// Phase two: Sweep
+	uint32_t off = 0;
+	for (uint32_t i=1; i<nl; i++)
+	{
+		Linkage lnx = &sent->lnkages[i];
+		if (lnx->dupe)
+		{
+			// Free stuff
+			continue;
+		}
+		off++;
+		if (off == i) continue; // Nothing to do, yet.
+
+		Linkage lpv = &sent->lnkages[off];
+
+		// Move one linkage.
+		memmove(lpv, lnx, sizeof(struct Linkage_s));
+	}
+
+	// Adjust the totals.
+	sent->num_linkages_alloced -= num_dupes;
+	sent->num_valid_linkages -= num_dupes;
+	sent->num_linkages_post_processed -= num_dupes;
 }
 
 static void sort_linkages(Sentence sent, Parse_Options opts)
--- a/link-grammar/parse/parse.c
+++ b/link-grammar/parse/parse.c
@@ -343,7 +343,7 @@ static void deduplicate_linkages(Sentence sent)
 		Linkage lnx = &sent->lnkages[isrc];
 		if (lnx->dupe)
 		{
-			// Free stuff
+			free_linkage(lnx);
 			continue;
 		}
 		tgt++;


ampli:
There is still a similar silence problem:
`The response was that there are several barriers that have to come down.`

In addition to fixing the problem, I propose to add an assert() that at least one linkage remain.

Questions:
https://github.com/opencog/link-grammar/blob/21df959757b7b9c64eb49cea6bf8ff8fde8efbcd/link-grammar/parse/parse.c#L406-L411
https://github.com/opencog/link-grammar/blob/21df959757b7b9c64eb49cea6bf8ff8fde8efbcd/link-grammar/dict-file/dictionary.c#L268-L271
For `any` there is a special check for not sorting. What is the reason?
Also, in that case there is no deduplication. Is it intended?
Now that we have dict `#define`, maybe it's better to define `shuffle_linkages` in the `any` dict?

linas:
Yeah ... There's still something weird going on. I reordered the code to mark during sorting, and am now finding many many more duplicates. And when this happens, somehow something doesn't work.

Yes, the `any` should be a dict `#define`

linas:
Ahhh there are more duplicates, than valid linkages. Then it fails... because I dedupe the bad linkages too.  Perhaps the bad linkages shouldn't be sorted or deduplicated, anyway!?

linas:
The `any` linkages all have the same cost, so in that sense, they are unsortable. However, the newest code (not yet pushed) could provide a unique sort, anyway, and thus allow deduplication. However ... we do not want that.  Because ...

The `any` language is supposed to provide a uniform statistical sample of all possible planar trees. Duplicate removal might ...!??

Certainly, if num linkages < limit, then *all* planar linkages will be enumerated, one each, there won't be duplicates, and so these are by definition "uniformly distributed".  If num linkages > limit, then ... we sample ... does that sampling result in a uniform distribution?   I don't really know. I assumed it did ... Would removing duplicates make it "more uniform" if it is not?  How many duplicates might there be, anyway?

The reason for the shuffle is that when I sample,  I set limit=1000 but then take only the first 24 linkages, ever. I'd tripped over a bug, where, no matter what the sentence, I always got back exactly the same 24 parses. So that wasn't uniform at all, over thousands of sentences. The shuffle guarantees that a different set of 24 appears each time.

linas:
I won't be fixing the `any #define` thing ... if you want to do that, go ahead ...

linas:
#1397 fixes the too-many-duplicates bug mentioned above.

ampli:
I still get problems with my checks.
Here is one of them.
The other one is with a sentence having 214568 linkages which apparently has a problem due to checking connector addresses in `linkage_equiv_p()`. I will investigate this further, but also see below.

My test sentence is:
`He said that, finding that it was impossible to get work as a waiter, he would work as a janitor`

In the current release (5.12.0) I get:
`Found 2280 linkages (241 of 1000 random linkages had no P.P. violations)`
If I filter the results through my duplicate elimination Perl program, it founds 61 duplicates and 180 w/o duplication.

In the repository release (9c071f78f, just before you recent dict changes):
Found 2280 linkages (185 of 944 random linkages had no P.P. violations)
My duplicate elimination programs then still find 5 duplicates!
For example, linkages 27 and 28 are duplicates of linkage 26.
I still don't understand why `linkage_equiv_p()` misses them.

So something in the deduplication algo apparently doesn't detect all the duplicates.

Some other issues:

> Use Connector::desc not tracon_id per #1396

(I cannot find `tracon_id` mentioned in # 1396, so I guess it was in another issue/PR.)

There are some problems in using desc:string:

1. It only makes the problem of potentially false duplicates (if possible) more severe, as indeed tracon addresses (or tracon ID's) may be the same in different disjuncts of the same word (that happen to end with the same tracon), but connector names may be the same in totally unrelated disjuncts, even not on the same word! (It is not checked that they are on the same word).
2. desc::string is common to regular `multi` connectors, i.e. it doesn't contain the `@`. This surely may lead to false duplicates. So Connector::multi should be checked too.

In addition, connector address comparisons are still done, as in:
`if (lpv->link_array[li].lc != lnx->link_array[li].lc)`
As I mentioned elsewhere, connector addresses may be equal on different disjuncts due to tracon compression.
This causes deduplication problems (**and I still need to give you exact examples**).

Last thing here:
For short sentences, there is +~10% added overhead. Since the linkage info index is negative when sampling is to be done, maybe `linkage_equiv_p()` may be skipped on negative indexes.
However, thinking about that brought me another idea for deduplication, which I will write about shortly.

ampli:
> However, thinking about that brought me another idea for deduplication, which I will write about shortly.

This idea is not as simple as I first thought, and anyway I recalled it had already been used in the old code:

In case there is no linkage overflow, say we found N linkages when L=linkage_limit.
The idea is to create L unique random numbers [0..N) and use them as the linkage index.
However, it may not be particularly simple to create such a list.

ampli:
In #1378 you said
> After fixing assorted bad tests & etc the deduplication code for Baird is no longer entered. But if let dedupe run anyway, I get this:
>
>No flags (default config) one duplicate found, the second linkage!
-test=min-len-encoding:0 same
-test=min-len-encoding:254 no duplicates seen
The difference between the first and second linkages is that the first uses vice.s and the second uses vice.n-u although the link types used are the same. To be backwards-compatible, I will need to also check for mismatched words.

There is still linkage differences in some sentences with `min-len-encoding:0` and `min-len-encoding:254', e.g.:
`In this case, surface tension of the liquid solder will pull on one side more than the other, and the part will tombstone.`

Since the linkages without deduplication are the same, I think that it is a bug.
To be on the safe side, there is a need to check this using the version just before the dict changes.
I also think that since the linkage comparison is subtle (and it is not clear yet how to fix it), it will be a good idea to add `-text=no-deduplication` so we can directly validate the results.

ampli:
> `In this case, surface tension of the liquid solder will pull on one side more than the other, and the part will tombstone.`

Note that it may not be easy to check that `min-len-encoding:0` and `min-len-encoding:254` indeed produce the same linkages when random sampling is not used.
To produce linkage diagrams file for this sentence, you can do:

`echo 'SENTENCE' | link-parser -test=min-len-encoding:0,auto-next-linkage -lim=300000 >  /tmp/l0`
`echo 'SENTENCE' | link-parser -test=min-len-encoding:254,auto-next-linkage -lim=300000 >  /tmp/l254`

But `diff /tmp/l0 /tmp/l254` will then show many differences due to slight linkage order changes (I yet need to investigate why there is an order change but for now, I found it is not due to address randomization).
But if I alphabetically sort the linkages there are no differences (and no duplicates - no random sampling has been used).

Another observation is that with `-limit=1000` (the default linkage_limit) there are still 6 undetected duplicates.

linas:
I apologize for changing the dictionary. I should not have done that.

ampli:
Let's start with one point I partially agree with you about:
> Instead of comparing links, I guess I could have compared to see if chosen_disjuncts are the same, or not.

Such a check is needed **too**. However, sentences with the same disjuncts may have different linkages.
Here is a minimal example:
``` test
test/4.0.dict:

w0: A+;
w1: A+;
w2: @A- & A+;
w3: @A-;

$ link-parser test
linkparser> w0 w1 w2 w3
...
Found 2 linkages (2 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS= 0.00 LEN=1)

 +--A--+
 |  +-A+-A+
 |  |  |  |
w0 w1 w2 w3

linkparser> 
	Linkage 2, cost vector = (UNUSED=0 DIS= 0.00 LEN=2)

 +----A---+
 |  +-A+-A+
 |  |  |  |
w0 w1 w2 w3
```

linas:
I fixed the bad linnkage compare in e35ea9f4db957009b9dba4404552499ef7aabd51 and pushed directly upstream.

linas:
> Such a check is needed too.

Hmm. I won't be doing that check. If two different disjuncts give the same links and use the same dict words, then, from the point of view of the linkage, they are the same, even if the dict managed to include two different disjuncts, for the same word (i.e. one with and one without a multi-connector, but otherwise being the same).  

Hmmm. At least, I think I want to ignore such differences. But for dict debugging, they do make a difference... 

From what I can tell, it is enough to compare pointers for the `chosen_disjuncts` ....

linas:
I added disjunct compare, but stubbed it out, in 9f6c9c4ae2803fd1b98280a7e005ff89c0667fdf

linas:
And one more: a performance optimization: 43be2538a6df49de795c1bb966e852a2bc7d7d32

After this, the connector compare is almost completely unneccesary. It does check for multi-connectors, but there will almost never be a difference.

Re: performance: I assume most of the performance loss is in the `memmove()` and not in the compare. I will measure perf now ...

ampli:
> Are you sure that `-test=min-len-encoding` is not changing how the random number generator is being called?

Yes. I'm sure. Changing ` min-len-encoding` doesn't change the disjuncts and their order at all, and doesn't add/delete disjuncts. If such a change would happen it would be considered a severe bug that could cause wrong linkages, and I am not aware of such a bug.
I have a program that validates that  `min-len-encoding` (among other several things) doesn't modify the linkages, and I run it every time I (or you) make a change.

Hence the sampling (random or not) always gives the exact linkage list, independently of the value of  `min-len-encoding`!
This is true also in the current repository version (e35ea9f4d).
However, `linkage_equiv_p()` depends on ` min-len-encoding` (whether connector address sharing is done or not).

Here is how:
https://github.com/opencog/link-grammar/blob/43be2538a6df49de795c1bb966e852a2bc7d7d32/link-grammar/parse/parse.c#L354-L357
The connector equality check above may not do the intended thing.
The intention is to check if both linkages have the same `lc` connector at the same `link_array` element.
It is done by comparing the connector address. If the connectors are not equal, they are surely not the same connector.

But what it means if they are equal? Consider this simple case (there is a complex one):
Disjunct on linkage `p`:        `A+ B+ <> C- D-`
Disjunct on linkage `n`:        `F+ G+ <> C- D-`
These are of course different disjuncts.

Suppose we reach this point for such disjuncts (we don't - after your last changes).
In that case, if `min-len-encoding` is 0 (the default case), then `(plk->lc != nlk->lc)` returns **false** (since the connectors then share addresses)! But for  `min-len-encoding` of 254, it returns **true** (no connector address sharing so connectors on different disjuncts never have the same address).

(The complex case involves another disjunct `x` `A+ B+ <> E- C- D-`. By default (`min-len-encoding` is 0) `C-` on `x` has the same address as `C-` on `p` since tracon shares addresses).

After your last changes (43be2538a), this problem doesn't happen when I run my linkage check script (it checks the `basic` batches of all the languages and also the English `fixes` batch. But I cannot see anything that guarantees it could never happen unless you check the disjuncts equality.

> `#if DOUBLE_CHECK`

I propose changing it to `#ifdef DEBUG`.

ampli:
In my previous post I added a missing line (marked by EDIT:)
Consider connector C-

ampli:
> I don't care if the C structures are shared or not; this seems irrelevant.

But as indicated above, you actually check the equality of pointers to these structs, which may be equal if they are shared even if they are found on different disjuncts (and even if the connector lists on the disjuncts are different).

linas:
> equality of pointers 

I still do not understand what you are saying. The pointer equailty is only used to speed up comparisons. It does not actually change the comparisons.  So, this:
```
      if (plk->lc != nlk->lc)
      {
         int diff = strcmp(plk->lc->desc->string, nlk->lc->desc->string);
         if (diff) return diff;
         int md = plk->lc->multi - nlk->lc->multi;
         if (md) return md;
      }
```
I could skip doing the pointer compares, and the result would not change at all, because if the pointers are equal, the `strcmp` will always return zero, and `md` will always be zero.

linas:
Re performance:  Working with the the same dict, comparing `pandp-union` I get that the current master is 5% faster than what it was at #1375 

However, I now get 1991 errors instead of 1834 errors  (using the same dict...)

ampli:
> I still do not understand what you are saying.

I think that it is important that we both agree on this thing.

Question:
Hypothetically, what happens if at this point the pointers are equal but are found on two different disjuncts?
Does it mean that the linkages are different?

linas:
> Hypothetically, what happens if at this point the pointers are equal but are found on two different disjuncts?

It is irrelevant, since the `strcmp` would return zero in this case.

> 1991 errors 

This was incorrect. I did not copy the original dictionary correctly.  So, for me, `pandp` do compare correctly, and the new code is slightly faster.

ampli:
> Re performance: Working with the the same dict, comparing `pandp-union` I get that the current master is 5% faster than what it was at #1375
> 
> However, I now get 1991 errors instead of 1834 errors (using the same dict...)

I tried these versions both with the old and new dicts. From `git describe`:
Version as of #1375:
link-grammar-5.12.0-79-g750315dfa
Current version:
link-grammar-5.12.0-192-g43be2538a

I got about the same speed in both versions, including on `pandp-union`.
Also the same number of errors (1834) on `pandp-union` with both dicts and both versions.
However, on `basic` I got one less error with the new dict, as this sentence now doesn't parse:
``` text
linkparser> The price of the stock rose three tenths of one dog
No complete linkages found.
Found 36 linkages (16 had no P.P. violations) at null count 2
	Linkage 1, cost vector = (UNUSED=2 DIS= 2.10 LEN=23)

                   +------------Os------------+------------Bsd------------+
                   |          +-----Ds**x-----+------Rn-----+------Sp-----+
    +----->Wg------+          |      +---AN---+     +---NW--+--Mf-+Js+    |
    |              |          |      |        |     |       |     |  |    |
LEFT-WALL [the] price.v [of] the stock.n-u rose.n three tenths.m of one dog.v
```

It could be interesting to find out why we got different speedup results and a different number of errors on  `pandp-union` - maybe I did something wrong.

ampli:
> ```
>       if (plk->lc != nlk->lc)
>       {
>          int diff = strcmp(plk->lc->desc->string, nlk->lc->desc->string);
>          if (diff) return diff;
>          int md = plk->lc->multi - nlk->lc->multi;
>          if (md) return md;
>       }
> ```

> > Hypothetically, what happens if at this point the pointers are equal but are found on two different disjuncts?
> 
> It is irrelevant, since the `strcmp` would return zero in this case.
> 
> >

But if the pointers are equal, i.e. `(plk->lc == nlk->lc)`, it will not enter the `if` body and `strcmp` would not even be invoked.

linas:
> number of errors (1834)

This is correct; my earlier report was incorrect.

> one less error with the new dict,

Yes. that is a side-effect of the change.

> I got about the same speed in both versions,

I have an older CPU that you do, the difference could be cache-locality effects. Maybe for you everything fits in the cache, and for me, older one spills and the newer doesn't.

Hard to imagine... with the duplicates removed, maybe there's less pressure on the data cache? Hard to see how, though, because this is almost the last step, before everything is freed...

Perhaps one of the other commits changed performance, e.g. #1393 or #1391 or #1385 -- The #1385 uses less stack space.

linas:
Anyway, for me, the current version is 5% slower than the all time best, which was just before version #1102 .. (yes, my current version does have `-O3` but not `-flto` or `-Ofast`)

linas:
Well, that's weird. Recompiling with `CFLAGS="-O3 -flto -Ofast"` makes it go 3% slower than without. So that is definitely a icache spell effect or a itlb collision.

ampli:
For me, LG with `-flto` never gave any speedup.

linas:
Hmm. I mis-spoke, again. I see a speedup with broken dictionaries.  No speedup when I fix the dicts.  Also, with the correct dicts, the `flto -Ofast` goes 2% faster... 

"Move fast and break things" .... and cleanup the mess afterwards. Oh well.

linas:
> But if the pointers are equal, i.e. `(plk->lc == nlk->lc)`, it will not enter the if body and `strcmp` would not even be invoked.

Yes, exactly. We don't need to invoke `strcmp` in this case, because we know what it would do. So the pointer compare is just an optimization that allows use to skip the `strcmp`

ampli:
> > But if the pointers are equal, i.e. `(plk->lc == nlk->lc)`, it will not enter the if body and `strcmp` would not even be invoked.
> 
> Yes, exactly. We don't need to invoke `strcmp` in this case, because we know what it would do. So the pointer compare is just an optimization that allows use to skip the `strcmp`

Ok, so let's sum up what we agreed (if I'm mistaken - please correct me where you don't actually agree).
1. The connectors are equal.
2.  The disjuncts are **different** (as this is a case of equal connectors on totally different disjuncts due to memory sharing).
3. The `if` conditions are not entered and the program continues below.

Say this happens at the last iteration(s) of the `for` loop, and it exists.
And then we get to `lnx->dupe = true;`. But this is incorrect, as the disjuncts are different.

linas:
> The disjuncts are different

How can they possibly be different? We've just finished loops that checked everything there is to check, and found that everything is the same.

ampli:
> > The disjuncts are different
> 
> How can they possibly be different? We've just finished loops that checked everything there is to check, and found that everything is the same.

What is supposed to ensure that they are the same? Let's see.

Loop 1: // Compare link endpoints
This of course doesn't ensure the disjuncts are the same.

Loop 2: Compare link names.
This also doesn't ensure the disjuncts are the same.
It also doesn't ensure the endpoint connectors have the same names (because the same link can theoretically be constructed from different connector pairs), but it is a  **heuristic** that they are the same (because the dict is not specially designed to break that). I can add a debug check to see if it always holds.

Loop 3. // Compare words.
This surely doesn't validate that the disjuncts are the same (the different disjuncts on the same word (2D slot) often have the same string).

Loop 4. // Compare connector
This is the loop we discuss. As far as I can see no loop before it ensures the disjuncts here are the same.

And indeed, when I added
``` c
	for (uint32_t wi=0; wi<lpv->num_words; wi++)
		assert(lpv->chosen_disjuncts[wi] == lnx->chosen_disjuncts[wi], "XXX");
```
just above `if DOUBLE_CHECK`, it failed, for example, for this sentence:
`Business is booming, Joe Smith, a car dealer, says`

This proves that the disjuncts can be different before loop 4 when loop 4 normally finishes with no `return`,
and `lnx->dupe = true;` can be reached when the disjuncts are not the same.
Have I missed something?

linas:
> `Business is booming, Joe Smith, a car dealer, says`

right before the `return 0;` I placed this:
```
for (uint32_t wi=0; wi<lpv->num_words; wi++) {
if (lpv->chosen_disjuncts[wi] != lnx->chosen_disjuncts[wi]) {
  printf("duuude difference at %u %s %s vs %s\n", wi,
lnx->chosen_disjuncts[wi]->word_string,
linkage_get_disjunct_str(lpv, wi),
linkage_get_disjunct_str(lnx, wi));
}}
```

prints
```
duuude difference at 3 boomingv @MXs+ @MXs+ Ss+ vs @MXs+ Ss+
```
Casual examination shows that neither of these disjunct strings is correct.  WTF
```
    |                                 |          |    +------Xd-----+
    +-------->WV-------->+            |          |    | +---Ds**x---+
    +---->Wd----+---Ss---+--Osm-+     |   +---G--+    | |   +---AN--+-Xc-+
    |           |        |      |     |   |      |    | |   |       |    |
LEFT-WALL business.n-u is.v booming.v 
```

linas:
Oh never mind. The printed parse is not the parse being shown

linas:
It should have been this one:
```
    |                           +----------------------Ss------------------
    |                           +----------------MXs----------------+
    |                           +-------MXs------+    +------Xd-----+
    +-------->WV-------->+      |     +----Xd----+    | +---Ds**x---+
    +---->Wd----+---Ss---+      |     |   +---G--+-Xca+ |   +---AN--+-Xc-+
    |           |        |      |     |   |      |    | |   |       |    |
LEFT-WALL business.n-u is.v booming.v , Joe.b Smith.m , a car.n dealer.n ,
```

So the disjuncts `@MXs+ @MXs+ Ss+` and  `@MXs+ Ss+` are printed differently, but they result in the same parse.  The "de feacto" disjunct would be `2@MXs+ Ss+` if you allow me to invent a new notation to say "the MXs connector was used exactly twice."

For linkage de-deuplication, it is the defacto links that I'm interested in, and not the dictionary entries that lead to them.   The reason for this is because ... (next post)

linas:
The re4ason for this ... just evaported. I was going to write something, and I see it won't support my argument.  One of my dicts has expressions entirely of the form `{A+/-} & {B+/-} & {C+/+} & {D+/-}` where the A,B,C,D can be any one of thousands or tens of thousands of connectors.  An earlier/alternate version used `@A+/- & @B+/- & @C+/+ & @D+/-` but for some reason, I decided "zero or one" is better than "zero or more".

So when I get a linkage, I am only interested in the actual connectors used, and not the hypothetical multi's that were never used.

linas:
FYI, if I turn on deduplication, I find that linkages 9, 12 35, 39, 47, 50, 63 and 67 are duplicates.  Presumably, one linkage corresponds to the `@MXs+ @MXs+ Ss+` disjunct while the other corresponds to the `@MXs+ Ss+` disjunct.

This raises the question: should deduplication be turned on in general?  Should it be a run-time flag?

ampli:
> For linkage de-deuplication, it is the defacto links that I'm interested in

Previously I said (multiple times) that the default is `-test=min-len-encoding:0`. However, in the current code it is actually 6 and not 0, due to a very slight overhead in doing connector encoding for very short sentences. So fo consistent results, it should be changed to 0. This is fine since the overhead is negligible. But I think it will be a good idea to ensure that the eliminated linkage is the one with more connectors. Is this what the current code does?

**However**, I think it will be better to just collapse  "@X @X" sequences to a single `@x` in `eliminate_duplocate_disjuncts()`.

ampli:
> This raises the question: should deduplication be turned on in general? Should it be a run-time flag?

At least, there should be a `-test` flag to disable the linkage deduplication.
If we find a good use of that, then we can convert it into a real option.

ampli:
> collapse "@x @x" sequences to a single `@x` in `eliminate_duplocate_disjunct()`

I just found an old code in which I added duplicate `multi` elimination in `build_disjuncts()`, in a hope of a speedup (and got a slight slowdown instead).
However, I did it then in an inefficient way and if eliminating such disjuncts is a good idea I can improve it.

linas:
Proceed as you wish; I'm done, here.

I'm now off to fry my explosively large dictionaries...

FYI, I'd like to publish version 5.12.1 in a few days or weeks or whenever its ready, before too long.

ampli:
> ```
> duuude difference at 3 boomingv @MXs+ @MXs+ Ss+ vs @MXs+ Ss+
> ```

The existence of `@MXs+ @MXs+` means the disjunct should connect exactly two words to the right. See below for my question.

I said:
>> collapse "@x @x" sequences to a single @x in eliminate_duplocate_disjunct()

I wrote the code to do that, and debug printing also found things like `hCO @hCO Wd` turning into ` @hCO Wd`.
However, in general, it is correct to just collapse sequences of N same string connectors (when at least one of them is multi) into a single multi-connector, unless there are similar disjuncts with a sequence of M such similar connectors when M < N. E.g. it will be wrong to collapse `hCO @hCO Wd` into ` @hCO Wd` unless there is already an identical such disjunct.

So I propose another solution to my problem of automatically comparing linkage results after making a change:

> This raises the question: should deduplication be turned on in general? Should it be a run-time flag?

Providing such a flag (e.g. -test=no-linkage-dedup`) will solve the above problem. This can also serve us if we suspect that a linkage is wrongly suppressed, e.g. after changes in the sorting function. So I will implement such a flag.

linas:
Some confusion above...

> `@MXs+ & @MXs+`

This just says "connect to zero or more `MXs` to the right" and is functionally equivalent to a single `@MXs+`

> `hCO & @hCO`

This means "connect to one or more `hCO`" and is NOT the same as `@hCO`.  However, it seems likely that the actual dictionary entry is `{hCO} & @hCO` which is zero or more and thus equivalent to a single `@hCO`.

FWIW, + and - directions commute, so if there are expressions like  `@MXs+ & B- & @MXs+` the `B-` can be slid past the MX's, and then the two MX's can be combined. I don't know if there's any code that automatically slides all the -'s to be left of all the +'s... This sliding or "normal ordering" is possible only for disjuncts, and not, in general, for expressions.

> See below for my question.

I don;'t see a question.

ampli:
> is functionally equivalent to a single `@MXs+`

Does `@MXs` actually mean "connect to one or more `@MXs`? (`{@MXs}` is zero or more.)

> I don't see a question.

It is an implied one, whether the "it will be wrong to collapse ..." sentence is correct.

linas:
Heh.

> @MXs actually mean "connect to one or more `MXs`

Yes it does. I was wrong; I misremembered. I was so used to seeing `{MXs}` that I forgot the true meaning of "one or more"!

The following rewrites are wrong:
* `hCO @hCO`  to `@hCO`  because the former is "two or more", the later is one or more.
* `@MX @MX` to `@MX` is wrong because the former is "two or more"

Thus, it would appear that my earlier insistence that "they are the same" (in https://github.com/opencog/link-grammar/pull/1396#issuecomment-1404555697 and other comments) is wrong.

And now that I see the error of my ways ... what to do? The disjuncts differ, but they give the same parse ... Hmmm .... perhaps you are correct, these should be considered to be different.

.. Should I patch this? Do you want to patch this?

linas:
I opened #1417 because otherwise, this conversation will get lost in history.

ampli:
I suppose that the last two loops (including the DOUBLE_CHECK loop) are to be removed.

The question is how to sort the linkages (that happen to look the same) in that case.
Maybe we can compare the disjuncts by their connectors (string and multi), taking into account that the number of connectors may be different (e.g. return -1 if an earlier NULL is encountered).

linas:
> I suppose that the last two loops (including the DOUBLE_CHECK loop) are to be removed.

??? I must be getting tired. The second-from-last loop lines 352 to 372, bails only if there are differences, so we want to have it.

https://github.com/opencog/link-grammar/blob/e74623007ae682c379f8a8a7aa4baf4608a2fafb/link-grammar/parse/parse.c#L345-L371

We want to keep the above.

The last loop, commented out with `#if DOUBLE_CHECK` lines 373-383, we want to enable that loop. It will enable the detection of different disjuncts, and will also provide a sort order.  Perhaps the `linkage_get_disjunct_str()` at line 380 is CPU-wasteful, and can be replaced by a more direct comparison?

ampli:
> We want to keep the above.

I now think that you are right.
But I still do not like the `if (plk->lc != nlk->lc)` and `if (plk->rc != nlk->rc)` comparisons, since they depend on whether connector encoding is used and whether it decides to share the addresses of particular connectors.
Why not use instead `if (plk->rc->desc->string == nlk->rc->desc->string)` etc.?

> Perhaps the linkage_get_disjunct_str() at line 380 is CPU-wasteful, and can be replaced by a more direct comparison?

Yes. My proposed connector comparison (string + multi) seems more direct.
I mean something similar to `disjuncts_equal()` (but returning a sort-order integer).

ampli:
I said:
> Why not use instead if (plk->rc->desc->string == nlk->rc->desc->string) etc.?

`if (plk->rc->desc == nlk->rc->desc)` is even better.

linas:
> But I still do not like

Read the code more carefully. It uses `!=` not `==` ... the optimization of checking the descs is in #1421

ampli:
> Read the code more carefully. It uses != not == ... the optimization of checking the descs is in #1421

I added a comment on that in your last PR.

linas:
> I added a comment on that in your last PR.

Click on "finish review" else I don't see comments.

ampli:
I added it directly on the commit (in the repository commit list). I think such comments are immediate.,

ampli:
> Click on "finish review" else I don't see comments.

I again forgot to do that...
Please check now.

PR #1407, opened 2023-02-04
Tail pointer is not needed

linas:
It is not needed, it makes no difference, it uses space, and it makes the code more complicated.


linas:
@ampli I am not going to merge this, until you say I should. I don't want to clobber your other pending pull reqs.  Let me know.

ampli:
> @ampli I am not going to merge this, until you say I should. I don't want to clobber your other pending pull reqs. Let me know.

I sent my PR (#1409). You can apply it now because it is complete by itself and my next changes there will take several more days.

PR #1408, opened 2023-02-04
Explicitly track the number of malloced bytes.

linas:


linas:
Merged, because I don't believe this impacts any of your pending pull reqs.  This should solve the comment in issue #1401

PR #1411, opened 2023-02-05
Provide a better x-table-size estimate

linas:
Per data collected in #1402


linas:
See https://github.com/opencog/link-grammar/discussions/1402#discussioncomment-4872249 which has this graph:

djl-dj-buck-sqrt

linas:
The earlier changes provide a 2.5% improvement in parse times for `pandp` and this change alone provides another 2.5%.

PR #1414, opened 2023-02-05
Change the match-list size to more appropriate size

linas:
Based on this graph: https://github.com/opencog/link-grammar/discussions/1402#discussioncomment-4826342

### DIFF
--- a/link-grammar/parse/fast-match.c
+++ b/link-grammar/parse/fast-match.c
@@ -236,14 +236,16 @@ static void sort_by_nearest_word(Match_node *m, sortbin *sbin, int nearest_word)
 /// data. The appropriate blocksize seems to be 1e-4 *numexp**1.5
 /// based on the graph in
 /// https://github.com/opencog/link-grammar/discussions/1402#discussioncomment-4826342
+/// The estimate is meant to be an *upper bound* for how many will be
+/// needed; reallocation to get more is expensive, and is to be avoided.
 static size_t match_list_pool_size_estimate(Sentence sent)
 {
 	size_t expsz = pool_num_elements_issued(sent->Exp_pool);
 	double scaling = expsz * sqrt((double) expsz);
-	scaling *= 1.0e-4;
+	scaling *= 3.0e-3;
 
 	size_t mlpse = scaling;
-	if (mlpse < 1020) mlpse = 1020;
+	if (mlpse < 2040) mlpse = 2040;
 	return mlpse;
 }
 
--- a/link-grammar/parse/fast-match.c
+++ b/link-grammar/parse/fast-match.c
@@ -12,6 +12,8 @@
 /*                                                                        */
 /**************************************************************************/
 
+#include <math.h>                       // sqrt()
+
 #include "api-structures.h"             // Sentence_s
 #include "connectors.h"
 #include "disjunct-utils.h"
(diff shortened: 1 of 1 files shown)


linas:
Specifically, this graph:

djm-exp-mlc

linas:
@ampli -- Comments? I notice that performance depends on the match list size.  I think I set it high enough that existing users are unaffected. However, my performance measuring is sloppier and less through than yours.  So if this looks like its not good enough, let me know.

linas:
Never mind. I'm pretty sure I got it right, finally.

ampli:
Note that vectors are allocated from this pool. Currently if the requested vector size is greater than a block, `pool_alloc_vec()` will fail (with the wold library), and this may happen in unpredictable situations.
This can be easily fixed in the memory-pool code by allocating a block for such big requests (exactly in the size of the request).

linas:
I set a lower bound size at 4090. I assume this would be enough? When would it not be?

I did see some vectors get as large as 1100 ...

ampli:
If `form_match_list()` would return a match list that more than 4090 of its elements would have a linkage (for both LHS and RHS), then a segfault would happen.

As I said before, this problem (of inability to allocate vectors that have more elements than `num_elements`) can relatively easily be solved in `pool_alloc_vec()`, so I can add that there. (A more drastic measure is not to cache such big lists but there is no need for that since the memory-pool fix is easy enough.)

linas:
> can relatively easily be solved

Feel free to fix it, if you wish! 

I do not understand what `form_match_list()` is actually doing, so I don't know why it would ever get larger than 4090.

ampli:
> I do not understand what `form_match_list()` is actually doing
``` c
form_match_list(fast_matcher_t *ctxt, int w,
                Connector *lc, int lw,
                Connector *rc, int rw,
                Disjunct ***mlcl, Disjunct ***mlcr)
```
From its comment header:
```
 * Forms and returns a list of disjuncts coming from word w, that
 * actually matches lc or rc or both. The lw and rw are the words from
 * which lc and rc came respectively.
```

So if there are many such matching disjuncts on word `w`, and more than 4090 of them just happen to yield a count (say LHS count - the same is for RHS count) then a problem will happen (only list elements that yield a count are cached).

linas:
>> I do not understand what `form_match_list()` is actually doing

Of course I should have RTFM. But I'm lazy. So ... Lets assume that word `w` has 50K disjuncts on it. (How many of these have both `lc` and `rc` on them? Lets assume all of them, for now.)  Words `lw` and `rw` can also have 50K disjuncts on them. Lets assume that ... most of those have `lc` and `rc` on them.  Then ... Hmmm... Yes, I see now ... the match list could get very large. 

FYI, the MST code is kind-of the "worst-case scenario". In my learning code, later stages group together words and connectors into classes. But that happens after MST, not before, and so, in these early stages, the total number of disjuncts is explosively large.  I am trying to mitigate this, by doing a form of incremental learning, but this is hard to do.  I'm still stuck in batch mode, and some of the batches get large and its hard to control.   The goal of incremental learning is to use "true disjuncts" when these are available, and supplement them with pair-wise links, when not.

ampli:
> the match list could get very large.

The question is also how many match list elements may have a count. I guess that it is possible that a large portion of them.

linas:
> The question is also how many match list elements may have a count. I guess that it is possible that a large portion of them.

I had originally set the limit to 1020 and then noticed that `pandp-union.txt` resulted in a handful of warnings that `1100 > 1020`  so I changed it to 4090 in response. I lost my log-files last night. I will try to remember to look when I run MST again tomorrow.

linas:
Oh, I have a quick and easy fix ... the match list can never be larger than the total number of disjuncts on a word, right? Then we can just set the pool block size to be the max number of disjuncts on a word.  Pull request coming shortly.

ampli:
> Then we can just set the pool block size to be the max number of disjuncts on a word. Pull request coming shortly.

But this will overestimate it for the `en` generation, which may have 3.5M disjuncts on the wildcard words!

ampli:
I guess that it can be just clamped.

linas:
> overestimate

I'll put in an upper bound?

Line 705 of count.c: 
```
if ((ml == NULL) || (c == NULL)) return; /* Cannot allocate cache array */
```
What happens if there's no cache array? Does it limp on, or die some horrible death?

ampli:
It seems that there will just not be a cache for this particular match list and it will be produced again from scratch if needed.
So after all this is not fatal (I was not sure I have implemented graceful handling but now I see that I have).

linas:
Well, I created #1422 just in case. The next set of choices are:
* Reject #1422 because it is yet more extra complexity.
* Accept #1422 the worst that will happen is a blocksize of 4 MBytes which won't kill anything.

ampli:
This change seems fine to me.

PR #1415, opened 2023-02-05
Provide estimates for the number of parse choices

linas:
Based on measurements presented in #1402. These are upper-bound estimates; they allow a a handful of large allocations to be made, with actual `Parse_choices` issued out of those blocks.

### DIFF
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -158,6 +158,32 @@ int estimate_log2_table_size(Sentence sent)
 	return log2_table_size;
 }
 
+/// Provide an upper-bound estimate for the number of Parse_choice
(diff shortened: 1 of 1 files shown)


linas:
The relevant diagram is this one: 

exp-both-cho-act

ampli:
There is a bug in determining the pool size for `Pset_bucket` and maybe also for `x_table_size`:
``` text
$ link-parser
linkparser> I expect
No complete linkages found.
link-grammar: Warning: Pool Pset_bucket: num_elements is too small 1 >= 0)
```
It then segfaults.


(`pool_new()` could catch this earlier, as it can disallow pool size=0.)

linas:
Ooops!   Fixed in #1418

PR #1428, opened 2023-02-10
Fix a FIXME from wordgraph

linas:
Fix an old FIXME about sharing the `Wordgraph_pathpos` structure.

### DIFF
--- a/link-grammar/linkage/sane.c
+++ b/link-grammar/linkage/sane.c
@@ -18,10 +18,38 @@
 #include "error.h"
 #include "linkage.h"
 #include "sane.h"
-#include "tokenize/tok-structures.h"    // Wordgraph_pathpos_s
+#include "tokenize/tok-structures.h"    // Morpheme_type
 #include "tokenize/word-structures.h"   // Word_struct
 #include "tokenize/wordgraph.h"
 #include "utilities.h"
+
+// Wordgraph_pathpos_s
+struct Wrdgr_path_s
+{
+	Gword *word;		/* Position in the Wordgraph */
+	const Gword **path; /* Linkage candidate wordgraph path */
+};
+typedef struct Wrdgr_path_s Wrdgr_pathpos;
+
+static size_t wrdgr_path_len(Wrdgr_pathpos *wp)
+{
+	size_t len = 0;
+	if (wp)
+		while (wp[len].word != NULL) len++;
+	return len;
+}
+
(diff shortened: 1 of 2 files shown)


linas:
I'm interested in this, as I wish to refactor when/how the `X_node` lookup is done.  I want to do `X_node` lookup *after* flattening the wordgraph.

linas:
I merged this immediately, since it seems to be 100% safe, identical to the original design, just a renamed structure, that's all.

ampli:
> I'm interested in this, as I wish to refactor when/how the `X_node` lookup is done. I want to do `X_node` lookup _after_ flattening the wordgraph.

This fix (I suppose it is correct) is fine by itself, but I missed what it has to do with postponing the  `X_node` lookup.
For that, it seems you just need to remove `determine_word_expressions()` from `flatten_wordgraph()` and invoke a new function after its invocation (named for example `create_x_nodes()`) that scans the 2D array and adds `X_node` elements.

> I want to do `X_node` lookup _after_ flattening the wordgraph.

 Why?

linas:
> you just need to remove `determine_word_expressions()`

Yes, exactly.  I'm doing that now.

> Why?

Sigh. Well, its ugliness in the Atomese word-pair spanning tree code.  Turns out that, after fixing a few more bugs, I have maybe 50K word pairs involving the word "the" and another 50K involving the word "is", and some more like that. So if I build an expression that is a product of four copies of the word-pairs, I get an expression of size $(50K)^4$ which is a petabyte or something like that. 

So, my plan is to pre-prune, so the only word -pairs considered are those actually in the sentence. But, for that, I need to know what the word-pairs are. And I need to know what those are before the first lookup happens. Which means .. after the flatten word-graph happens.

linas:
Sorry, I meant "number of disjuncts" being $(50K)^4$.  The size of the expression itself is fine. But normal pruning seems to happen on on disjuncts, not on expressions.

Maybe there is some way of doing pruning on expressions, directly, but this seems too complicated to even think about.

ampli:
My plan is to extend the expression pruning to do also some power pruning, in the hope to save significant time in `power_prune()`. (Note that the SAT parser makes power pruning on expressions - but of course, I plan to do it w/o SAT.)

linas:
> My plan is to extend the expression pruning 

Cool. I wish you'd done it yesterday. But that's OK; the changes I'm making are "not that big", and I'm doing a code review as I go along, and cleaning up minor things. Anther minor cleanup in `tokenize.c` coming in a few minutes.

PR #1429, opened 2023-02-11
Deindent if-elseif chain. Move debug printfs.

linas:
This makes the code easier to read; no functional change.

### DIFF
--- a/link-grammar/tokenize/tokenize.c
+++ b/link-grammar/tokenize/tokenize.c
@@ -3213,10 +3213,9 @@ static Dict_node *dictionary_all_categories(Dictionary dict)
 static X_node * build_word_expressions(Sentence sent, const Gword *w,
                                        const char *s, Parse_Options opts)
 {
-	Dict_node * dn, *dn_head;
-	X_node * x, * y;
 	const Dictionary dict = sent->dict;
 
+	Dict_node * dn_head = NULL;
 	if (IS_GENERATION(dict) && (NULL != strstr(w->subword, WILDCARD_WORD)))
 	{
 		if (0 == strcmp(w->subword, WILDCARD_WORD))
@@ -3238,11 +3237,11 @@ static X_node * build_word_expressions(Sentence sent, const Gword *w,
 		dn_head = dictionary_lookup_list(dict, NULL == s ? w->subword : s);
 	}
 
-	x = NULL;
-	dn = dn_head;
+	X_node * x = NULL;
+	Dict_node * dn = dn_head;
 	while (dn != NULL)
 	{
-		y = (X_node *) pool_alloc(sent->X_node_pool);
+		X_node * y = (X_node *) pool_alloc(sent->X_node_pool);
 		y->next = x;
 		x = y;
 		x->exp = copy_Exp(dn->exp, sent->Exp_pool, opts);
@@ -3275,11 +3274,12 @@ static X_node * build_word_expressions(Sentence sent, const Gword *w,
 		 * result is valid (in case of "\*" it means an empty dict or a dict
 		 * only with walls that are not used).  Use a null-expression
 		 * instead, to prevent an error at the caller. */
-		y = pool_alloc(sent->X_node_pool);
+		X_node * y = pool_alloc(sent->X_node_pool);
 		y->next = NULL;
 		y->exp = make_zeroary_node(sent->Exp_pool);
 	}
 
+	assert(NULL != x, "Word '%s': NULL X-node", w->subword);
 	return x;
 }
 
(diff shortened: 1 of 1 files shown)


linas:
Merging; this is safe. Sadly, the diff is hard to read; it just converts an error print into an assert.

link-grammar/tokenize/tokenize.c
```diff
@@ -3275,11 +3274,12 @@ static X_node * build_word_expressions(Sentence sent, const Gword *w,
 		 * result is valid (in case of "\*" it means an empty dict or a dict
 		 * only with walls that are not used).  Use a null-expression
 		 * instead, to prevent an error at the caller. */
-		y = pool_alloc(sent->X_node_pool);
+		X_node * y = pool_alloc(sent->X_node_pool);
 		y->next = NULL;
 		y->exp = make_zeroary_node(sent->Exp_pool);
 	}
 
+	assert(NULL != x, "Word '%s': NULL X-node", w->subword);
```
ampli, line 3282: It was not an `assert()` by intention, as historically it happened in releases due to rare conditions (that have been fixed for the next release) and I didn't think there is a need to crash the whole program due to that.

link-grammar/tokenize/tokenize.c
```diff
@@ -3393,45 +3393,23 @@ static bool determine_word_expressions(Sentence sent, Gword *w,
 	{
 		we = build_word_expressions(sent, w, w->regex_name, opts);
 	}
-	else
+	else if (IS_GENERATION(dict) && (NULL != strstr(s, WILDCARD_WORD)))
 	{
-#ifdef DEBUG
```
ampli, line 3398: This happened sometimes while development and the condition were too costly to check on production so I included it in debug mode. It could be done compactly by:
`dassert(!dict_has_word(dict, w->subword, ...)`.

link-grammar/tokenize/tokenize.c
linas: I did see this myself, when I had done something wrong with the dict. It was "obviously" a dict bug, so I had to fix it.  And that's why turning this into an assert seemed OK.

I can chance it ... what do you want? Make it a debug print, again?

linas: s/chance/change/

link-grammar/tokenize/tokenize.c
linas: I can put it back, but from my reading, it seemed to be the same as the null `X_node` assert.  Perhaps I misread.

link-grammar/tokenize/tokenize.c
ampli: I think that `assert()`  is better to be reserved for internal errors and not for definition errors external to the program, for which I think `prt_error()` + `exit(NONZERO)` is more appropriate.
Here, this condition could also happen due to a complex tokenization error that happens only on rare conditions, so a crash is not needed (as opposed to an `assert()` that validates an "impossible" condition that if it is false it means the program might have a memory trashing and all chances are lost.

link-grammar/tokenize/tokenize.c
ampli: Isn't these the opposite cases? Or did I miss something?
NULL X_node happens if a word is marked as known but is not found in the dict.
"Known word is unknown" happens when an unknown word is found in the dict.

linas:
I mean, if this makes you unhappy, I can revert this pull req. There's nothing in here I much care about; I was just reacting to this comment: 
```
/* FIXME Change it to assert() when the Wordgraph version is mature. */
```
and after I changed that, the rest of this pull req evolved naturally.

My goal is to keep you happy. Let me know what to do.

link-grammar/tokenize/tokenize.c
ampli: > ... the opposite cases?

Please recheck if they are indeed so.

link-grammar/tokenize/tokenize.c
linas: This assert is now in `tokenize/lookup-exprs.c` line 126 

I did recently hit it, due to a code bug: for the atomese backend, `exists_lookup` reported that a word was present in the dict, but the actual `dictionary_lookup_list` returned nothing.

I don't see how it is possible for dictionary errors to trigger this assert, but ... maybe. Thus, I don't know what kind of error message to print, that would make sense to the user.

link-grammar/tokenize/tokenize.c
linas: This code has moved to `tokenize/lookup-exprs.c` lines 46-128 and the flow of control is now a bit more clear.  

The `build_word_expressions()` is called if and only if `dict_has_word()` returned `true` during tokenization. Therefore, there must be word expressions; if there aren't any, then it's a bug in the code, not a bug in the dict.

ampli: The code is still not clear to me.
1.
In `dictionary_setup_defines()` we set `unknown_word_defined` iff the dict contains UNKNOWN_WORD.
https://github.com/opencog/link-grammar/blob/4672e07635c0c1772e59f806862cd31c17df4d07/link-grammar/dict-common/dict-locale.c#L426-L432

But in `determine_word_expressions()` you have changed the error message:
``` diff
 	else if (dict->unknown_word_defined && dict->use_unknown_word)
 	{
 		we = build_word_expressions(sent, w, UNKNOWN_WORD, opts);
-		assert(we, UNKNOWN_WORD " must be defined in the dictionary!");
+		assert(we, UNKNOWN_WORD " supposed to be defined in the dictionary!");
 		w->status |= WS_UNKNOWN;
 	}
```
However, it is not a case in which UNKNOWN_WORD must be defined in the dictionary and it is not (i.e. dict error) since we already know it **is** defined there. The old words indicated this is a code error ("supposed to be defined". Maybe they could be made clearer:
`"Internal error: " UNKNOWN_WORD " supposed to be defined in the dictionary");

2.
Regarding `assert(NULL != x, "Word '%s': NULL X-node", w->subword);`, I think I once got it when there was an idiom with a word that is not in the dict (e.g. `test_xxx: test.n;`)but matches a regex when there is no `<UNKNOWN-WORD>` in the dict.
I don't get it anymore in such a case, so it might have been fixed. But you see that there may be very strange dict-definition cases that the code may not handle properly.

(BTW, when I tested it just now I found an inconsistency in the case of an idiom with an unknown word when `<UNKNOWN-WORD>` **is** defined: It is not marked as an unknown word (e.g. `test_xxx: text.n;`). I don't have an idea how to fix that in the current tokenization infrastructure. I am also not sure whether a warning should be issued since it can be an idiom that includes a "dead" word that is no more a stand-alone word in the language).

3. This check was done in the old code when a word couldn't be issued (3320-3326):
https://github.com/opencog/link-grammar/blob/66f215a45a921660a6ae402f7f8dab74bed90b9e/link-grammar/tokenize/tokenize.c#L3310-L3326
The intention is to check for an internal error in which we arrive here for an existing word (the check is buggy since it should have use `dictionary_word_is_known()` to include a regex).

d(I made it only in DEBUG because it is maybe not so likely. But usually make my tests, like linkage comparisons and `tests.py` also in DEBUG mode.)

In the new code, such a check is missing.

linas: 1)  I'm pushing this change, shortly:
```
--- a/link-grammar/tokenize/lookup-exprs.c
+++ b/link-grammar/tokenize/lookup-exprs.c
@@ -242,7 +242,7 @@ static bool determine_word_expressions(Sentence sent, Gword *w,
        else if (dict->unknown_word_defined && dict->use_unknown_word)
        {
                we = build_word_expressions(sent, w, UNKNOWN_WORD, opts);
-               assert(we, UNKNOWN_WORD " must be defined in the dictionary!");
+               assert(we, UNKNOWN_WORD " has no expressions in the dictionary!");
                w->status |= WS_UNKNOWN;
        }
        else
```

2) I don't understand. If I do this:
```
test_grx test_amorite: Wd- & S+;
```
then `grx` is no longer an "unknown word"; its part of an idiom. Likewise `amorite` which would otherwise get matched by a regex. If I recall correctly, regex matching is NOT done, if a word is in the dict. 
```
test grx goes
test amorite goes
test glamorite goes
```

3)  I'm making this change right now:
```
        else
        {
+               if (dictionary_word_is_known(dict, s))
+               {
+                       prt_error("Internal Error: Word '%s' has bad status\n", s);
+                       return false;
+               }
                prt_error("Error: Word '%s': word is unknown\n", s);
                return false;
        }
```

ampli: > 2. [...]
> If I recall correctly, regex matching is NOT done, if a word is in the dict.

You are right, it is currently so also for words that appear only in idioms (however, I don't know any such words).
As far as I recall, the original problem was reported in an issue (by a user) as a crash, and I I found it got involved an idiom in the `ru` dict, which is second word is the digit `2`, which was not in the dict (and it still not there). But:
1. I could not locate this issue.
2. I have already removed it  from my TODO list because the problem doesn't happen any more.
3. It seems the dict got replaced or fixed since this idiom is not there any more.
4. The NUMBER regex is disabled now, and it seems NUMBER has never found in the dict.

So I guess this problem didn't involve a regex after all. I will continue to search for the issue.

PR #1430, opened 2023-02-11
Refactor the build of word expressions.

linas:
This splits out the lookup of word-expressions from the word-graph flattening. This is a significant reorganization.

### DIFF
--- a/link-grammar/tokenize/tokenize.c
+++ b/link-grammar/tokenize/tokenize.c
@@ -1062,6 +1062,17 @@ static void altfree(const char **alts)
 	free(alts);
 }
 
+static void gwappend(Word * w, Gword * gw)
+{
+	size_t len = 0;
+	if (w->gwords)
+		while (w->gwords[len] != NULL) len++;
+
+	w->gwords = realloc(w->gwords, (len+2) * sizeof(Gword *));
+	w->gwords[len] = gw;
+	w->gwords[len+1] = NULL;
+}
+
 /*
 	Here's a summary of how subscripts are handled:
 
@@ -3155,10 +3166,11 @@ static Word *word_new(Sentence sent)
 		const size_t len = sent->length;
 
 		sent->word = realloc(sent->word, (len+1)*sizeof(*sent->word));
-		sent->word[len].d= NULL;
-		sent->word[len].x= NULL;
+		sent->word[len].d = NULL;
+		sent->word[len].x = NULL;
 		sent->word[len].unsplit_word = NULL;
 		sent->word[len].alternatives = NULL;
+		sent->word[len].gwords = NULL;
 		sent->word[len].optional = false;
 		sent->length++;
 
@@ -3366,16 +3378,17 @@ static void add_empty_word(Sentence sent, X_node *x)
  */
 #define D_X_NODE 9
 #define D_DWE 8
-static bool determine_word_expressions(Sentence sent, Gword *w,
+static bool determine_word_expressions(Sentence sent,
+                                       Gword *w,
                                        unsigned int *ZZZ_added,
                                        Parse_Options opts)
 {
 	Dictionary dict = sent->dict;
-	const size_t wordpos = sent->length - 1;
 
 	const char *s = w->subword;
 	X_node * we = NULL;
 
+	const size_t wordpos = w->sent_wordidx;
 	lgdebug(+D_DWE, "Word %zu subword %zu:'%s' status %s",
 	        wordpos, w->node_num, s, gword_status(sent, w));
 	if (NULL != sent->word[wordpos].unsplit_word)
@@ -3383,7 +3396,6 @@ static bool determine_word_expressions(Sentence sent, Gword *w,
 
 	/* Generate an "alternatives" component. */
 	altappend(sent, &sent->word[wordpos].alternatives, s);
-	w->sent_wordidx = wordpos;
 
 	if (w->status & WS_INDICT)
 	{
@@ -3491,9 +3503,7 @@ bool flatten_wordgraph(Sentence sent, Parse_Options opts)
 	Gword **next;                 /* The next words */
 	const Gword *last_unsplit_word = NULL;
 	size_t max_words = 0;
-	bool error_encountered = false;
 	bool right_wall_encountered = false;
-	unsigned int ZZZ_added = 0;   /* ZZZ+ has been added to previous word */
 
 	assert(0 == sent->length, "Word array already exists.");
 
@@ -3567,11 +3577,12 @@ bool flatten_wordgraph(Sentence sent, Parse_Options opts)
 				assert(!wpp_old->used, "Word %zu:%s has been used",
 				       wg_word->node_num, wpp_old->word->subword);
 
-				/* This is a new wordgraph word.
-				 */
+				/* This is a new wordgraph word.  */
 				assert(!right_wall_encountered, "Extra word");
-				if (!determine_word_expressions(sent, wg_word, &ZZZ_added, opts))
-					error_encountered = true;
+
+				wg_word->sent_wordidx = sent->length - 1;
+				gwappend(wa_word, wg_word);
+
 				if ((MT_WALL == wg_word->morpheme_type) &&
 				    (0 == strcmp(wg_word->subword, RIGHT_WALL_WORD)))
 					right_wall_encountered = true;
@@ -3713,6 +3724,18 @@ bool flatten_wordgraph(Sentence sent, Parse_Options opts)
 		free(out);
 	}
 
+	bool error_encountered = false;
+	unsigned int ZZZ_added = 0;   /* ZZZ+ has been added to previous word */
+	for (size_t i=0; i<sent->length; i++)
+	{
+		Gword *gw = sent->word[i].gwords[0];
+		while (gw)
+		{
+			error_encountered |=
+				!determine_word_expressions(sent, gw, &ZZZ_added, opts);
+			gw ++;
+		}
+	}
 	return !error_encountered;
 }
 #undef D_FW
--- a/link-grammar/tokenize/tokenize.c
+++ b/link-grammar/tokenize/tokenize.c
@@ -3381,17 +3381,16 @@ static void add_empty_word(Sentence sent, X_node *x)
  */
 #define D_X_NODE 9
 #define D_DWE 8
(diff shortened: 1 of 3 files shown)


linas:
@ampli, please review and test. After this is merged, I want to split `tokenize.c` into two files, once focused on flattening, and another on the `X_node` lookup.  I'm itching to do this now, but can hold off till you've reviewed and are happy about the changes.

link-grammar/tokenize/tokenize.c
```diff
@@ -1062,6 +1062,17 @@ static void altfree(const char **alts)
 	free(alts);
 }
 
+static void gwappend(Word * w, Gword * gw)
```
ampli, line 1065: I think `gwordlist_append()` can be used instead.
This has the benefit of sharing code.
In this particular case, I would like to find a way to reduce the number of `malloc()` calls and change the function accordingly. (The same for `altappend()`.)

link-grammar/tokenize/tokenize.c
```diff
@@ -3567,11 +3577,12 @@ bool flatten_wordgraph(Sentence sent, Parse_Options opts)
...
 
-				/* This is a new wordgraph word.
-				 */
+				/* This is a new wordgraph word.  */
 				assert(!right_wall_encountered, "Extra word");
-				if (!determine_word_expressions(sent, wg_word, &ZZZ_added, opts))
-					error_encountered = true;
+
+				wg_word->sent_wordidx = sent->length - 1;
+				gwappend(wa_word, wg_word);
```
ampli, line 3585: `gwordlist_append(&wa_word[wg_word->sent_wordidx], wg_word);`

link-grammar/tokenize/tokenize.c
linas: Remarks:
* I like to pick "funny" names like `gwappend` instead of `gwordlist_append` because I study code using `grep`. When everything is called "gwordlist", it becomes very hard to find greps that return only those parts that are relevant. In this case, it probably doesn't matter. But still ... easy-to-search names are important.

* I looked at `altappend` and it is used in a lot of different places. At first, I thought I would be able to replace `altappend` by gwappend, everywhere, but this seems impossible.

link-grammar/tokenize/tokenize.c
linas: ?

linas:
I can see additional simplification and cleanup that can be done, here, but I don't want to do it in this pull req, because it will this make reviewing this pull req more complicated. !?

link-grammar/tokenize/tokenize.c
ampli: This is a suggestion on how to use the existing `gwordlist_append()` (defined in `wordgraph.c`) instead of  `gwappend()` (because I recommended removing `gwappend()`).

linas: I just now pushed a patch to make this clearer.

link-grammar/tokenize/tokenize.c
linas: Oh, I see ... there already is a function called `gwordlist_append` .. I misunderstood. Fixing now.

ampli:
> I can see additional simplification and cleanup that can be done, here, but I don't want to do it in this pull req, because it will this make reviewing this pull req more complicated. !?

I don't know how much more complicated it will be...
(Anyway, as I previously indicated, maybe not so detailed, my main tool for finding the correctness of these changes is to run my test that compares the detailed first 150K parses of each sentence, including the parses with nulls and the sentences starting with '*', for all the `basic` corpora of all languages but `th`. For `en`, the `fixes` batch is also checked.)

linas:
> my main tool

Perhaps some variant of this should be checked into the codebase? It wouldn't have to be perfect; just enough to be minimally usable; I could run it.

Anyway, should I merge this? I'm itching to split the big tokenize file into two parts.

ampli:
> Anyway, should I merge this? I'm itching to split the big tokenize file into two parts.

I guess that if it seems it provides the same parses (especially when checked on the languages with morphology (`ru`, `he`, `tr`, `lt`, `ge`) than it could be find to just merge it.

linas:
> if it seems it provides the same parses

I dunno, does it? I think it should, but I don't have any (easy) way of checking.

ampli:
> Perhaps some variant of this should be checked into the codebase? It wouldn't have to be perfect; just enough to be minimally usable; I could run it.

I think that if checked into the codebase, it should be exactly what I'm using and not a variant, so I can maintain it (it already needed changes many times due to library code changes (e.g. subtle details in the linkage output formatting).
It consists of several shell & Perl scripts, that have "--help" for arguments, but I will need to add help text with usage examples. It currently uses `/tmp` for the output files (taking about 17GB of space if all is OK so the `diff` output files are of zero sizes).

(Besides this tool, I have several more utilities to ease the development, for example, `lgtests` which looks in the code and lists the various tests, `lgverbosity` that do the same for the various verbosity numbers, `lgerror` that compares two regular batch runs under different branches and output the sentences that have an error in one but not the other, `lg` to run `link-parser` in sophisticated ways, `lgspeedup` to compare the speed and sentence errors of two branches on a list of corpora, and many more.)

linas:
> it should be exactly what I'm using 

Sure.  Put it into a root-level directory, called `application-test` or `system-test`. Or something like that. Don't put it in the `tests` directory, that's for unit-tests only.

linas:
Oh, and don't make it part of the tarball. That would be too much. I don't want bug reports from casual users.

ampli:
> I dunno, does it? I think it should, but I don't have any (easy) way of checking.

The first step is to produce batch output for all the corpora (of all the languages) from a branch before and after the patches and validate that the errors are the same. (If the patch changes the disjunct order then this is tricky because the number of errors for English sentences with random sampling may get changed due to `linkage_limit`, but I guess this is not the case with this patch).

For that purpose, I can send a PR with my script `lgerror` (that lists the problematic sentences if any). It includes documentation on how to run it.

linas:
> The first step

Heh. Given that you do not magically have the answer yet tells me that this script takes hours to run. ... and its hard to interpret the results. :-/

ampli:
> Put it into a root-level directory, called application-test or system-test. Or something like that.

We already have a `debug` directory, maybe there? Or maybe under a directory `scripts` there?
I also don't mind toplevel `system-test`. Just tell me.

ampli:
> > The first step
> 
> Heh. Given that you do not magically have the answer yet tells me that this script takes hours to run. ... and its hard to interpret the results. :-/

`lgerror` takes a fraction of a second. But you need to run the batches of all the languages first from two branches.  This take several minutes, and I only have ready scripts to do that for `en` and `ru`. (But running by hand takes maybe 15 minutes.)

What takes most of the time is to download the PR into an new branch + configure/compile.
The detailed linkages comparison runs by two scripts that take about 3 minutes to complete (together).

linas:
Not the `debug` directory; we ship that in the tarball, and I won't want to ship this.  So, either in a top-level `systest` or in a `scripts/systest`. But since there is no `scripts` dir yet ... perhaps a top-level `systest`  would be better. 

I suspect you have more experience with diverse systems than I do. How do other people do it?

ampli:
> How do other people do it?

Do what?

> perhaps a top-level `systest` would be better.

 Ok. I will send `lgerror` first since it is standalone (it doesn't invoke other scripts).
You can test it on, for example the `en` batches, especially the big ones.

ampli:
Ok, I will download the PR and test it, please stay tuned...

linas:
>> How do other people do it?

How do other people put ssystem-test-like tools into git repos? What's the common convention? 

Back when I worked for a large company, there was this sad team of system testers, and they kept all their code in a giant indecipherable mega-pile. It was the accumulation of 10 or 20 years of customer complaints. The capable engineers would look at this, and then run in the opposite direction. Took me a long time to understand the psychology of developers. Still don't, I guess you could say.

ampli:
Strangely I got some conflicts. How come?
``` diff
--- a/link-grammar/tokenize/tokenize.c
+++ b/link-grammar/tokenize/tokenize.c
@@@ -3567,11 -3567,12 +3567,18 @@@ bool flatten_wordgraph(Sentence sent, P
                                assert(!wpp_old->used, "Word %zu:%s has been used",
                                       wg_word->node_num, wpp_old->word->subword);
  
 -                              /* This is a new wordgraph word. */
 +                              /* This is a new wordgraph word.
 +                               */
                                assert(!right_wall_encountered, "Extra word");
++<<<<<<< HEAD
 +                              if (!determine_word_expressions(sent, wg_word, &ZZZ_added, opts))
 +                                      error_encountered = true;
++=======
+ 
+                               wg_word->sent_wordidx = curr_widx;
+                               gwordlist_append(&wa_word->gwords, wg_word);
+ 
++>>>>>>> f7edead34 (Use the existing append function.)
                                if ((MT_WALL == wg_word->morpheme_type) &&
                                    (0 == strcmp(wg_word->subword, RIGHT_WALL_WORD)))
                                        right_wall_encountered = true;
```

linas:
> Strangely I got some conflicts

I pushed two more commits, to solve the two comments that you made.

ampli:
> I pushed two more commits, to solve the two comments that you made.

I just applied PR 1430 to the master branch. Still puzzling and checking what happens.

ampli:
Found the problem. Start testing.

linas:
? You said "use gwordlist_append instead of gwappend" and at first, I did not understand what you meant, ... and then when I did, I fixed it. 

I have no idea why you're merging into you're own master.   You could just create a branch, say `test=pull-1430` and then `git pull git@github/linas/link-grammar lookup-refactor` into it.

ampli:
> I have no idea why you're merging into you're own master. 

No, I merged to a copy of master in another worktree (`git worktree add ...`).

ampli:
It passed the detailed linkage test and the `pandp-union` "same batch errors" test.
It seems it runs somewhat faster than the current `master` for some of the batches.
I also ran it in DEBUG mode + ASAN/UBSAN on the `basic`, `fixes` and `fix-long` batches and didn't find problems.

So I guess it is "safe" to merge it.

linas:
Yayy!

linas:
> it runs somewhat faster 

Probably due to fewer icache misses.

PR #1431, opened 2023-02-11
systest/lgerror: Compare corpora batch output

ampli:


ampli:
It also can be used after dict changes.

linas:
That's all? I was imagining 6 different scripts and a thousand lines of code.  Maybe this should be moved to the debug directory, after all. I didn't expect this to be so small.

linas:
I mean, I already do something like this, manually. Except at the end, I just do a plain-old diff.

linas:
The biggest issue is that I don't have a good large corpus for russian, nor for thai.

I do keep a diary of old test results, but I haven't checked it into git, and I've accidentally deleted it more than once. Perhaps I should keep it in a branch.

ampli:
> That's all? I was imagining 6 different scripts and a thousand lines of code. Maybe this should be moved to the debug directory, after all. I didn't expect this to be so small.

As I said, I have **many** scripts that I use for testing, some I already detailed (and some are big...).
This one is **only** for finding the error diff between the regular output of batch runs.
E.g.
`link-parser < data/en/corpus-fixes.batch > /tmp/f0     # on master`
`link-parser < data/en/corpus-fixes.batch > /tmp/f1     # on some other branch`

And then using this script:
`lgerror /tmp/f[01]`

ampli:
> The biggest issue is that I don't have a good large corpus for russian, nor for thai.

It seems the current Russian corpus serves well for detecting program bugs and performance problems (by comparing to previous detailed linkage results).

For example, the current master runs about 10% slower than the current released version, and I guess this is due to not optimal table sizing done in the recent changes. Since much of the CPU for Russioan is not for parsing, I guess there is a huge performance hit on parsing, but I still need to CPU-profile it.

PR #1432, opened 2023-02-11
Move the sentence-expression code to it's own file.

linas:
This makes the huge file `tokenize.c` a little bit smaller, by moving all the word-expression code out of it, and into it's own distinct file. This should be "even safer", as it does not change the flow of control in any way. It just moves things around. Despite this, the patch may be hard to read.


linas:
FYI, after this change, I think it becomes clear that `sentence_in_dictionary()` is no longer needed; it does nothing useful, because the same lookup failure will happen earlier, in `build_sentence_expressions()`.

I will sleep on this, do something about this tomorrow.

ampli:
> I think it becomes clear that `sentence_in_dictionary()` is no longer needed

I also think so.

linas:
Performance: Sometime after pull req #1411 I saw an all-time best performance. Now its about 8% slower than that all-time best.  You mentioned seeing something like this as well. I'm bisecting to see why.

linas:
That's weird. I checkout out #1411 again, and I cannot reproduce the performance numbers I was measuring a few weeks ago.  It is not 8% slower than it used to be.

ampli:
My benchmark of the current master repository (4cdf9c562) vs 5.12.0 (average of 4 x best-of-3):
``` text
   2.068 -   2.038 =   -0.030  1.015x    data/en/corpus-basic.batch
   6.101 -   6.014 =   -0.087  1.014x    data/en/corpus-fixes.batch
   1.242 -   1.362 =   +0.120  0.912x    data/ru/corpus-basic.batch
   7.250 -   7.032 =   -0.218  1.031x    data/en/corpus-fix-long.batch
  52.600 -  53.380 =   +0.780  0.985x    data/en/corpus-failures.batch
 138.900 - 140.600 =   +1.700  0.988x    ../test2/data/en/pandp-union.txt
```
I looked at the CPU profile of the `ru` run and pcre2 takes there about 50% time less due to the added JIT. This saves several percent of CPU. But it is still shown here with 9% slowdown. I didn't see a particular spot with so much slowdown so it is still not clear to me. I will try to run a benchmark for `ru` after each commit starting from 5.12.0.
Note also that there is a significant speedup for `fix-long`.

PR #1436, opened 2023-02-14
PCRE2 reg_comp(): Disable JIT

ampli:
(Enabled in PR #1365.)
It is what causes the significant slowdown of `ru`.
I still left it enabled in `anysplit.c` since it seems to be slightly faster.

P.S. Maybe combining same-name patterns as alternations, may make JIT faster (need to check).
For `ru` it may be done by supporting something like:
``` text
MORPH-END-\0: /оряя$/
MORPH-END-\0: /еряя$/
...
```

### DIFF
--- a/link-grammar/dict-common/regex-morph.c
+++ b/link-grammar/dict-common/regex-morph.c
@@ -214,7 +214,8 @@ static bool reg_comp(Regex_node *rn)
 	                            options, &rc, &erroffset, NULL);
 	if (re->re_code != NULL)
 	{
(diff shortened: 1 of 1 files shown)


linas:
Thanks. Sorry for late response, my internet is down, my ISP Grande Communications, literally sabotaged my internet connection, and for the last week now, have refused to even look at the support ticket to fix it.  This is moving from criminal negligence to actual fraud and I'm not sure what to do any more. 

(I'm working from a temp wifi hotspot for now, with limited connectivity.)

PR #1441, opened 2023-02-20
Lengthen the empty connector

linas:
More characters in `EMPTY_CONNECTOR`

### DIFF
--- a/link-grammar/dict-common/dict-common.h
+++ b/link-grammar/dict-common/dict-common.h
@@ -22,7 +22,9 @@
 #include "memory-pool.h"                // Pool_desc
 #include "utilities.h"                  // locale_t
 
-#define EMPTY_CONNECTOR "ZZZ"
+// EMPTY_CONNECTOR must be at least 6 or 7 chars long, to avoid
+// collisions with autogened connectors in the Atomese dict.
+#define EMPTY_CONNECTOR "ZZZZZZZZZ"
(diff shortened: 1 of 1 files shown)


linas:
@ampli .. I'm merging this now, because I need it; however, a question for you: The English (and Indonesian) dict has `ZZZ` links in them. Is this the same `ZZZ` as `EMPTY_CONNECTOR` or is this some unrelated link that happens to have the same name?

ampli:
This is the same connector. It is currently used only for quotes and the virtual caps code (that it is a WIP that I need to find time to continue).

linas:
So in other words, I need to change those as well?  Pull req coming shortly.

I read through the code, and it was not clear that these were "the same connector".  The C code seemed to be about segmentation; and the dict entries seemed to be about rando quote marks. ... I'm somewhat surprised that they are related.

ampli:
See `determine_word_expressions()`. It invokes `is_exp_like_empty_word()` and `add_empty_word()`. You may also want to change the references to `ZZZ+` and `ZZZ-` in the comments there.

There is also a test named `removeZZZ` in linkage.c, with a hard-coded `ZZZ` in it. It should be change to fetch the name using `string_set_lookup(EMPTY_CONNECTOR, dict->string_set)` like the other places that use the empty connector.
(The test name should remain `removeZZZ` unless changed in ``tokenize/README.md`.)

PR #1445, opened 2023-02-21
Revised ZZZ-connector handling.

linas:
This revises handling of ZZZ links to avoid the issue of #1441

### DIFF
--- a/link-grammar/dict-common/dict-locale.c
+++ b/link-grammar/dict-common/dict-locale.c
@@ -443,7 +443,8 @@ bool dictionary_setup_defines(Dictionary dict)
 
 	// Used for unattached quote marks, in the English dict only.
 	dict->zzz_connector = linkgrammar_get_dict_define(dict, EMPTY_CONNECTOR);
-	dict->zzz_connector = string_set_add(dict->zzz_connector, dict->string_set);
+	if (NULL != dict->zzz_connector)
+		dict->zzz_connector = string_set_add(dict->zzz_connector, dict->string_set);
 
 	dictionary_setup_locale(dict);
 
--- a/link-grammar/dict-common/dict-locale.c
+++ b/link-grammar/dict-common/dict-locale.c
@@ -441,6 +441,10 @@ bool dictionary_setup_defines(Dictionary dict)
 
 	dict->shuffle_linkages = false;
 
+	// Used for unattached quote marks, in the English dict only.
+	dict->zzz_connector = linkgrammar_get_dict_define(dict, EMPTY_CONNECTOR);
+	dict->zzz_connector = string_set_add(dict->zzz_connector, dict->string_set);
+
 	dictionary_setup_locale(dict);
 
 	if (!dictionary_setup_max_disjunct_cost(dict)) return false;
--- a/link-grammar/dict-common/dict-utils.c
+++ b/link-grammar/dict-common/dict-utils.c
@@ -288,7 +288,7 @@ static bool exp_has_connector(const Exp * e, int depth,
 }
 
 /**
- * Find if an expression has a connector ZZZ- (that an empty-word has).
+ * Check if an expression has a connector ZZZ- (that an empty-word has).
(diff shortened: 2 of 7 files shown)


linas:
@ampli is this OK?

link-grammar/dict-common/dict-utils.c
```diff
@@ -297,7 +297,7 @@ static bool exp_has_connector(const Exp * e, int depth,
  **/
 bool is_exp_like_empty_word(Dictionary dict, Exp *exp)
 {
-	const char *cs = string_set_lookup(EMPTY_CONNECTOR, dict->string_set);
+	const char *cs = linkgrammar_get_dict_define(dict, EMPTY_CONNECTOR);
```
ampli, line 300: It may not be a good idea to use `linkgrammar_get_dict_define()` in parse run time due to these reasons:
1. It is currently a "costly" operation, as it linearly scans a list of definitions. (In any case, I should re-implement it as a dict word starting with `\4` and has a string value instead of an expression.)
2. In general connectors are expected to use the dict string set, and current /future code may depend on that for connector comparison. E.g. `exp_has_connector()`, which is invoked by `is_exp_like_empty_word()`, uses `string_set_cmp(e->condesc->string, cs)`.

Solutions:
1. Add `Dictionary::empty_connector`. Invoke `linkgrammar_get_dict_define(dict, EMPTY_CONNECTOR)` once, `dictionary_setup_defines()` to set it.
2. Use here and elsewhere `string_set_lookup(empty_connector, dict->string_set)` to make sure it is from the dict string set.

link-grammar/linkage/linkage.c
```diff
@@ -692,12 +692,16 @@ static void compute_chosen_words(Sentence sent, Linkage linkage,
 	 * to facilitate using diff on sentence batch runs. */
 	if (test_enabled("removeZZZ"))
 	{
-		for (i=0; i<linkage->num_links; i++)
+		const char* ZZZ = linkgrammar_get_dict_define(sent->dict, EMPTY_CONNECTOR);
```
ampli, line 695: Since this is just a test code, it can remain with `linkgrammar_get_dict_define()`.

link-grammar/dict-common/dict-utils.c
linas: OK, I pushed an update for solution 1.

link-grammar/dict-common/dict-locale.c
```diff
@@ -441,6 +441,10 @@ bool dictionary_setup_defines(Dictionary dict)
 
 	dict->shuffle_linkages = false;
 
+	// Used for unattached quote marks, in the English dict only.
+	dict->zzz_connector = linkgrammar_get_dict_define(dict, EMPTY_CONNECTOR);
+	dict->zzz_connector = string_set_add(dict->zzz_connector, dict->string_set);
```
ampli, line 446: To avoid a crash in case there is no definition:
``` c
if (NULL != dict->zzz_connector)
	dict->zzz_connector = string_set_add(dict->zzz_connector, dict->string_set);
```

ampli:
https://github.com/opencog/link-grammar/blob/01b836da26560295aa43a466de83134ff8e1df88/link-grammar/tokenize/lookup-exprs.c#L165-L180

On the same occasion, we can remove lines 177-179.

ampli:
The problem in the current master (01b836da2) happened in commit 417b59ac (unrecognized dict `ZZZ`).
The linkages of this PR (with my changes) compared to the master without this commit are identical.

link-grammar/dict-common/dict-locale.c
linas: Dohh. Yes, of course.

linas:
> On the same occasion

Fixed this too. Sorry, it's late at night, I'm in a hurry to get other things working, I keep making mistakes.  I think everything is in order, now.

ampli:
It passed my linkage tests (i.e. it doesn't modify the previous linkages).

PR #1446, opened 2023-02-21
Stop littering .lg_history files everywhere.

### LINKED ISSUE #1443: Stop litering `.lg_history` in current directory.
Currently, this line:

https://github.com/opencog/link-grammar/blob/af7033e541160a0d0b941d9ed70c7dff7f979361/link-parser/lg_readline.c#L375

causes the `.lg_history` file to be written in the current working directory, resulting in LG litter everywhere. It's getting annoying. It should be changed to either `~/.lg_history` or to `~/.cache/link-grammar/lg_history` 

I tried changing it to `~/.lg_history` but that did not work.  Seems that `history_w(hist, &ev, H_LOAD, HFILE);` does not know what to do with the tilde. And, at any rate, this won't work in windows, anyway.  So I'm opening this bug report instead ...

linas:
They are just .. everywhere on my system. Fixes #1443

This is a Linux-only solution. I don't know what the convention for Windows and Apple should be.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,6 +9,7 @@ Version 5.12.1 (XXX 2023)
  * English dict: paraphrasing fixes. #1398
  * Report CPU time usage only for the current thread. #1399
  * Extensive performance optimizations for MST dictionaries. #1402
+ * Stop littering `.lg_history` files in current working dir. #1446
 
 Version 5.12.0 (26 Nov 2022)
  * Fix crash when using the Atomese dictionary backend.
--- a/link-parser/lg_readline.c
+++ b/link-parser/lg_readline.c
@@ -372,7 +372,7 @@ char *lg_readline(const char *mb_prompt)
 
 	if (!is_init)
 	{
-#define HFILE ".lg_history"
+#define HFILE "~/.cache/link-grammar/history"
 		is_init = true;
 
 		size_t sz = mbstowcs(NULL, mb_prompt, 0) + 4;
@@ -384,7 +384,9 @@ char *lg_readline(const char *mb_prompt)
 		history_w(hist, &ev, H_SETSIZE, 100);
 		history_w(hist, &ev, H_SETUNIQUE, 1);
 		el_wset(el, EL_HIST, history_w, hist);
-		history_w(hist, &ev, H_LOAD, HFILE);
+		char * chist = expand_homedir(HFILE);
+		history_w(hist, &ev, H_LOAD, chist);
+		free(chist);
 
 		el_set(el, EL_SIGNAL, 1); /* Restore tty setting on returning to shell */
 
@@ -421,7 +423,9 @@ char *lg_readline(const char *mb_prompt)
 	if (1 < numc)
 	{
 		history_w(hist, &ev, H_ENTER, wc_line);
-		history_w(hist, &ev, H_SAVE, HFILE);
+		char * chist = expand_homedir(HFILE);
+		history_w(hist, &ev, H_SAVE, chist);
+		free(chist);
 	}
 	/* fwprintf(stderr, L"==> got %d %ls", numc, wc_line); */
 
--- a/link-parser/lg_readline.c
+++ b/link-parser/lg_readline.c
@@ -372,6 +372,8 @@ char *lg_readline(const char *mb_prompt)
 
 	if (!is_init)
 	{
+// This seems like the defacto convention for modern-day
+// dot-files on Linux systems. But what about Windows or Apple?
 #define HFILE "~/.cache/link-grammar/history"
 		is_init = true;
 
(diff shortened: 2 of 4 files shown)


ampli:
It turned out that this change is not good.

Many years ago I mentioned that it is rather inconvenient that the history of various languages is mixed in the same file. You then told me to  chdir first to the language dictionary directory before invoking link-parser. This solved my problem. Later I started to invoke `link-parser` only through a shell script, which can do many useful things, and among them, it changes the directory to that of the dictionary. With this change, this trick will not work.
In addition, `~/.cache/` is an XDG directory, but it is not intended for history files.

Solutions:
A. Leave it as it is currently, and just cd to the dictionary directory before invoking `link-parser`. Alternatively, use the script `lg` which I would like to eventually submit as PR. In any case - close this PR w/o applying it.
B. Implement the [XDG specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for the dictionary location of the history file.
1. Basically, it should be in the directory `~/.local/link-grammar/` unless the XDG environment variables tell otherwise.
3. Use per-language history file names, e.g. `lg_history-en`.
4. Handle macOS the same as above (as other POSIX programs do).
5. No need for history Windows support,  as Windows is built without libedit since its console supports editing.

**My recommendation is to first to do A and then I will send a PR for B.**

In any case, note that if the history file is maintained in a central place (and not per branch), the number of the history entries may get very big (slow `link-parser` startup), but this can be solved by something like `history size 200` in `~/.editrc`.

linas:
* OK, yes, history files can be per-language, I guess.
* Ideally, there is some little itty-bitty library that provides the correct API for dealing with XDG. I don't want to have to read that spec and try to guess if we've implemented it correctly, or not.
* On my system, `~/.local/` has only one subdir: `~/.local/share` which seems to hold config files (only).  By contrast, the `.cache` directory seems to hold the time-varying data. So I think we'll still want to use the `.cache` instead of `.local` but I have not carefully read the spec.

I'm closing this pull req. It's not that big an issue to me: I run `link-parser` from 1001 different directories, depending on what I'm doing, and these history files end up as litter.  I'm not going to start `cd`ing around to yet other locations, my life is already too complicated.

ampli:
According to the spec, the `.cache` directory is for files that can be removed without any functional implication for the user.
The spec mentions history file in one place (I marked it in boldface):
<pre>
$XDG_STATE_HOME defines the base directory relative to which user-specific state files should be stored. If
$XDG_STATE_HOME is either not set or empty, a default equal to $HOME/.local/state should be used.
The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not
important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
- actions history (logs, <strong>history</strong>, recently used files, …)
- current state of the application that can be reused on a restart (view, layout, open files, undo history, …)
</pre>

As you can see, the default location is `$HOME/.local/state`. This location first appeared in the newest XDG specification, of May 2021. My system has two programs there.

**Implementing an XDG location for the `link-parser` history using `libxdg-basdir` is easy, and I can send a PR.**
I propose that if not configured with `libxdg-basedir`, the library will just continue to create the history file as it does now.

BTW, a history file in a central place also has the benefit when testing the same sentence in different branches (currently I use copy/paste for that). As I mentioned, a problem appears if it becomes very big, but a limit in `.editrc` should solve that. Maybe we can also make a hard limit, e.g. 500 (or even lower), in `link-parser`.

linas:
> I can send a PR.

Yes, sure.

> we can also make a hard limit, e.g. 500 (or even lower),

Works for me. I can't imagine hitting the up-arrow key more than 100 times, anyway. searching for some old sentence.

BTW, I want to create a version 5.12.1 "real soon now".  Are there any pending pull reqs I should wait for, or can I do it, whenever?

ampli:
> BTW, I want to create a version 5.12.1 "real soon now". Are there any pending pull reqs I should wait for, or can I do it, whenever?

There are some of the problems I would like to fix for the new version:
1. Fix the Emscripten problem.
2. From the problems in `memory-pool.c`, fix the fake `pool_free().`
3. General cleanup, including typo/comment fixes.
4. Fix PR #1429 (restore a check).
5. Fix and update documentation.
6. Fix a string variable setting bug in `link-parser`.

Maybe also add this, which I mostly finished:
1. Add #define's for enabling/disabling the various caches in `do_count()` (finished, I just need to merge with the latest changes).
2. Simplify the match-list cache code and reduce memory (mostly done, need to debug and test).
3. Possible addition: Add match-list cache support for null_count>0 (if turned out not to be trivial, then not for this version).
4. Possible addition: Add an alternatives-compatibility check in `do_count()` (to reduce the number of unsane linkages in `amy`) (mostly finished, needs extensive tests, will be for the next version if cannot be done in a timely manner).

Make tests:
1. Linkage test.
2. Windows tests.
3. macOS tests.

linas:
OK.

Issue #1434, opened 2023-02-12
classic_parse: Sentence disjunct count 108279 exceeded limit 105123


ampli:
``` text
echo 'And yet he should be always ready to have a perfectly terrible scene, whenever we want one, and to become miserable, absolutely miserable, at a moment’s notice, and to overwhelm us with just reproaches in less than twenty minutes, and to be positively violent at the end of half an hour, and to leave us for ever at a quarter to eight, when we have to go and dress for dinner when, after that, one has seen him for really the last time, and he has refused to take back the little things he has given one, and promised never to communicate with one again, or to write one any foolish letters, he should be perfectly broken-hearted, and telegraph to one all day long, and send one little notes every half-hour by a private hansom, and dine quite alone at the club, so that every one should know how unhappy he was.' | lg -lim=10000 -sh=20 -v=2
%link-grammar-5.12.0-312-g4cdf9c562
%/tmp/link-grammar/master
%Dictionary directory: /usr/local/src/link-grammar-devel/master
+ time -o /tmp/link-grammar/master/time /tmp/link-grammar/master/link-parser/.libs/link-parser -lim=10000 -sh=20 -v=2
verbosity set to 2
link-grammar: Info: Dictionary found at ./data/en/4.0.dict
limit set to 10000
short set to 20
link-grammar: Info: Dictionary version 5.12.1, locale en_US.UTF-8
link-grammar: Info: Library version link-grammar-5.12.1. Enter "!help" for help.
#### Finished tokenizing (174 tokens)
++++ Finished expression pruning                 0.00 seconds
++++ Built disjuncts                             0.09 seconds
++++ Eliminated duplicate disjuncts              0.05 seconds
++++ Encoded for pruning                         0.08 seconds
++++ power pruned (for 0 nulls)                  0.07 seconds
++++ Built mlink_table                           0.00 seconds
++++ power pruned (for 0 nulls)                  0.01 seconds
++++ pp pruning                                  0.00 seconds
++++ power pruned (for 0 nulls)                  0.01 seconds
++++ Built mlink_table                           0.00 seconds
++++ power pruned (for 0 nulls)                  0.01 seconds
++++ Encoded for parsing                         0.00 seconds
++++ Initialized fast matcher                    0.00 seconds
++++ Counted parses (0 w/0 nulls)                0.93 seconds
++++ Finished parse                              0.00 seconds
No complete linkages found.
++++ Finished expression pruning                 0.00 seconds
[...]

PR #1447, opened 2023-02-21
Remove the max-number-of-disjuncts test.

linas:
Resolves issue #1443


linas:
Resolves issue https://github.com/opencog/link-grammar/issues/1434

Issue #1453, opened 2023-02-24
Get rid of max-cost


linas:
There is a long, old discussion about max-cost in #783 which suggests that getting rid of the concept of max-cost is a good idea. Newer evidence, in #1449 confirms this.

My memory of `max-cost` was that, originally, it was related to panic-parsing, and that setting it to a higher value would somehow make panic parsing "find something, anything".  But this seems untrue.  In the English dict, there is only **one thing** with three brackets: `[[[@Xc-]]]` that would be chopped by max-cost. (There is nothing with `[FOO+]3.0` because the panic-parse stuff predates the floating point costs.)

Rather than continuing conversations in the already-long #783, lets just do it here.  Or rather, lets just .. get rid of it, and be done with it.

As a side-effect, maybe get rid of panic-parse, also. Maybe.

ampli:
> **one thing** with three brackets: `[[[@Xc-]]]`

There are also things that get a cost >= 4 through `4.0.dialect`, the first one had originally  4 brackets:
1. `<no-det-null>: [()]headline;`
2. `then.#than: [[than.e]0.65]bad-spelling;` etc.

ampli:
> get rid of it

1. Do you mean that we should get rid of `parse_options_set_disjunct_cost()`?
2. What is so bad about this concept?
3. Without a disjunct cost limit, how do you implement dialects?

> As a side-effect, maybe get rid of panic-parse, also. Maybe.

The current panic parse parameters are indeed not useful to find a parse in a shorter time.
But its implementation infrastructure may be used to find a full parse in case a full parse has not been found (disregarding the parse time, by using new facilities that take more time). But the details may be considered as not belonging to this discussion...

linas:
> we should get rid of `parse_options_set_disjunct_cost()`?

Yes. For backwards compat, it would be a no-op.

> What is so bad about this concept?

The discussion and confusion in #783 and about what Dennis had done indicate that its a poorly formed concept, at least for the English dict. The English parses seem to work better, if it is set to infinity.

For the Atomese dicts, it is perhaps ... not such a bad concept, as there, some outrageously large costs show up, and could be discarded.

> Without a disjunct cost limit, how do you implement dialects?

!? Hmm. Interesting question. I'm not sure it's a good idea to turn dialects on and off, based on some cost threshold.

------------
OK, how about a proposal B:

* Keep the cost_cutoff idea
* Get rid of the cost_max computations in the code
[...]

PR #1455, opened 2023-02-25
Disable the use of cost cutoffs in the English dict.

linas:
Per discussions in #1453


ampli:
For some reason, this PR causes the `ru` parsing not to work at all.

ampli:
It seems that now `ru` needs at least cost-max=6 to work. But with any cost-max it doesn't work as intended, and these sentences are now parsed incorrectly:
``` text
<  *это тестами
<  *это тестом
<  *это тестями
<  прежде всего , требуется определить само понятие жизни .
<  скорее всего , дракон несет свою обычную стражу .
```

linas:
If russian does not work, then it is not due to this PR, but to #1456

Issue #1451, opened 2023-02-24
Flood-counting


linas:
One of the "deep" conceptual problems with LG parsing is that there are usually either not enough parses (i.e. null words) or there are too many (count overflow) and its hard to strike a middle ground. This particularly affects linkage generation: there (currently) isn't  any way to know the cost of a linkage, without actually computing it.

This issue is meant to be a place to daydream about possible solutions to this.

I have only one very vague and not really workable suggestion, for right now.  During counting, instead of counting 1 for each link, instead do a weighted count of $2^{-cost}$  for each link. 

Counting in this way replaces a grand-total by a weighted count, hinting at how many of these potential linkages are low-cost, vs. how many are high-cost. We could even make it smell more like simulated annealing by using $exp(-\beta \times cost)$ instead of $2^{-cost}$  where $\beta$ is a user-supplied "inverse temperature" parameter $\beta=1/kT$.

What is not yet clear to me is how to use this to replace the random sampling of linkages (when there is a count overflow) by a weighted random sampling, that would make lower-cost linkages more likely to be sampled.

ampli:
Can we just ignore the count overflow?
As each linkage is a path in the parse-set/parse-choice graph, maybe we can apply a heuristic algo to find the shortest (lower metric cost) paths. It is known that it is NP-hard, but I understood there are approximate algos that are fast and give results that are close to the optimum.

---
From [Wikipedia](https://en.wikipedia.org/wiki/Travelling_salesman_problem):

Various heuristics and approximation algorithms, which quickly yield good solutions, have been devised. These include the Multi-fragment algorithm. Modern methods can find solutions for extremely large problems (millions of cities) within a reasonable time which are with a high probability just 2–3% away from the optimal solution.

---
(The parse-set graph may also have special properties that enable shortcuts in these algos, but I guess it could be very time-consuming to investigate that.)

linas:
Idea two: issue-links in sorted order.

Currently, in `extract-links.c` the `exctract_links()` calls `list_links() with an index. This walks the left and right trees using a modulo numbering scheme.  This module scheme always explores the left parse-set exhaustively, before taking the next step on the right. I thought that perhaps it could work in a quasi-sorted order, by stepping left or right, depending on which one has lower cost. So I tried this:
```
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
[...]

PR #1457, opened 2023-02-25
Remove sort_match_list

linas:
Per comments in #1451 and from direct experiments from last night, it does not seem to make any difference.


ampli:
It actually changes the sampled linkages but generally doesn't reduce the disjunct cost of the lower-cost linkage.

linas:
Yes; I experimented with this during my earlier attempts.

PR #1460, opened 2023-03-02
Improve the CPU and memory efficiency of the match-list cache

ampli:
This patch simplifies the match-list cache code and saves 8 bytes per "word vector" element that doesn't have a count.
It also uses a single memory pool for the cache (instead of two).


linas:
Cool!

linas:
BTW, if you are doing other refactoring, and it looks like there is a blob of stand-alone code that it relatively unrelated to anything else, please move it to it's own file. 

The problem I'm having is that several of the existing c files are quite large, and have multiple "almost unrelated" subsystems in them. This makings reading and understanding the code harder, because its not clear where one subsystem stops, and the next one starts. I'm aiming for 500 lines-of-code or less, rule-of-thumb, for file sizes.

PR #1461, opened 2023-03-05
Provide a superior estimate for the xtable size.

linas:
From what I can tell, the tracon pool and the pset pool sizes are almost (not quite) identical. Because... there should be a parse set per link type. At least, that would be the case if each link type is used exactly once in a sentence. (Right? If used more than once, then there would be multiple parse-sets for that link type...)

(I have not measured this for English; only for MST, which has its own peculiar way of doing things.)

### DIFF
--- a/link-grammar/parse/extract-links.c
+++ b/link-grammar/parse/extract-links.c
@@ -127,6 +127,7 @@ static void record_choice(
 	s->num_pc++;
 }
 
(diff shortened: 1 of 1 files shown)


linas:
The result here should be checked for English & russian, before merging.

linas:
Revised the estimator to provide a happy medium.

PR #1469, opened 2023-03-25
count.c: Definitions to disable the use of various caches

ampli:
Add definitions to control the 4 caches used in `do_count()`:
``` c
const bool ENABLE_WORD_SKIP_VECTOR = true;
const bool ENABLE_MATCH_LIST_CACHE = true;
const bool ENABLE_TABLE_LRCNT = true;  // Also controls the above two caches.
const bool USE_TABLE_TRACON = true;    // The table is always maintained.
```
Disabling the lrcnt table also disables the match list cache and the word skip vector since they are kept in it.
On `USE_TABLE_TRACON=false`, the tracon table is still maintained since `mk_parse_set()` needs it. Also, in that case, its existence is used to validate for consistency when counts are recomputed (instead of using their cached value).

You can see that the caches have a low impact on short sentences but dramatically on long ones.
Disabling individual caches may help to debug changes in `do_count()`. e.g. it helped me to debug adding an "alternatives" check (an unpublished WIP).

### DIFF
--- a/link-grammar/parse/count.c
+++ b/link-grammar/parse/count.c
@@ -36,6 +36,9 @@
 typedef uint8_t null_count_m;  /* Storage representation of null_count */
 typedef uint8_t WordIdx_m;     /* Storage representation of word index */
 
+/* Allow to disable the use of the various caches (for debug). */
+const bool ENABLE_WORD_SKIP_VECTOR = true;
+
 typedef struct Table_tracon_s Table_tracon;
 struct Table_tracon_s
 {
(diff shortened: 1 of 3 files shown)


linas:
I'll review this a bit later; but this remark caught my eye:
> Disabling individual caches may help to debug

I'm sitting here, chuckling to myself. You can take any algorithm whatsoever, any at all, and add caches to it, and what you will get is hard-to-debug blob of software that misbehaves only in the corner cases that your customers are able to find.

PR #1471, opened 2023-05-07
Use correct guard macro for glibc heap functions

SoapGentoo:
* `__GNUC__` == Compiling with `-std=gnu*`
* `__GLIBC__` == Compiling against glibc

Bug: https://bugs.gentoo.org/903749


linas:
huh. Thanks!

PR #1473, opened 2023-06-18
link-grammar: add build patch

chenrui333:
Add build patch for 5.12.3 release build on macos

relates to https://github.com/Homebrew/homebrew-core/pull/134103

### DIFF
--- a/bindings/java/build.xml.in
+++ b/bindings/java/build.xml.in
@@ -2,8 +2,8 @@
 <project basedir="." default="build" name="linkgrammar">
     <property environment="env"/>
     <property name="debuglevel" value="source,lines,vars"/>
-    <property name="target" value="1.7"/>
-    <property name="source" value="1.7"/>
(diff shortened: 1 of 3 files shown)


linas:
Thanks! This will be included in version 5.12.4 to be published in a few weeks(?) maybe a few months.

linas:
Actually, I'm going to rever the part that says `HAVE_MALLOC_TRIM` -- this is not set anywhere. That block of code is protected by `__GLIBC__` ... are you saying that MacOS is pretending to be glibc, without actually being glibc?

linas:
I'm also confused by the `_WIN32` and not `__APPLE__` part -- are you saying that MacOS is pretnding to be WIN32 ??? that also seems kind of crazy, to me, but what do I know.  Anything is possible, I suppose ...

chenrui333:
I can help with some macos CI work tomorrow (that would surface the build issue more easily)

PR #1480, opened 2024-03-27
mk_partse_set: Avoid empty `Parse_set` elements

ampli:
Significantly reduce the number of the `Parse_set` elements. It addresses the issue where `mk_parse_set()` encounters unconnected elements in the absence of a complete parse for a word range. This not only conserves memory but also reduces CPU usage.

In the same occasion:
- Refactor for readability.
- Improve the x_table stats (also needed to check the  `Parse_set` saving).
- Add a comment on a bug if PERFORM_COUNT_HISTOGRAMMING is defined.


linas:
This fails the python unit tests; I get
```
bindings/python-examples/lg_testutils.py'>
............................................s...............sFs.........E.............s.s..............link-grammar: Fatal error:
Assertion (pc != NULL) failed in list_links() (../../link-grammar/parse/extract-links.c:920): walked off the end in list_links
```

PR #1483, opened 2024-03-27
Trim max number of allowed disjuncts.

linas:
Some Atomese dictionaries specify far too many disjuncts: one recent case has a billion (! with a b) in the left-wall (discussion #1479) This is due to a fault construction of the dictionary, but controlling this is difficult.  By contrast, it is easy to control at the link-grammar level. That's what this pull req does.

Disjuncts are selected randomly, up to the max number. This is a reasonable thing to do, as almost all disjuncts are going to give a valid parse, anyway, so there was never any "real reason" for the dictionary to have had that many. So it goes.

The max number is handled as a parse option, but it is up to the dictionary to provide a default value.


ampli:
More than 2G disjuncts will create a problem anyway, due to `rand_r()`. More than 4G will crash the library because the disjunct number variable is unsigned int. But I'm not sure we need to add tests for that...

PR #1485, opened 2024-03-31
Rework tracon hash function

linas:
Proposed fix for issue found in #1479

The Atomese backend can have 100K upper-case connectors, and no lower-case subscripts. This seems like a good compromise.


linas:
@ampli I'm merging this now, based on your earlier commentary. If this looks wrong, let me know.

ampli:
It seems there are still plenty of collisions, even with the new function. In tests conducted with just a small number of "normal" disjuncts, both this function and the original one led to significant collision problems, primarily in a few table entries that experienced the majority of collisions. Currently, I'm diving into debugging to figure this out.

PR #1487, opened 2024-04-15
Solves issue #1479 by fixing the connector/tracon hashing

ampli:
The cause of the performance problem is almost unbelievable: Neglecting to use a single bit from the connector descriptor. This bit is `Connector:desc->lc_mask & 1`. If we neglect it, the hash components for the connector strings `"dSTRING"` and `"STRING"` are identical.

The hash function used in `eliminate_duplicate_disjuncts()` has the same problem,  but it didn't cause extra collisions, most probably because the hash of the opposite jet is totally independent and thus adds enough entropy. However, I changed it too.

The new hash function is defined in `connectors.h`.

Additional changes:
- **tracon-set.c: Remove table shrinking**
The idea was to adapt to the pattern we see in the `en` and `ru` dicts, when the middle sentence words have many more disjuncts than the other words. It may not be so in the language learning project. This heuristic also has a bug in the case of packing for parsing (it happens to use the number of disjuncts per originating Gword, not word). The removal didn't have a noticeable performance reduction, and there are several better performance enhancement ideas to implement for packing speed.
- **Better debugging**
This is how I found the problem.
- **tracon-set: Add typedef for hash integer width**
Added to see if a wider hash will be better. It was somewhat slower, so it is still `uint32_t`.

I also found a bug that causes a few additional `tracon-set_reset()` calls, but its overhead is low so it is not fixed here.

BTW, I noted another potential problem, but a fix I tried didn't have a noticeable effect:
If a collision happens again on the same table slot (usually having a different hash value than the one already cached in that slot) the quadratic probing will be done independently on the hash value, colliding with all the previous collision resolution slots. It seems this can be mostly fixed by adding a term that depends on the hash value, e.g. `(h & 0x0f)`. I tried this and it didn't seem to improve the performance, I guess because collisions are now very rare.  So I didn't implement this.

### DIFF
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -309,6 +309,29 @@ static inline uint32_t string_hash(const char *s)
 	return i;
 }
 
+typedef uint32_t connector_hash_t;
+
+static inline connector_hash_t connector_hash(const Connector *c)
+{
+	return c->desc->uc_num +
+		(c->multi << 19) +
+		(((connector_hash_t)c->desc->lc_mask & 1) << 20) +
+		(connector_hash_t)c->desc->lc_letters;
+}
+
+/**
+ * \p c is assumed to be non-NULL.
+ */
+static inline connector_hash_t connector_list_hash(const Connector *c)
+{
+	connector_hash_t accum = connector_hash(c);
+
+	for (c = c->next; c != NULL; c = c->next)
+		accum = (19 * accum) + connector_hash(c);
+
+	return accum;
+}
+
 /**
  * Hash function for the classic parser linkage memoization.
  */
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -50,7 +50,7 @@
 
 typedef uint64_t lc_enc_t;
 
-typedef uint32_t connector_hash_t;
+typedef uint32_t connector_uc_hash_t;
 
 #define CD_HEAD_DEPENDENT    (1<<0) /* Has a leading 'h' or 'd'. */
 #define CD_HEAD              (1<<1) /* 0: dependent; 1: head; */
@@ -85,7 +85,7 @@ struct condesc_struct
 
 	const char *string;  /* The connector name w/o the direction mark, e.g. AB */
 	// float *cost;    /* Array of cost by connector length (cost[0]: default) */
-	connector_hash_t uc_num; /* uc part enumeration. */
+	connector_uc_hash_t uc_num; /* uc part enumeration. */
 	uint8_t length_limit; /* If not 0, it gives the limit of the length of the
 	                       * link that can be used on this connector type. The
 	                       * value UNLIMITED_LEN specifies no limit.
@@ -111,7 +111,7 @@ typedef struct length_limit_def
 typedef struct hdesc
 {
 	condesc_t *desc;
-	connector_hash_t str_hash;
+	connector_uc_hash_t str_hash;
 } hdesc_t;
 
 typedef struct
--- a/link-grammar/connectors.c
+++ b/link-grammar/connectors.c
@@ -541,7 +541,7 @@ static bool condesc_grow(ConTable *ct)
 
 condesc_t *condesc_add(ConTable *ct, const char *constring)
 {
-	uint32_t hash = (connector_hash_t)connector_str_hash(constring);
+	uint32_t hash = (connector_uc_hash_t)connector_str_hash(constring);
 	hdesc_t *h = condesc_find(ct, constring, hash);
 
 	if (NULL == h->desc)
--- a/link-grammar/disjunct-utils.c
+++ b/link-grammar/disjunct-utils.c
@@ -236,17 +236,15 @@ struct disjunct_dup_table_s
 static inline unsigned int old_hash_disjunct(disjunct_dup_table *dt,
                                              Disjunct * d, bool string_too)
 {
-	unsigned int i;
-	i = 0;
-	for (Connector *e = d->left; e != NULL; e = e->next) {
-		i = (41 * (i + e->desc->uc_num)) + (unsigned int)e->desc->lc_letters + 7;
-	}
-	for (Connector *e = d->right; e != NULL; e = e->next) {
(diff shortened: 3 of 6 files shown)


link-grammar/connectors.h
```diff
@@ -309,6 +309,29 @@ static inline uint32_t string_hash(const char *s)
...
 }
 
+typedef uint32_t connector_hash_t;
+
+static inline connector_hash_t connector_hash(const Connector *c)
+{
+	return c->desc->uc_num +
+		(c->multi << 19) +
+		(((connector_hash_t)c->desc->lc_mask & 1) << 20) +
+		(connector_hash_t)c->desc->lc_letters;
```
linas, line 319: Perhaps you want `((connector_hash_t)c->desc->lc_letters)<<7` or something like that? (It probably doesn't matter, but the butterfly in my stomach says I should mention this.)  or perhaps `* 101`? and maybe `(c->multi)*(1+(1<<3)+(1<<4)+(1<<19))` or some other pseudo-random-generating polynomial?

linas (approved):

linas:
Wow!

link-grammar/connectors.h
linas: For the record: I attempted the below, and saw no change in performance:
```
--- a/link-grammar/connectors.h
+++ b/link-grammar/connectors.h
@@ -314,9 +314,9 @@ typedef uint32_t connector_hash_t;
 static inline connector_hash_t connector_hash(const Connector *c)
 {
        return c->desc->uc_num +
-               (c->multi << 19) +
-               (((connector_hash_t)c->desc->lc_mask & 1) << 20) +
-               (connector_hash_t)c->desc->lc_letters;
+               c->multi * 266281 +
+               (((connector_hash_t)c->desc->lc_mask & 1) * 528397) +
+               ((connector_hash_t)c->desc->lc_letters) * 101;
 }
 
 /**
```

ampli: This one needs more CPU than the currently used one, which may offset its possible better hashing ability.
Also (assuming 32-bit hashes) for connectors with 4 subscript letters, the 101 shifts out about one letter.

The current hash formula and those that are suggested above have another problem: when there is more than one connector in a jet, some significant bits are shifted out due to the multiplication in `connector_list_hash()`. This could cause hash table trashing due to a problematic disjunct list that has enough jets with this bit loss. It can be fixed by adding/xoring the about-to-be-lost significant bits into the hash value before the multiplications (or shifts) but this will cause an extra overhead in our currently typical cases.

I started a branch that computes the connector hashes when creating connector descriptors. These hashes are to be used in the disjunct deduplication and the packing stages, saving significant CPU by preventing recomputations. I also found a way to exploit tracons for saving hash operations. Maybe then we could use a more complex hash function because it will be invoked much fewer times. (I also added code to always maintain the disjunct count because counting disjuncts several times in the code has also a significant CPU impact).

linas: I just noticed this comment just now.

* Are you saying that multiplication e.g. `(((connector_hash_t)c->desc->lc_mask & 1) * 528397)`  is slower than the shift `(((connector_hash_t)c->desc->lc_mask & 1) << 20)`  I would have guessed that there's no difference on modern CPUs.
* I just now pushed demo hash code in this commit:  506fc7f9d4fc710fcfae1ec7e5c37824a35f20cf which feeds back the high-order bits to the low-order bits. I did not enable it.  I don't know if there's some clever way of using multiplies to "trick" the cpu into dong that right-shift.

ampli: > * I would have guessed that there's no difference on modern CPUs.

AFAIK, 32-bit multiplication has typically a latency 3 cycles vs 1 for shift. The throughput of a series of multiplications can be one per cycle and for simpler operations several per cycle. See for example the instruction table for [Intel Ice Lake and Tiger Lake](https://www.agner.org/optimize/instruction_tables.pdf#page=348), starting at page 348.

I also made a benchmark of 10 billion calls to each of the following functions (quoted from the benchmark GPT4 created for me...)
``` c

void empty_function(connector *c) {
}

connector_hash_t simple_hash(connector *c) {
    return c->desc->uc_num +
           (c->multi << 19) +
           (((connector_hash_t)c->desc->lc_mask & 1) << 20) +
           (connector_hash_t)c->desc->lc_letters;
}

connector_hash_t complex_hash(connector *c) {
    return c->desc->uc_num +
           c->multi * 266281 +
           (((connector_hash_t)c->desc->lc_mask & 1) * 524429) +
           ((connector_hash_t)c->desc->lc_letters) * 101;
}
```
The main function, in another file, sums their results. The empty function is for estimating the loop and call overhead.

Results (ns/call):
empty_function: 0.94
simple_hash: 0.24
complex_hash: 0.45

---

Regarding feeding back the high-order bits: I don't think it is possible to achieve the effect of a shift-right by a mere multiplication. However, another method to prevent loss of entropy due to overflow is
``` c
accum = (19 * accum + connector_hash(c)) % large_prime;
```
This modulo operation can be converted to a shift-right and multiplication (Barrett reduction).

To compare the various hashes, collision and average chain stats can be generated by configuring with
`CFLAGS=-DTRACON_SET_DEBUG` and inspected by
`echo LONG_SENTENCE | link-parser -v8 -debug=tracon-set.c | grep delete`. But it is most probably a waste of time unless we encounter another hashing pathological case.

BTW, I tried to run the English batches with a tracon-set table loading 3/8 and 1/2, and even this doesn't have noticeable runtime differences (despite more collisions).  I avoid for now more accurate benchmarks because they take hours.  (When trying it I found a slight bug introduced in the recent load-factor change, and I will send a PR soon.)

linas: Wow. OK. I saw (almost) no difference between 3/8, 3/4 and 7/8 but a lot slower at 8/8. Then I tried 8/8*N -12 and got performance almost the same as 7/8 -- So something only a little more roomy than straight N is enough to get good behavior.

ampli: > 8/8*N -12

This looks strange. I need to run it in debug mode to see what really happens...

PR #1488, opened 2024-04-16
Reduce the size of the string set

linas:
After recent fixes, it seems that the size of the string set can be reduced.

### DIFF
--- a/link-grammar/string-id.h
+++ b/link-grammar/string-id.h
@@ -49,8 +49,8 @@ struct String_id_s
 };
 
 /* If the table gets too big, we grow it. Too big is defined as being
- * more than 3/8 full. There's a huge boost from keeping this sparse. */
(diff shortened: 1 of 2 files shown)


linas:
@ampli if you think this is wrong, let me know. It seemed like the right thing to do.

ampli:
The recent fixes were in tracon-set.
When I last revised the string-set code, I benchmarked it using the English corpus batches and found that 3/8 is better than even 1/2. Since it doesn't take much memory, there is no need to save memory there. Also, if I succeed in my plan to significantly speed up the parsing, every CPU saving in the other parts of the library will have a bigger impact. (Regarding string-id, it only has very few items in all its current usages, so I assume its initial table size will not increase anyway. But even if it will have usages with many items - there is no need to save memory there.)

PR #1498, opened 2024-04-20
Remove reference to argp-standalone from README

ryandesign:
argp-standalone isn't required anymore as of f4300ee9.


linas:
OK Thanks; I assume you've speaking from experience.

ryandesign:
I verified link-grammar 5.12.4 build on macOS versions 10.6–14 inclusive without argp-standalone. I didn't test any other BSD-derived operating systems.

ampli:
Indeed argp-standalone is not required anymore.

Issue #1495, opened 2024-04-20
Build errors


ryandesign:
Building link-grammar 5.12.4 on macOS 12 fails for me:

```
[...]

PR #1504, opened 2024-04-21
Only use once_flag/tss_t vars if we have threads.h

ryandesign:
Partial fix for #1495


linas:
Thanks! I was planning on publishing a version 5.12.5 "real soon now": maybe a few days, maybe a few weeks.  However, if you think you'll have more work on #1495 I can hold off on doing this, so as to get everything all wrapped up in one nice package.

Issue #1351, opened 2022-11-04
Surprising multi-connector behavior


linas:
Multi-connectors are behaving in a fashion I do not fully understand. Consider this dict
```
LEFT-WALL: @ANY+;

<many>: {[@ANY-]-0.1 or [@ANY+]-0.11};

<UNKNOWN-WORD>:  <many>;
% <UNKNOWN-WORD>:  <many> and <many>;
% <UNKNOWN-WORD>:  <many> and <many> and <many>;
% <UNKNOWN-WORD>:  <many> and <many> and <many> and <many>;
```
and then parse the sentence `asdf asdf asdf asdf. I get this:
```
linkparser> asdf asdf asdf asdf
Found 14 linkages (14 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS=-0.43 LEN=6)

    +---------------ANY--------------+
    |        +----------ANY----------+
    |        |       +------ANY------+
    |        |       |       +--ANY--+
    |        |       |       |       |
LEFT-WALL asdf[?] asdf[?] asdf[?] asdf[?]
```
and the others are all trees.  Not surprising.

When I set `<UNKNOWN-WORD>:  <many> and <many>;` then I get
```
Found 435 linkages (435 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS=-0.84 LEN=3)

             +----------ANY----------+
             |       +------ANY------+
    +---ANY--+       +--ANY--+--ANY--+
    |        |       |       |       |
LEFT-WALL asdf[?] asdf[?] asdf[?] asdf[?]
```
loops now appear. Not surprising. this is what I wanted. making the costs negative means that the loopy parses are shown first.  But oddly, this is not showing the maximum-possible number of loops!

With `<UNKNOWN-WORD>:  <many> and <many> and <many>;` I get
```
Found 939 linkages (939 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS=-1.15 LEN=3)

             +----------ANY----------+
             |       +------ANY------+
    +---ANY--+--ANY--+--ANY--+--ANY--+
    |        |       |       |       |
LEFT-WALL asdf[?] asdf[?] asdf[?] asdf[?]
```
Much better!  But wait, there's more: with four`<many>`'s I get
```
Found 1020 linkages (1020 had no P.P. violations)
	Linkage 1, cost vector = (UNUSED=0 DIS=-1.25 LEN=6)

[...]

PR #1528, opened 2024-05-17
Improve the connector hashing and extend condesc_t

ampli:
This patch enumerates the connector descriptors and uses this number in the connector hash functions.
(I also plan to use it as an array index in a future parse speedup.)

It also extends the information obtained through Connector::desc without increasing its 32-bit size by adding a pointer to `condesc_t` pointing to extra data. This will allow us to add more data per connector type, e.g., costs for multi-connectors (#1351, #1352) and a cost table per connection length (I think mentioned in #632, and a placeholder added in 2c1d9791d2). The idea is that the data needed during parsing for easy_match is still directly available, while obtaining other data before and after the parsing may need an extra redirection.

The hashing functions have also improved, and they should now keep the collisions low, hopefully without pathological cases (the current hash doesn't differentiate between connectors with more than 4 UP and 4 LC characters, and increasing `connector_hash_t` to `size_t` turned out to be too costly).
I handled two types of collisions:
1. Different connector lists that map to the same 32-bit hash value. Enabling the existing per-connector feedback hash (with different multiplications constants) reduced them by 2+ orders of magnitude.
2. Extra collisions due to different hash values that hashed to the same slot. The added Fibonacci multiplication significantly reduced them.

I've also included a few minor efficiency tweaks. I tested the speed on the corpus batches and the `big-dict` (from #1479), and it seems there is a small speedup (but this patch is not about speedup).

### DIFF
--- a/link-grammar/connectors.c
+++ b/link-grammar/connectors.c
@@ -46,7 +46,7 @@ static unsigned int get_connector_length_limit(condesc_t *cd,
 
 	int short_len = opts->short_length;
 	bool all_short = opts->all_short;
-	int length_limit = cd->length_limit;
+	int length_limit = cd->more->length_limit;
 
 	if ((all_short && (length_limit > short_len)) || (0 == length_limit))
 		return short_len;
@@ -165,7 +165,7 @@ static void set_condesc_length_limit(Dictionary dict, const Exp *e, int length_l
 		if (econlist[en]->uc_num != sdesc[cn]->uc_num) continue;
 		restart_cn = cn;
 
-		const char *wc_str = econlist[en]->string;
+		const char *wc_str = econlist[en]->more->string;
 		char *uc_wildcard = strchr(wc_str, LENGTH_LINIT_WILD_TYPE);
 
 		for (; cn < ct->num_con; cn++)
@@ -181,11 +181,11 @@ static void set_condesc_length_limit(Dictionary dict, const Exp *e, int length_l
 			else
 			{
 				/* The uppercase part is a prefix. */
-				if (0 != strncmp(wc_str, sdesc[cn]->string, uc_wildcard - wc_str))
+				if (0 != strncmp(wc_str, sdesc[cn]->more->string, uc_wildcard - wc_str))
 					break;
 			}
 
-			sdesc[cn]->length_limit = length_limit;
+			sdesc[cn]->more->length_limit = length_limit;
 		}
 	}
 }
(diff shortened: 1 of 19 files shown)


link-grammar/connectors.c
```diff
@@ -424,7 +424,7 @@ static bool sort_condesc_by_uc_constring(Dictionary dict)
 		condesc_t *condesc = dict->contable.hdesc[n].desc;
 
 		if (NULL == condesc) continue;
-		calculate_connector_info(condesc);
+		calculate_connector_info(&dict->contable.hdesc[n]);
 		sdesc[i++] = dict->contable.hdesc[n].desc;
```
linas, line 428: this seems to be same as `condesc`

link-grammar/connectors.h
```diff
@@ -97,11 +100,24 @@ struct condesc_struct
...
+ * other steps. The con_num field is used a lot in steps that need
+ * connector hashing, and it is included here to avoid extra
+ * redirections.
+ * Multi connectors are considering the same type as their non-multi
+ * version, so the multi indication is kept in Connector_struct. */
+struct condesc_struct
+{
+	lc_enc_t lc_letters;
+	lc_enc_t lc_mask;
+	hdesc_t *more;       /* More information, for keeping small struct size. */
```
linas, line 117: So `hdesc` points at `condesc`, and `condesc` points at `hdesc`. Interesting ... the old code did not have this loop.

linas (commented):
the circular pointers seem unusual. Are you sure you meant that? Merging anyway.

link-grammar/connectors.c
ampli: `condesc` is `dict->contable.hdesc[n].desc`, not `dict->contable.hdesc[n]`.
To be clearer,  these lines can be changed to (not tested):
```
hdesc_t *hdesc = &dict->contable.hdesc[n];

if (NULL == hdesc->desc) continue;
calculate_connector_info(hdesc);
sdesc[i++] = hdesc->desc;
```

link-grammar/connectors.h
ampli: `hdesc` are the slots of the connector hash table. Each one that is used, has a `desc` pointer to a `condesc_t` element. Up to here it is a standard implementation of an open hash table.
The `more` elements of `condesc_t` need to point somewhere, and it was convenient to use the `hdesc` slots, which are already allocated (it requires less code). However, this is memory inefficient (and indirectly CPU inefficiently due to cache trashing) because most of the `hdesc` elements are empty (NULL `desc`).

The solution is to use a memory pool for the `more` elements. I will eventually send a cleanup PR for that (and the  `sort_condesc_by_uc_constring()` code.

linas: Oh! Ah! OK. I think I understand.  This wasn't obvious from casual perusal, so I thought I'd point it out.

PR #1534, opened 2024-05-24
link-parser's history file location control

ampli:
In this PR:
The location of the history file is according to, in this order.
`XDG_STATE_HOME`
`HOME`

The history file name is X_history when X is the dict name to support a different history file per dict. For example, `en_history`. To support using the same history file for more than one dict, X is the dict name until the first `:`. For example, the dicts `en` and `en:test` use the save history file `en_history`.

I tested it on MinGW (no editline, just that it doesn't break compilation), Cygwin, macOS, and Linux.
It cannot be tested on MSVC because the MSVC build recently became broken in more than one way (PR soon). (The only test needed for now is that the changes of this PR don't break the build, as there is no editline support for Windows.)

This PR includes a minimal XDG support module I wrote instead of using a system library. One reason is that the current libraries don't support XDG_STATE_HOME, which was added to the standard relatively recently.

Comments:
- If XDG_STATE_HOMOCE is invalid (not absolute or not writeable), it is ignored. However, if HOME is invalid (undefined, not absolute, or not writeable), there is no fallback, and the history is disabled.
- Currently, verbosity=4, intended to show the location of files, doesn't show the location of the history file because the `verbosity` variable is currently not in the needed scope.

EDIT: I tried to configure with HAVE_ASPRINTF=0 and got a segfault. So, I forced-push a fix. (This also demonstrates that using `#if HAVE_FEATURE` is better than `#ifdef HAVE_FEATURE` because it allows such easy tests.)

Ref. PR #1446.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -202,7 +202,7 @@ CXXFLAGS="-D_DEFAULT_SOURCE ${CXXFLAGS}"
 AX_GCC_BUILTIN(__builtin_types_compatible_p)
 AC_C_TYPEOF
 
-AC_CHECK_FUNCS(strndup strtok_r sigaction malloc_trim)
+AC_CHECK_FUNCS(strndup strtok_r sigaction malloc_trim asprintf)
 AC_CHECK_FUNCS(aligned_alloc posix_memalign _aligned_malloc)
 
 # For the Wordgraph display code.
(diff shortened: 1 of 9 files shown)


linas:
wow!

linas:
Interesting. I see the answer now: https://unix.stackexchange.com/questions/699333/xdg-cache-home-vs-xdg-state-home

PR #1536, opened 2024-05-29
link-parser fixes

ampli:
- `save_default_opts()`: Bugfix saving string defaults
  Fix the defaults for the verbosity and string parse-options.
- `make_dirpath(`: Bugfix dir memory leak on MS Windows
- `lg_readline.c`: At verbosity 4, print the history file location
- link-parser.c: Be really silent on `--quite`
- 
In addition:
- Fix message and comment typos.
- command-line.c: Correctly use `D_USER_FILES`
The intention is to show file paths only at this level and not at higher levels.
- Simplify code.
- Rearrange code.

### DIFF
--- a/link-parser/lg_readline.c
+++ b/link-parser/lg_readline.c
@@ -376,8 +376,9 @@ static unsigned char lg_complete(EditLine *el, int ch)
 		{
 			if (fchdir(cwdfd) < 0)
 			{
-				/* This shouldn't happen, unless maybe the directory to which
-				 * cwdfd reveres becomes unreadable after cwdfd is created. */
+				/* Unexpected error: may occur only if the directory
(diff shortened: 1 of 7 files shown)


link-parser/lg_readline.c
```diff
@@ -83,10 +83,11 @@ void find_history_filepath(const char *dictname, const char *argv0,
 		prt_error("Warning: xdg_get_home(XDG_BD_STATE) failed; "
 		          "input history will not be supported.\n");
 		history_file = strdup("dev/null");
-		return;
 	}
 
 	history_file = hfile;
```
linas, line 88: `else` is needed here. I'll put it in.

linas: fixed in 1a4775aefcdf609c4ad1f7f7b9150454eb7a3836

PR #1537, opened 2024-05-30
Windows fixes

ampli:
These fixes enable the build:
- `LinkGrammar.vcxproj`: Update the list of source file names
- `form_match_list()`: Don't use _Atomic

These fixes fix code that causes warnings:
- `table_lookup()`: On `!USE_TABLE_TRACON`, return Count_bin type
- `estimate_log2_table_size()`: Avoid warning on conversion to/from double
- Change `GNUC_NORETURN` to `NORETURN` and define for MSVC too

- MSVC: Remove the `vcxproj.filters` files
GitHub notifies: "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository." I don't understand this and hope it is just a GitHub bug.

- table_count(): Bugfix !USE_TABLE_TRACON
This is not a bug in the production code, but it happens when defining `USE_TABLE_TRACON` as `false` for debugging. In that case, the function always returns an irrelevant wrong count 1.  

The rest of the `table_lookup()` are improvements I found when fixing the w_Count_bin-related warning.


link-grammar/parse/fast-match.c
```diff
@@ -595,8 +595,8 @@ form_match_list(fast_matcher_t *ctxt, int w,
 	}
 
 #ifdef VERIFY_MATCH_LIST
-	static _Atomic(uint16_t) id = 0;
-	uint16_t lid = ++id; /* A local copy, for multi-threading support. */
+	static uint16_t id = 0;
+	uint16_t lid = ++id; /* A stable local copy, for multi-threading support. */
```
linas, line 599: This will clearly fail when multi-threading, as the increment `++id` can occur in two different threads (and so two threads get the same `lid`) which means that the assert at line 1470 of  `parse/count.c` might trigger.  I'm going to put this back, but then disable `VERIFY_MATCH_LIST` by default.

ampli: Such a problem cannot happen, since each LG thread accesses only its own data structures!

linas: fixed in fac11ac24b33244f5a524f2b23cbc7c1165d7341

ampli: Note that the lid is put in the element list that a `form_match_list()` returns, and since it is an auto variable, all the elements of the list are guaranteed to get the same lid. When  `form_match_list()` returns, the thread that has invoked it fetches the id from the first element and validates that the rest use the same id (to validate that the list has not been corrupted, as happens sometimes during development). So the code here doesn't care at all what lid values are used in other threads, and even not in the same thread (as long as lid is constant in a single invocation of form_match_list()`, which is guaranteed). They can be always the same, or any random ones.
The bug that got fixed than was due to an uninitialized id field on id==0.

linas: OK. I skimmed the code and saw this: 

https://github.com/opencog/link-grammar/blob/c09894035017c0cf1e4142b825ae4c23f3645175/link-grammar/parse/count.c#L1472

and (without over-analyzing it) made me think that assert might now trigger by accident. So I got nervous about it. So I put the Atomic back in, and undef'ed `VERIFY_MATCH_LIST` so it becomes invisible.

linas: Go ahead and remove the `_Atomic` if you are sure. I wasn't going to think too hard about it.

PR #1538, opened 2024-05-31
Small fixes (no. 2)

ampli:
Before I send more interesting changes:

- Fix `autogen.sh` behavior on bad arguments.
- Fix typos in comments, docs and code.


linas:
wow. English: it's confusing to think about its structure. Did I get that right?

PR #1541, opened 2026-05-26
Viterbi extractor

ampli:
Hello Linas,
  
  This PR changes linkage extraction for cases where a sentence has more
  linkages than requested, especially combinatorial-explosion cases. Instead of
  randomly sampling raw linkages, the link extractor returns linkages in cost
  order, so low linkage limits such as !limit=1 can still find the lowest-cost
  valid linkage.

  To make cost-ordered extraction practical, the English post-processing rule
  set was reduced to the single rule that still needs classic postprocessing.
  Most rules were moved into the dictionary, and a few remaining exact checks
  were moved into the metric/Viterbi-style extractor. The remaining bounded s
  rule uses feedback from postprocessing so the extractor can skip later
  same-shape violations before linkage materialization.

  The PP-rule migration was validated on `corpus-knowledge.batch`, a focused
  English grammar-regression corpus, and on the broader English corpora
  corpus-basic.batch, corpus-fixes.batch, corpus-fix-long.batch, and
  corpus-failures.batch. The observed sentence-status changes were improvements
  rather than regressions. Linkage structure and disjunct costs were also
  compared on many focused examples where replacement paths could affect the
  preferred linkage. Performance improved substantially on the main English
  corpora, while corpus-failures.batch is somewhat slower. For sentences with
  combinatorial explosion, returning the lowest-cost valid linkages first is a
  much more meaningful result than sorted random extraction, which can still miss
  the low-cost region.

  This is a large piece of work touching the parser, English PP-rule migration,
  corpus linkage limits, tests, and documentation. Most needed is a review of
  the PP-rule migration into the English dictionary, since this part is far from
  my own expertise and I cannot judge Codex's grammar changes well enough by
  myself. Review of the metric/Viterbi-style extractor approach and validation
  strategy is also welcome.

  The PP-rule migration is documented in `data/en/GRAMMAR-FIXES.md`.
  The metric/Viterbi-style extractor design, PP-blocker handling, and validation
  strategy are documented in `link-grammar/parse/METRIC-EXTRACTION.md`.

  Amir


linas:
Amir, if I recall correctly, I told you back in May that I'd be busy for months and would not be able to connect and comment. Well, sadly, more than a few months have gone by, but this is still on my to-do list. I'm not ignoring you, just that there are a lot of things going on, and arranging an uninterrupted time slot where I can think clearly about this remains slightly out of reach.  Thinking requires effort.  But soon... maybe a few weeks, a month maybe still two months, the more you make noise in my direction, the higher it goes up on my priority scale.

ampli:
Hi Linas,

Thanks for letting me know. I understand that this needs some uninterrupted time.

One aspect you may find interesting is how much of the postprocessing could be moved into the dictionary. It reminded me of your README discussion about lexical functions and propagating constraints through links. The changes provide concrete examples, including the filler-it constructions, and also make clearer why the remaining cases were harder to express that way. The explanations are in `data/en/GRAMMAR-FIXES.md`.

The cycle constraints and the remaining inversion checks are handled by the Viterbi extractor. For the one rule that still needs PP, feedback allows the extractor to skip further linkages with the same violating structure. In the difficult long-sentence example, it learned two such constraints and examined only three raw linkages before finding a valid one. So the PP bottleneck has largely been eliminated too.

BTW, the slight slowdown on long sentences mentioned in the PR has been resolved in an additional patch that I have not published yet.

Amir

PR #1543, opened 2026-09-16
link_grammar.i: Use PyLong_* instead of PyInt_*

daeho-ro:
SWIG 4.5 removed the Python 2 compatibility macros (`PyInt_Check`, `PyInt_AsLong`, `PyInt_FromLong`, ...) that it used to keep in the generated wrapper for user typemaps (see the 2026-07-06 entry in SWIG's `CHANGES.current`). The typemaps in `bindings/swig/link_grammar.i` still use them, so the Python bindings fail to compile with SWIG 4.5


linas:
Thank you!

